Issue #27, opened 2018-02-09
Update docker image to be based on Atomic


disrani-px:
Currently, the docker image is based on Ubuntu, we should move it to atomic.

PR #28, opened 2018-02-17
Move the container to be based on atmoic

disrani-px:
Issue #27


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu
```
harsh-px: can we build this without registry access?

disrani-px: That is the official registry for atomic images. Doesn't require any login.

PR #37, opened 2018-03-02
Support to prioritize nodes based on rack/zone/region info

disrani-px:


disrani-px:
Yes, we can just update the driver to use info from the cluster.

PR #41, opened 2018-03-06
Fix error check in extender prioritize request

disrani-px:


pkg/extender/extender.go
```diff
@@ -237,6 +241,7 @@ func (e *Extender) processPrioritizeRequest(w http.ResponseWriter, req *http.Req
 		}
 	}
 
+sendResponse:
 	// For any nodes that didn't have any volumes, assign it a
 	// default score so that it doesn't get completelt ignored
```
piyush-nimbalkar: nit: completely
can do this later

Issue #49, opened 2018-03-27
Check other source for provisioner name


disrani-px:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG REPORT
**What happened**:
The Portworx driver checks the storage class to determine if a volume has been provisoned by it.
The storage class can be deleted after a PVC has been provisioned though. So we need to use other methods to check the provisioner

**What you expected to happen**:
Should be able to detect that a PVC  was provisioned by Portworx even after the storageclass has been deleted.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create PVC
2. Delete StorageClass
3. Start pod which uses PVC. 
4. Stork logs should complain about not being able to find storage class

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #50, opened 2018-03-27
Portworx driver: Check other sources for provisioner

disrani-px:
Issue #49

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -18,6 +18,7 @@ import (
 	"github.com/libopenstorage/stork/pkg/snapshot"
 	"github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -32,8 +33,11 @@ const (
 	// namespace is the kubernetes namespace in which portworx daemon set runs
 	namespace = "kube-system"
 
-	//provisionerName is the name for the driver provisioner
+	// provisionerName is the name for the driver provisioner
 	provisionerName = "kubernetes.io/portworx-volume"
+
+	// pvcProvisionerAnnotation is the annotation on PVC which has the provisioner name
+	pvcProvisionerAnnotation = "volume.beta.kubernetes.io/storage-provisioner"
```
piyush-nimbalkar: this should be defined in k8s package, right? can we use it from there?

disrani-px: Unfortunately, it's not exported from anywhere in kubernetes.

piyush-nimbalkar: ah okay. Just saw that we have defined it in sched-ops as well.

PR #53, opened 2018-03-30
Add pvc name label to the restored snapshot

piyush-nimbalkar:


drivers/volume/portworx/portworx.go
```diff
@@ -274,6 +277,9 @@ func (p *portworx) SnapshotRestore(
 	logrus.Debugf("SnapshotRestore for pvc: %+v", pvc)
 	locator := &api.VolumeLocator{
 		Name: "pvc-" + string(pvc.UID),
+		VolumeLabels: map[string]string{
+			pvcNameLabel: pvc.Name,
```
disrani-px: Also need to add the pvc namespace

PR #58, opened 2018-04-03
Don't check source PV when restoring from snapshot

disrani-px:
The original PV could have been deleted.

Issue #57


disrani-px:
Tested manually, currently not possible to test this scenario with torpedo

PR #59, opened 2018-04-05
Check PVC first for provisioner

disrani-px:
Issue #49


disrani-px:
No, that would be an invalid state.

Issue #67, opened 2018-04-20
Use uuid instead of snapshot name when creating OSD volumes


piyush-nimbalkar:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG REPORT

**What happened**:
Cannot create snapshots with same names even if different namespaces.

**What you expected to happen**:
Allow creating snapshots with same names from different namespaces.

**How to reproduce it (as minimally and precisely as possible)**:
Create two snapshots with same name in two different namspaces.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): all supported versions
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #68, opened 2018-04-20
Portworx: Use snapshot UID for snapshot name to make it unique

disrani-px:
Also add snapshot name and namespace as labels when creating the
snapshot

Issue #67


piyush-nimbalkar:
This will break the integration test, right?

disrani-px:
Good catch. Updated integration test.
Updated torpedo in vendor, will need to update torpedo when next version of stork is released.

PR #83, opened 2018-05-10
verify scheduled node only for mysql-data-1

harsh-px:

### DIFF
--- a/test/integration_test/snapshot_test.go
+++ b/test/integration_test/snapshot_test.go
@@ -90,13 +90,22 @@ func groupSnapshotTest(t *testing.T) {
 	err = schedulerDriver.InspectVolumes(ctxs[0])
 	require.NoError(t, err, "Error waiting for volumes")
 
-	allVolumeNames := getVolumeNames(t, ctxs[0])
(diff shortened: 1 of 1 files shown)


test/integration_test/snapshot_test.go
```diff
@@ -94,9 +94,13 @@ func groupSnapshotTest(t *testing.T) {
 	require.Len(t, allVolumeNames, 3, "Should only have two volumes and 1 snapshot")
 
 	dataVolumesNames := make([]string, 0)
+	dataVolumesInUse := make([]string, 0)
 	for _, volume := range allVolumeNames {
 		if !strings.HasPrefix(volume, "snapshot") {
 			dataVolumesNames = append(dataVolumesNames, volume)
+			if volume == "mysql-data-1" {
```
disrani-px: The volume name will be pvc-<UUID>, right?

harsh-px: Oh actually, the above check is incorrect. It again passed on my cluster since the data replicas get placed differently. Let me fix this.

harsh-px: Fixed

PR #84, opened 2018-05-11
Fix cloudsnap integration to check only data volumes in use

harsh-px:

### DIFF
--- a/test/integration_test/snapshot_test.go
+++ b/test/integration_test/snapshot_test.go
@@ -184,9 +184,21 @@ func cloudSnapshotTest(t *testing.T) {
 	require.Equal(t, 3, len(volumeNames), "Should only have two volumes and a snapshot")
 
 	dataVolumesNames := make([]string, 0)
-	for _, volume := range volumeNames {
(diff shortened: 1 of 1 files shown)


test/integration_test/snapshot_test.go
```diff
@@ -184,9 +184,21 @@ func cloudSnapshotTest(t *testing.T) {
...
 	dataVolumesNames := make([]string, 0)
-	for _, volume := range volumeNames {
-		if !strings.HasPrefix(volume, "snapshot") {
-			dataVolumesNames = append(dataVolumesNames, volume)
+	dataVolumesInUse := make([]string, 0)
+
+	allVolumes, err := schedulerDriver.GetVolumes(ctxs[0])
+	require.NoError(t, err, "failed to get volumes")
+	require.Len(t, allVolumes, 2, "should have 2 volumes")
+	for _, v := range allVolumes {
```
disrani-px: Looks like the same code as above test. Move into a method.

harsh-px: done

PR #93, opened 2018-05-23
update group snapshot annotation keys

harsh-px:

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -90,14 +90,15 @@ type cloudSnapStatus struct {
 
 // snapshot annotation constants
 const (
-	pxAnnotationPrefix        = "portworx/"
-	pxSnapshotTypeKey         = pxAnnotationPrefix + "snapshot-type"
-	pxSnapshotNamespaceIDKey  = pxAnnotationPrefix + "namespace"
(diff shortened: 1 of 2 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1117,7 +1118,7 @@ func parseGroupLabelsFromAnnotations(annotations map[string]string) map[string]s
 			continue
 		}
 
-		matches := pxAnnotationKeyRegex.FindStringSubmatch(k)
```
disrani-px: The changes to filter out labels above isn't really required.

drivers/volume/portworx/portworx.go
```diff
@@ -1132,7 +1133,7 @@ func isGroupSnap(snap *crdv1.VolumeSnapshot) bool {
 			continue
```
disrani-px: Same here, this check isn't required.

drivers/volume/portworx/portworx.go
harsh-px: Kept pxSnapshotGroupIDKey and removed others. Filtering out pxSnapshotGroupIDKey is required since that's a special annotation for group ID and not for generic selector labels.

drivers/volume/portworx/portworx.go
harsh-px: removed

PR #94, opened 2018-05-23
Ensure version check for group and cloud snapshot

### LINKED ISSUE #82: Check for PX version 1.4 and above for cloud and group snapshots
**BUG REPORT**: CS and group snapshot support is only in 1.4 

Fail cloud and group snapshots for 1.3 and lower versions. Also skip integration tests.

harsh-px:
Fixes #82

Example output when one attempts to create group snapshot on 1.2

```
 kubectl describe volumesnapshot
Name:         mysql-snapshot
Namespace:    default
Labels:       SnapshotMetadata-PVName=pvc-b707813c-5ee3-11e8-8cb9-080027ee1df7
              SnapshotMetadata-Timestamp=1527119320930117146
Annotations:  kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"volumesnapshot.external-storage.k8s.io/v1","kind":"VolumeSnapshot","metadata":{"annotations":{"portworx.selector/tier":"prod","portworx....
              portworx.selector/tier=prod
              portworx.selector/type=db
              portworx/snapshot-type=local
API Version:  volumesnapshot.external-storage.k8s.io/v1
Kind:         VolumeSnapshot
Metadata:
  Cluster Name:
  Creation Timestamp:  2018-05-23T23:48:32Z
  Generation:          0
  Resource Version:    6084733
  Self Link:           /apis/volumesnapshot.external-storage.k8s.io/v1/namespaces/default/volumesnapshots/mysql-snapshot
  UID:                 cceb976c-5ee3-11e8-8cb9-080027ee1df7
Spec:
  Persistent Volume Claim Name:  mysql-data-1
  Snapshot Data Name:
Status:
  Conditions:
    Last Transition Time:  2018-05-23T23:48:40Z
    Message:               snapshot failed due to err: group snapshot is supported only on PX version 1.4 and above. node: k2n1 has version: 1.2.22.0-3ebbe3f
    Reason:
    Status:                True
    Type:                  Error
  Creation Timestamp:      <nil>
Events:                    <none>

```

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -378,9 +378,13 @@ func (p *portworx) SnapshotCreate(
 	switch snapType {
 	case crdv1.PortworxSnapshotTypeCloud:
 		logrus.Debugf("Cloud SnapshotCreate for pv: %+v \n tags: %v", pv, tags)
-		err = p.ensureNodesDontMatchVersionPrefix(pre14VersionRegex)
+		ok, msg, err := p.ensureNodesDontMatchVersionPrefix(pre14VersionRegex)
 		if err != nil {
-			err = fmt.Errorf("cloud snapshot is supported only on PX version 1.4 and above. %v", err)
+			return nil, nil, err
+		}
+
+		if !ok {
+			err = fmt.Errorf("cloud snapshot is supported only on PX version 1.4 and above. %s", msg)
 			return nil, getErrorSnapshotConditions(err), err
 		}
 
@@ -389,7 +393,7 @@ func (p *portworx) SnapshotCreate(
 			VolumeID:       volumeID,
 			CredentialUUID: snapshotCredID,
 		}
-		err := p.volDriver.CloudBackupCreate(request)
+		err = p.volDriver.CloudBackupCreate(request)
 		if err != nil {
 			return nil, getErrorSnapshotConditions(err), err
 		}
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1084,6 +1097,22 @@ func (p *portworx) validatePVForGroupSnap(pvName, groupID string, groupLabels ma
 	return nil
 }
 
+func (p *portworx) ensureNodesDontMatchVersionPrefix(versionRegex *regexp.Regexp) error {
+	result, err := p.clusterManager.Enumerate()
+	if err != nil {
+		return err
```
disrani-px: This should be a retryable error, shouldn't mark it as failed in this case.

harsh-px: You mean put the VolumeSnapshot in Pending state so that the controller keeps retrying? Is the use case that users will upgrade Portworx to the 1.4 version and then the pending VolumeSnapshot will succeed?

disrani-px: No, if the version is old fail it. But if the cluster enumerate returns an error it needs to be retried.

harsh-px: done

harsh-px: updated as per discussion

PR #96, opened 2018-05-31
 Add an executor CLI for running async commands in pods

harsh-px:
This will eventually be used by 3D snapshots to execute commands in pods.


harsh-px:
The async part is not working with this approach. Closing until I figure out a path forward.

PR #98, opened 2018-06-07
Add a cache for node info 

disrani-px:
Querying the API server for each node takes too long and times
out the requests in a large cluster


pkg/extender/extender.go
```diff
@@ -129,15 +117,14 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
-
 	if err != nil {
 		storklog.PodLog(pod).Warnf("Error getting volumes for Pod for driver: %v", err)
 		if _, ok := err.(*volume.ErrPVCPending); ok {
 			http.Error(w, "Waiting for PVC to be bound", http.StatusBadRequest)
 			return
 		}
 	} else if len(driverVolumes) > 0 {
-		driverNodes, err := e.Driver.GetNodes()
+		driverNodes, err := e.Driver.GetNodes(args.Nodes.Items)
```
disrani-px: This still causes delays if only a small number of nodes are passed into the Filter request and there are a large number of nodes in the cluster.

drivers/volume/portworx/portworx.go
```diff
@@ -227,6 +276,17 @@ func (p *portworx) mapNodeStatus(status api.Status) storkvolume.NodeStatus {
 	}
 }
 
+func (p *portworx) getNodeLabels(hostname string) (map[string]string, error) {
+	obj, exists, err := p.store.GetByKey(hostname)
```
harsh-px: Does this assume that the k8s node name is always the hostname?

disrani-px: Yes, that was the assumption previously too when calling k8s.Instance().GetLabelsOnNode()

Issue #71, opened 2018-04-27
Allow clone from snapshot into a different namespace


disrani-px:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What happened**:
1. Take snapshot in namespace N1
2. Try to create in PVC in namespace N2 from the snapshot in step 1. This will fail

**What you expected to happen**:
Should be allowed to clone snapshots to any allowed namespace.

- When creating the snapshot, take in a list of namespaces where it can be cloned.
- When creating the clone PVC, also take in the namespace of the snapshot.
- Make sure the PVC is allowed to be cloned in the namespace

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

cehoffman:
This feature while not immediately something we could leverage is something that would be useful to us to enable better permissions control around production and staging/development versions of our services.

PR #100, opened 2018-06-07
Add support to restore snapshots to different namespaces

disrani-px:
- When creating snapshots users need to provide comma seperated regexes
with "stork/snapshot-restore-namespaces" annotaion to specify which
namespaces the snapshot can be restored to
- When creating PVC from snapshots, if a snapshot exists in another
namespace, the snapshot namespace should be specified with
"stork/snapshot-source-namespace" annotation

Issue #71

### DIFF
--- a/pkg/snapshot/snapshotprovisioner.go
+++ b/pkg/snapshot/snapshotprovisioner.go
@@ -1,8 +1,11 @@
 package snapshotcontroller
 
 import (
+	"encoding/csv"
 	"errors"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/kubernetes-incubator/external-storage/lib/controller"
 	crdv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
@@ -18,6 +21,13 @@ import (
 // Most of this has been taken from the kubernetes-incubator snapshot
 // provisioner with some changes. It is a part of the main package there,
 // so can't vendor it in here.
+
+const (
+	storkSnapshotRestoreNamespacesAnnotation = "stork/snapshot-restore-namespaces"
+	// StorkSnapshotSourceNamespaceAnnotation Annotation used to specify the
+	// source of the snapshot when creating a PVC
+	StorkSnapshotSourceNamespaceAnnotation = "stork/snapshot-source-namespace"
+)
 
 type snapshotProvisioner struct {
 	// Kubernetes Client.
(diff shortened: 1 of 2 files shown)


harsh-px:
In the integration tests, can you add a spec dir for snapshots across namespaces?

disrani-px:
The problem is torpedo creates random namespaces, so can't do that right now. Let me check if there is another way

pkg/snapshot/snapshotprovisioner.go
```diff
@@ -19,6 +22,13 @@ import (
 // provisioner with some changes. It is a part of the main package there,
 // so can't vendor it in here.
 
+const (
+	storkSnapshotRestoreNamespacesAnnotation = "stork/snapshot-restore-namespaces"
```
harsh-px: +1 on allowing multiple namespaces for restore.

pkg/snapshot/snapshotprovisioner.go
```diff
@@ -74,6 +84,35 @@ func (p *snapshotProvisioner) snapshotRestore(
...
+	}
+
+	csvReader := csv.NewReader(strings.NewReader(allowedNamespaces))
+	namespaces, err := csvReader.ReadAll()
+	if err != nil {
+		log.Errorf("Error parsing allowed namespaces: %v", allowedNamespaces)
+		return false
+	}
+
+	for _, namespaceRegEx := range namespaces[0] {
```
harsh-px: this will panic if `allowedNamespaces = ""` 

`csvReader.ReadAll` does not return err in this case.

pkg/snapshot/snapshotprovisioner.go
```diff
@@ -74,6 +84,35 @@ func (p *snapshotProvisioner) snapshotRestore(
...
+		regex, err := regexp.Compile(strings.TrimSpace(namespaceRegEx))
+		if err != nil {
+			log.Errorf("Invalid regex for allowed namespaces: %v", namespaceRegEx)
+			return false
+		}
+		if regex.MatchString(namespace) {
```
harsh-px: if I have namespaces `dev1` and `dev11` and if the allowed namespaces is just `dev1`, this will match `dev11` too.

pkg/snapshot/snapshotprovisioner.go
disrani-px: Good catch. Fixed

pkg/snapshot/snapshotprovisioner.go
disrani-px: Added start and end delimiters around each regex which should catch those cases.

disrani-px:
Checked and can't add the integration test right now. Can specify the allowed namespace as a regex bug won't be able to add the source namespace in the PVC spec

PR #102, opened 2018-06-11
Add snapshot scale test

disrani-px:


test/integration_test/snapshot_test.go
```diff
@@ -222,3 +174,73 @@ func parseDataVolumes(
 
 	return dataVolumesNames, dataVolumesInUse
 }
+
+func createSnapshot(t *testing.T) []*scheduler.Context {
+	ctx, err := schedulerDriver.Schedule(generateInstanceID(t, ""),
```
harsh-px: If the `testName` arg to `generateInstanceID` is empty, the generated instance ID will be of format "-<uid>". Let's pass the testname here. Will be also useful in debugging failures since the namespace names will have the test name.

test/integration_test/common_test.go
```diff
@@ -180,3 +183,11 @@ func verifyScheduledNode(t *testing.T, appNode node.Node, volumes []string) {
 	logrus.Infof("Scores: %v", scores)
 	require.Equal(t, highScore, scores[appNode.Name], "Scheduled node does not have the highest score")
 }
+
+func init() {
+	flag.IntVar(&snapshotScaleCount,
+		"snapshot-scale-count",
+		10,
+		"Number of volumes to use for scale snapshot test")
```
harsh-px: nit: this isn't just Number of volumes but rather the scale factor by which apps and volumes both get scaled.

test/integration_test/snapshot_test.go
disrani-px: The namespace is of the format `mysql-snap-group--88602b0b-ae0c-4d4a-8092-7b1350b28ec1`, with the test name in there it hits the length limit usually that's why most of them don't pass in the test name

test/integration_test/common_test.go
disrani-px: Yes, but the purpose is to test the scaling of volumes and snapshots, which is what it is referring to.

PR #103, opened 2018-06-11
Add a mysql locker container to flush mysql and lock the database

harsh-px:


harsh-px:
discussed offline with @jvinod  and @disrani-px . Closing this one to try using a different approach (waiting on a file that acts as a signal to terminate the command).

PR #107, opened 2018-06-27
Add gosimple checker

disrani-px:
Also fix issues found by gosimple

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BIN :=$(BASE_DIR)/bin
 .DEFAULT_GOAL=all
 .PHONY: test clean
 
-all: stork vet lint integration-test
+all: stork vet lint simple integration-test
 
 deps:
 	GO15VENDOREXPERIMENT=0 go get -d -v $(PKGS)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -66,11 +66,15 @@ lint:
 vet:
 	go vet $(PKGS)
 
+simple:
+	go get -u honnef.co/go/tools/cmd/gosimple
+	gosimple $(PKGS)
```
zoxpx: You might want to split out the `go get -u honnef.co/go/tools/cmd/gosimple` into a separate make-rule, e.g.:

```
$(GOPATH)/bin/gosimple:
        go get -u honnef.co/go/tools/cmd/gosimple

simple: $(GOPATH)/bin/gosimple
	$(GOPATH)/bin/gosimple $(PKGS)
```

.. that way you won't keep recompiling/updating the "gosimple" binary (e.g. the "gosimple" utility will get built only when missing).

disrani-px:
Code coverage drop is because of an unnecessary return statement that was removed.

PR #108, opened 2018-06-29
Check node cache with ID and hostname

disrani-px:


drivers/volume/portworx/portworx.go
```diff
@@ -274,12 +274,17 @@ func (p *portworx) mapNodeStatus(status api.Status) storkvolume.NodeStatus {
...
+	obj, exists, err := p.store.GetByKey(nodeInfo.ID)
 	if err != nil {
 		return nil, err
 	} else if !exists {
-		return nil, fmt.Errorf("Node %v not found in cache", hostname)
+		obj, exists, err = p.store.GetByKey(nodeInfo.Hostname)
+		if err != nil {
+			return nil, err
+		} else if !exists {
+			return nil, fmt.Errorf("Node %v not found in cache", nodeInfo.ID)
```
harsh-px: nit: also log `nodeInfo.Hostname` in the error message since both ID and hostname lookups failed.

Issue #112, opened 2018-07-07
Extender: Needs to deal with IP set as hostname in kubernetes node object


disrani-px:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG

**What happened**:
It's possible that kubernetes clusters are setup so that the IP is used as the hostname in the node information. This will cause issues with matching driver nodes to kubernetes nodes

**What you expected to happen**:
The extender should also try to use the node name in that case and match nodes

**How to reproduce it (as minimally and precisely as possible)**:
Set the hostname for the kubernetes node object to an IP instead of a hostname.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

disrani-px:
Fixed with #113

PR #113, opened 2018-07-07
Extender: Also check node name to match driver nodes

disrani-px:
Issue #112

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -124,7 +124,7 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
 	pod := &args.Pod
 	storklog.PodLog(pod).Debugf("Nodes in filter request:")
 	for _, node := range args.Nodes.Items {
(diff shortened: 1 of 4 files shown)


pkg/extender/extender.go
```diff
@@ -141,15 +141,20 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
+								continue
+							}
 
-					for _, driverNode := range driverNodes {
-						if e.isHostnameMatch(driverNode.Hostname, address.Address) &&
-							driverNode.Status == volume.NodeOnline {
-							filteredNodes = append(filteredNodes, node)
+							if e.isHostnameMatch(driverNode.Hostname, address.Address) &&
+								driverNode.Status == volume.NodeOnline {
+								filteredNodes = append(filteredNodes, node)
```
harsh-px: you can break out of the for loop here. No need to process subsequent `node.Status.Addresses`

disrani-px:
Absolute code coverage has increased.

PR #124, opened 2018-07-19
Rename stork.com to stork in stork rules CRD

harsh-px:


pkg/apis/stork/register.go
```diff
@@ -2,7 +2,7 @@ package stork
 
 const (
 	// GroupName is the group name of the CRD
-	GroupName = "stork.com"
```
disrani-px: GroupName should be in domain format, stork.libopenstorage.com

harsh-px: Updated. It's a ghost link and has only symbolic purpose.

disrani-px: It should actually be used to register the CRD. The actual group name used to generate the boilerplate code in doc.go should also be stork.libopenstorage.com

harsh-px: fixed the group name

PR #125, opened 2018-07-20
Update sched-ops vendor so that snapshot validation will continue even if the first validate fails

harsh-px:


disrani-px:
Please use 72 chars or less in the first line of the commit, otherwise, it gets truncated in github.

Details of why it is needed can be in the summary of the message

harsh-px:
> Please use 72 chars or less in the first line of the commit message, otherwise, it gets truncated in github.

Fixed.

PR #133, opened 2018-08-07
Background command fixes

### LINKED ISSUE #131: Portworx: Cloudsnap not completing
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG

**What happened**:
Cloudsnap backup requests aren't completing. It is getting stuck trying to signal a channel for a pre-exec rule even if it none was defined in the spec.

**What you expected to happen**:
Should be able to create cloudsnap without pre-exec rule.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a cloudsnap
2. It will finish in PX successfully but stork will never update status in volumesnapshot object

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

harsh-px:
1) Don't send the background termination channel if there are no background commands
2) Handle situations where there could be more than one background actions and the first one has a single pod

Fixes #131

### DIFF
--- a/pkg/snapshot/rule/rule.go
+++ b/pkg/snapshot/rule/rule.go
@@ -208,7 +208,6 @@ func executeSnapRule(pvcs []v1.PersistentVolumeClaim, snap *crdv1.VolumeSnapshot
 		return nil, err
 	}
 
-	backgroundCommandTermChan := make(chan bool, 1)
 	if snap.Metadata.Annotations != nil {
 		ruleName, present := snap.Metadata.Annotations[ruleAnnotationKeys[rType]]
 		if present && len(ruleName) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/snapshot/rule/rule.go
```diff
@@ -251,19 +258,35 @@ func executeSnapRule(pvcs []v1.PersistentVolumeClaim, snap *crdv1.VolumeSnapshot
...
+						if action.Background {
+							backgroundActionPresent = true
+						}
+
 						if action.Type == apis_stork.StorkRuleActionCommand {
-							err := executeCommandAction(filteredPods, rule, snap, action, backgroundCommandTermChan, rType, taskID)
+							err := executeCommandAction(filteredPods, rule, snap, action, backgroundPodListChan, rType, taskID)
 							if err != nil {
-								return backgroundCommandTermChan, err
+								if backgroundActionPresent {
```
disrani-px: If any action fails we should terminate all the actions. If we are returning an error we shouldn't return any channels, its error prone since callers might just continue if err is set.

harsh-px: done in the incremental

PR #136, opened 2018-08-14
Add option to integration test script to pick docker image name

disrani-px:


test/integration_test/test-deploy.sh
```diff
@@ -33,7 +40,7 @@ fi
 
 KUBEVERSION=$(kubectl version -o json | jq ".serverVersion.gitVersion" -r)
 sed -i 's/<kube_version>/'"$KUBEVERSION"'/g' /specs/stork-scheduler.yaml
-sed -i 's/'stork:.*'/'stork:master'/g' /specs/stork-deployment.yaml
+sed -i 's/'stork:.*'/'"$image_name"'/g' /specs/stork-deployment.yaml
```
harsh-px: `envsubst` is a cleaner way to do spec replacements since it will replace the exact variable instead of a pattern match.

for e.g https://github.com/portworx/pxdocs/blob/master/.travis.yml#L50

disrani-px: The spec doesn't have a variable name unfortunately, it just has the defaults.

PR #138, opened 2018-08-24
 Rename storkrule to rule

### LINKED ISSUE #128: Rename storkrule CRD to just rule
**BUG REPORT**: The group name already has stork. So no need of having storkrule in actual resource name.

harsh-px:
Closes #128 

Once this gets merged, will update sched-ops and torpedo and update vendor.json here.


pkg/apis/stork/register.go
```diff
@@ -2,7 +2,7 @@ package stork
 
 const (
 	// GroupName is the group name of the CRD
-	GroupName = "stork.libopenstorage.com"
+	GroupName = "stork.openstorage.org"
```
disrani-px: I think we should leave that as libopenstorage.org which is the GitHub org name for the project.

PR #140, opened 2018-09-07
Add namespace and other info for logs

disrani-px:


pkg/log/log.go
```diff
@@ -12,9 +12,19 @@ import (
...
 func PodLog(pod *v1.Pod) *logrus.Entry {
 	if pod != nil {
-		return logrus.WithFields(logrus.Fields{
-			"PodName": pod.Name,
-		})
+		fields := logrus.Fields{
+			"PodName":   pod.Name,
+			"Namespace": pod.Namespace,
+		}
+		if pod.OwnerReferences != nil {
```
harsh-px: nit: `if pod.OwnerReferences != nil {` is not needed. `range pod.OwnerReferences` handles nil and empty.

PR #141, opened 2018-09-13
When matching nodes check for internalIP incase hostnames are different

disrani-px:
This is for environments like K8s on DC/OS where the kubelet is running
in a containerized environment and has a different hostname than the physical
node

Need to make some more changes in the integration test to match the node correctly when verifying the scheduled node. Will submit that as a separate change.

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -102,46 +102,6 @@ func (e *Extender) getHostname(node *v1.Node) string {
 	return ""
 }
 
-// There are a couple of things that need to be checked to see if the driver
-// node matched the k8s node since different k8s installs set the node name,
-// hostname and IPs differently
-func (e *Extender) isNodeMatch(k8sNode *v1.Node, driverNode *volume.NodeInfo) bool {
-	if driverNode == nil {
-		return false
-	}
-
-	if e.isHostnameMatch(driverNode.ID, k8sNode.Name) {
-		return true
-	}
-	for _, address := range k8sNode.Status.Addresses {
-		switch address.Type {
-		case v1.NodeHostName:
-			if e.isHostnameMatch(driverNode.Hostname, address.Address) {
-				return true
-			}
-		case v1.NodeInternalIP:
-			for _, ip := range driverNode.IPs {
-				if ip == address.Address {
-					return true
-				}
-			}
-		}
-	}
-	return false
-}
-
-// The driver might not return fully qualified hostnames, so check if the short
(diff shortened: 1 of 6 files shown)


pkg/extender/extender.go
```diff
@@ -219,8 +236,7 @@ func (e *Extender) getNodeScore(
 						for _, rack := range rackInfo.PreferredLocality {
 							if rack == nodeRack || nodeRack == "" {
 								for _, datanode := range volumeInfo.DataNodes {
-									if e.isHostnameMatch(idMap[datanode], address.Address) ||
-										e.isHostnameMatch(idMap[datanode], node.Name) {
+									if e.isNodeMatch(&node, idMap[datanode]) {
```
harsh-px: this can panic in `isNodeMatch` if datanode is not present in idMap.

pkg/extender/extender.go
```diff
@@ -357,10 +363,10 @@ func (e *Extender) processPrioritizeRequest(w http.ResponseWriter, req *http.Req
 			zoneInfo.PreferredLocality = zoneInfo.PreferredLocality[:0]
 			regionInfo.PreferredLocality = regionInfo.PreferredLocality[:0]
 			for _, node := range volume.DataNodes {
-				log.Debugf("ID: %v Hostname: %v", node, idMap[node])
-				regionInfo.PreferredLocality = append(regionInfo.PreferredLocality, regionInfo.HostnameMap[idMap[node]])
-				zoneInfo.PreferredLocality = append(zoneInfo.PreferredLocality, zoneInfo.HostnameMap[idMap[node]])
-				rackInfo.PreferredLocality = append(rackInfo.PreferredLocality, rackInfo.HostnameMap[idMap[node]])
+				log.Debugf("ID: %v Hostname: %v", node, idMap[node].Hostname)
```
harsh-px: If `Driver.GetPodVolumes()` returns volumes which have an outdated data node which is no longer present in Driver.GetNodes(), then we could panic here since `node` won't be present in `idMap`.

disrani-px: Yes, can happen in the rare case if nodes are decommissioned right between those 2 calls. Updated to ignore missing nodes.

pkg/extender/extender.go
disrani-px: Ignoring nil node in isNodeMatch

disrani-px:
Code coverage dropped because the code was refactored.

PR #154, opened 2018-10-06
Check if PV and PVC are owned by driver when migrating

disrani-px:


drivers/volume/portworx/portworx.go
```diff
@@ -334,7 +334,7 @@ func (p *portworx) GetNodes() ([]*storkvolume.NodeInfo, error) {
 	return nodes, nil
 }
 
-func (p *portworx) isPortworxPVC(pvc *v1.PersistentVolumeClaim) bool {
+func (p *portworx) OwnsPVC(pvc *v1.PersistentVolumeClaim) bool {
 	storageClassName := k8shelper.GetPersistentVolumeClaimClass(pvc)
 	if storageClassName == "" {
 		logrus.Debugf("Empty StorageClass in PVC %v", pvc.Name)
```
harsh-px: Not related to this change but in the change below, just because `provisioner == snapshotcontroller.GetProvisionerName() ` does not mean portworx owns the PVC. It just means stork created the cloned PVC.

Instead, the check in `OwnsPVC` is generic enough to be moved out of the volume driver interface. The volume driver should just return it's provisioner name and the cluster migration controller can perform the rest of the checks.

disrani-px: There is a Portworx specific check on line 387

disrani-px:
@harsh-px Pushed some more commits to this branch, can you review those too?

PR #162, opened 2018-10-20
Add storkctl to container and upload to s3

disrani-px:


specs/stork-daemonset.yaml
```diff
@@ -103,7 +58,7 @@ subjects:
   namespace: kube-system
 roleRef:
   kind: ClusterRole
-  name: stork-role
+  name: cluster-admin
```
harsh-px: we want to make stork cluster admin?

disrani-px: For migrating it needs to look at all resources, might as well make it an admin

disrani-px: This is how a lot of system level services do it (helm, ark, to name a new). We already have so many permissions here which is borderline cluster admin anyways.

PR #164, opened 2018-10-22
Add support in stork to generate clusterPair spec file

ram-infrac:
This PR add support in stork-integration test to create on fly cluster-pair spec from destination cluster kubeconfig file. It's assumed that before running integration test configmap is created

  - Create clusterPair spec from remote kubeconfig
  - read and save kubeconfig from configmap
  - parse kubeconfig and create clusterpair spec

Signed-off-by: Ram Suradkar <ram.suradkar@portworx.com>

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -4,11 +4,16 @@ package integrationtest
 
 import (
 	"flag"
+	"fmt"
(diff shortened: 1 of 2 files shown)


ram-infrac:
Outdated

PR #169, opened 2018-10-26
Add command to generate clusterpair spec from destination cluster

disrani-px:


pkg/storkctl/clusterpair.go
```diff
@@ -3,19 +3,22 @@ package storkctl
...
+	"reflect"
 
 	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/portworx/sched-ops/k8s"
 	"github.com/spf13/cobra"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/pkg/printers"
 )
 
+var clusterPairSubcommand = "clusterpair"
```
harsh-px: nit: this can be a const

pkg/storkctl/clusterpair.go
```diff
@@ -73,3 +76,38 @@ func clusterPairPrinter(clusterPairList *storkv1.ClusterPairList, writer io.Writ
...
+					APIVersion: storkv1.SchemeGroupVersion.String(),
+				},
+				ObjectMeta: meta.ObjectMeta{
+					Name: "<insert_name_here>",
+				},
+
+				Spec: storkv1.ClusterPairSpec{
+					Config: config,
+					Options: map[string]string{
+						"<insert_storage_options_here>": "",
```
harsh-px: is the use case to create a template spec which users have to complete?

disrani-px: Yes, the template has the kubeconfig populated, they only need to add the storage options

pkg/storkctl/clusterpair.go
```diff
@@ -73,3 +76,38 @@ func clusterPairPrinter(clusterPairList *storkv1.ClusterPairList, writer io.Writ
...
+				handleError(err)
+			}
+
+			clusterPair := &storkv1.ClusterPair{
+				TypeMeta: meta.TypeMeta{
+					Kind:       reflect.TypeOf(storkv1.ClusterPair{}).Name(),
+					APIVersion: storkv1.SchemeGroupVersion.String(),
+				},
+				ObjectMeta: meta.ObjectMeta{
+					Name: "<insert_name_here>",
```
harsh-px: We can add a prefix based on the `config` here so that one can generate difficult cluster pair specs per context. config has CurrentContext which we could use. for `cluster-pair-ctx-<current-context>`

disrani-px: There needs to be one context per cluster pair. If you use `storkctl generate clusterpair --context` it'll generate it for that context

PR #180, opened 2018-11-06
Add SetOutputFormat() to Factory interface

rodrigc:
This is needed to facilitate tests.


rodrigc:
See https://github.com/libopenstorage/stork/pull/171 for usage

PR #182, opened 2018-11-07
Add gcloud to container and prune info when generating cluster pair

disrani-px:


pkg/storkctl/factory.go
```diff
@@ -68,7 +68,15 @@ func (f *factory) GetConfig() (*rest.Config, error) {
 }
 
 func (f *factory) RawConfig() (clientcmdapi.Config, error) {
-	return f.getKubeconfig().RawConfig()
+	config, err := f.getKubeconfig().RawConfig()
+	if err != nil {
+		return config, err
+	}
+
+	if f.context != "" {
```
piyush-nimbalkar: Can the context be nil? If yes, we will end up pruning everything, right?

piyush-nimbalkar: Yes `f.context`. I meant empty and not nil.

PR #183, opened 2018-11-08
Read files when generating cluster pair and populate inline

disrani-px:


pkg/storkctl/clusterpair.go
```diff
@@ -125,6 +149,36 @@ func newGenerateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptio
 					}
 				}
 			}
+
+			if config.AuthInfos[currentAuthInfo].ClientCertificate != "" && len(config.AuthInfos[currentAuthInfo].ClientCertificateData) == 0 {
```
harsh-px: If `config.AuthInfos[currentAuthInfo]` is nil, this will panic.

disrani-px: The config loader already validates that the config is usable, ie all maps have valid keys, so don't need the check here.

harsh-px: In that case, we can remove the `config.AuthInfos[currentAuthInfo] != nil` check from line 144.

disrani-px: Actually looks like it is required if there are no auth infos at all in the config. Updated.

PR #189, opened 2018-11-09
Make sure namespaces exists before starting migration

disrani-px:
Also don't check for volume migration status if none are being migrated


pkg/snapshot/rule.go
```diff
@@ -51,9 +51,15 @@ func validateSnapRules(snap *crdv1.VolumeSnapshot) error {
 	return nil
 }
 
+func setKind(snap *crdv1.VolumeSnapshot) {
+	snap.Kind = "VolumeSnapshot"
```
harsh-px: VolumeSnapshot objects don't have kind set by default?

disrani-px: From the commit message
```
It doesn't seem to be set always when being passed in from the snapshot
controller.
```

pkg/migration/controllers/migration.go
```diff
@@ -172,40 +193,43 @@ func (m *MigrationController) migrateVolumes(migration *storkv1.Migration) error
...
+					string(vInfo.Status),
+					fmt.Sprintf("Error migrating volume %v: %v", vInfo.Volume, vInfo.Reason))
+				migration.Status.Stage = storkv1.MigrationStageFinal
+				migration.Status.Status = storkv1.MigrationStatusFailed
+			} else if vInfo.Status == storkv1.MigrationStatusSuccessful {
+				m.Recorder.Event(migration,
+					v1.EventTypeNormal,
+					string(vInfo.Status),
+					fmt.Sprintf("Volume %v migrated successfully", vInfo.Volume))
+			}
```
harsh-px: we should collect all progress, failed and successful volumes and then dump events so that they are consistent.

The way it is now, we may or may not see `"Volume %v migrated successfully` events in the migration object if one of the volume is still in progress or failed and it comes before and after the successful volume.

pkg/migration/controllers/migration.go
```diff
@@ -213,10 +214,9 @@ func (m *MigrationController) migrateVolumes(migration *storkv1.Migration) error
 		// Now check if there is any failure or success
 		// TODO: On failure of one volume cancel other migrations?
 		for _, vInfo := range volumeInfos {
-			// Return if we have any volume migrations still in progress
 			if vInfo.Status == storkv1.MigrationStatusInProgress {
 				logrus.Infof("Volume Migration still in progress: %v", migration.Name)
```
harsh-px:
```suggestion
				logrus.Infof("Volume %v migration still in progress", vInfo.Volume)
```

harsh-px: Above will give more use info as it prints each volume instead of repeating the same log with the migration name for multiple vols.

disrani-px: Yeah should have been volumename.

PR #190, opened 2018-11-10
 Add version to storkctl and unittest

disrani-px:


pkg/storkctl/version_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"k8s.io/kubernetes/pkg/kubectl/genericclioptions"
+)
+
+func TestVersion(t *testing.T) {
+	f := NewTestFactory()
+	streams, _, buf, _ := genericclioptions.NewTestIOStreams()
+	cmd := newVersionCommand(f, streams)
+	cmd.SetOutput(buf)
+	cmd.SetArgs([]string{"version"})
+	if err := cmd.Execute(); err != nil {
```
harsh-px: `if err != nil` is not needed. Since that's what `require.NoError` does.

PR #191, opened 2018-11-13
Add tests for creating and deleting snapshots

rodrigc:


pkg/storkctl/snapshot_test.go
```diff
@@ -50,12 +52,17 @@ func testSnapshotsCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []stri
...
-	if err = cmd.Execute(); err != nil {
-		require.NoError(t, err, "Error executing command: %v", cmd)
-	}
 
-	if e, a := expected, buf.String(); e != a {
-		t.Errorf("expected %v, got %v", e, a)
+	err = cmd.Execute()
+
+	if !behaviorOnFatal {
+		if err != nil {
```
disrani-px: No need of the if, require.NoError() checks that

pkg/storkctl/snapshot_test.go
```diff
@@ -50,12 +52,17 @@ func testSnapshotsCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []stri
...
+			require.NoError(t, err, "Error executing command: %v", cmd)
+		}
+
+		if e, a := expected, buf.String(); e != a {
+			t.Errorf("expected [%v], got [%v]", e, a)
```
disrani-px: You can use require.Equal(). Check `https://github.com/libopenstorage/stork/blob/master/pkg/storkctl/version_test.go#L21`

pkg/storkctl/snapshot_test.go
```diff
@@ -15,12 +15,13 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/rest/fake"
+	cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 	"k8s.io/kubernetes/pkg/kubectl/genericclioptions"
 )
 
 type NewTestCommand func(cmdFactory Factory, ioStreams genericclioptions.IOStreams) *cobra.Command
 
-func testSnapshotsCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []string, snapshots *snapv1.VolumeSnapshotList, expected string) {
+func testSnapshotsCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []string, snapshots *snapv1.VolumeSnapshotList, expected string, behaviorOnFatal bool) {
```
disrani-px: I think errorExpected is more suitable for the last parameter

pkg/storkctl/snapshot_test.go
```diff
@@ -95,5 +102,164 @@ func TestGetVolumeSnapshotsOneSnapshot(t *testing.T) {
...
+		})
+		testSnapshotsCommon(t, newCreateCommand, cmdArgs, &snapshots, expected, true)
+	}()
+}
+
+func TestCreateSnapshots(t *testing.T) {
+	cmdArgs := []string{"volumesnapshots", "-p", "pvc_name", "snap1"}
+
+	var snapshots snapv1.VolumeSnapshotList
+	expected := `Snapshot snap1 created successfully
```
disrani-px: Add `\n` in the string instead of the new lines?

pkg/storkctl/snapshot_test.go
```diff
@@ -95,5 +102,164 @@ func TestGetVolumeSnapshotsOneSnapshot(t *testing.T) {
...
+
+`
+	testSnapshotsCommon(t, newCreateCommand, cmdArgs, &snapshots, expected, false)
```
disrani-px: Can be done as a separate PR, but the fake API server should store the object and when we do a get we should get the same object back to confirm it was created correctly.

pkg/storkctl/snapshot_test.go
```diff
@@ -95,5 +102,164 @@ func TestGetVolumeSnapshotsOneSnapshot(t *testing.T) {
...
+			if calledFatal {
+				return
+			}
+			calledFatal = true
+			require.Equal(t, 1, code, "Unexpected error code")
+			require.Equal(t, expected, e)
+		})
+		testSnapshotsCommon(t, newDeleteCommand, cmdArgs, &snapshots, expected, true)
+	}()
+}
```
disrani-px: Need to assert that fatal callback was called
`require.True(t, calledFatal, msg)`

Same in other tests where this is being done.

pkg/storkctl/snapshot_test.go
rodrigc: Fixed

pkg/storkctl/snapshot_test.go
rodrigc: Fixed

pkg/storkctl/snapshot_test.go
rodrigc: Fixed

pkg/storkctl/snapshot_test.go
rodrigc: Fixed

pkg/storkctl/snapshot_test.go
rodrigc: Fixed: https://github.com/libopenstorage/stork/pull/191/files#diff-f2f14386f17b4772c682fe79cfd98da5R69

PR #195, opened 2018-11-13
Add clusterpair tests

rodrigc:


pkg/storkctl/snapshot_test.go
```diff
@@ -21,29 +25,42 @@ import (
 
 type NewTestCommand func(cmdFactory Factory, ioStreams genericclioptions.IOStreams) *cobra.Command
 
-func testSnapshotsCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []string, snapshots *snapv1.VolumeSnapshotList, expected string, errorExpected bool) {
+func testCommon(t *testing.T, newCommand NewTestCommand, cmdArgs []string, obj runtime.Object, expected string, errorExpected bool) {
```
disrani-px: Move this to common_test.go

pkg/storkctl/snapshot_test.go
```diff
@@ -21,29 +25,42 @@ import (
...
 	var err error
-
+	typeName := reflect.ValueOf(obj).Elem().Type().Name()
+
+	var addToScheme func(*runtime.Scheme) error
+	var schemeGroupVersion schema.GroupVersion
+	if typeName == "VolumeSnapshotList" {
+		addToScheme = snapv1.AddToScheme
```
disrani-px: I think we can call addToSchem() for all the crds. You can do that in an init() function in common_test.go

pkg/storkctl/snapshot_test.go
rodrigc: Done

pkg/storkctl/snapshot_test.go
rodrigc: OK, I refactored into an `init()` method:
https://github.com/libopenstorage/stork/pull/195/commits/8e1cc53f8c5e855b96ec05d6a00c79bdc7988651#diff-a7665dbba9e98c4f01c65f3725c7bb93R26

pkg/storkctl/migration_test.go
```diff
@@ -0,0 +1,70 @@
...
+			IncludeResources:  true,
+			StartApplications: true,
+		},
+	}
+	migration.Name = "migration1"
+
+	var migrationList storkv1.MigrationList
+	migrationList.Items = append(migrationList.Items, *migration)
+
+	expected := "No resources found.\n"
```
disrani-px: This should return the above migration, right?

pkg/storkctl/clusterpair_test.go
```diff
@@ -0,0 +1,49 @@
...
+func TestGenerateClusterPair(t *testing.T) {
+	cmdArgs := []string{"clusterpair", "pair1"}
+
+	var clusterPairs storkv1.ClusterPairList
+	clusterPair := &storkv1.ClusterPair{
+		TypeMeta: meta.TypeMeta{
+			Kind:       reflect.TypeOf(storkv1.ClusterPair{}).Name(),
+			APIVersion: storkv1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: meta.ObjectMeta{},
```
disrani-px: Set the name here

pkg/storkctl/clusterpair_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+		Spec: storkv1.ClusterPairSpec{
+			Options: map[string]string{},
+		},
+	}
+
+	clusterPairs.Items = append(clusterPairs.Items, *clusterPair)
+
+	expected := ""
```
disrani-px: This should expect the table output, right?

pkg/storkctl/clusterpair_test.go
rodrigc: Fixed

pkg/storkctl/clusterpair_test.go
rodrigc: I think in my current test `config.Contexts[currentContext]` is nil, so it skips over a bunch of stuff in `newGenerateClusterPairCommand` and doesn't emit any output.  This is not ideal, but I couldn't figure out how to populate the currentContext

pkg/storkctl/migration_test.go
rodrigc: This is where I was getting stuck trying to figure out how to use the `Fake`.  In the debugger, I was hitting this code:

```
=>  51:	func (c *FakeMigrations) List(opts v1.ListOptions) (result *v1alpha1.MigrationList, err error) {
    52:		obj, err := c.Fake.
    53:			Invokes(testing.NewRootListAction(migrationsResource, migrationsKind, opts), &v1alpha1.MigrationList{})
```

in `github.com/libopenstorage/stork/pkg/client/clientset/versioned/typed/stork/v1alpha1/fake/fake_migration.go`

That seems to return an empty migration list.  I couldn't figure out how to prepopulate that.

PR #202, opened 2018-11-22
Change clusterpair and migration to be namespaced

disrani-px:
Limit migration to namespace of migration object

TODO: Add parameter to set one namespace that can migrate all namespaces


harsh-px:
Aren't pairings cluster scoped? What is the use case of having a pair only in a namespace? Since both ends of a pair are entire clusters, they have crossed the namespace boundary already.

disrani-px:
Pairings shouldn't be cluster scoped. The kubeconfig in the clusterpair can have limited scope (ie access to some namepsaces on the remote cluster), so each user can create their own cluster pair. Users also shouldn't be able to see clusterpairs created in other namespaces.
There is a usecase where the pair is created by a cluster admin. That is the TODO I've added in the PR. Admins will be able to specify one namespace that will be able to migrate any other namespace.

harsh-px:
> Pairings shouldn't be cluster scoped. The kubeconfig in the clusterpair can have limited scope (ie access to some namepsaces on the remote cluster), so each user can create their own cluster pair. Users also shouldn't be able to see clusterpairs created in other namespaces.
There is a usecase where the pair is created by a cluster admin. That is the TODO I've added in the PR. Admins will be able to specify one namespace that will be able to migrate any other namespace.


If I have a cluster pair in dev namespace, does the current implementation allow me to migrate apps and pvcs from prod namespace using that pair?

PR #203, opened 2018-11-22
 Add some more UTs for storkctl

disrani-px:


pkg/storkctl/clusterpair.go
```diff
@@ -159,20 +164,23 @@ func newGenerateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptio
...
 							util.CheckErr(err)
+							return
 						}
 						config.AuthInfos[currentAuthInfo].ClientCertificate = ""
 					}
 					if config.AuthInfos[currentAuthInfo].ClientKey != "" && len(config.AuthInfos[currentAuthInfo].ClientKeyData) == 0 {
 						config.AuthInfos[currentAuthInfo].ClientKeyData, err = getByteData(config.AuthInfos[currentAuthInfo].ClientKey)
 						if err != nil {
 							util.CheckErr(err)
+							return
```
rodrigc: Yeah, I found I needed to add return statements after `util.CheckErr()`, because in "test" mode, `util.CheckErr()` does not call `os.Exit()` and keeps on going.

PR #204, opened 2018-11-27
When logging pod info check for owner pointer before dereferencing

disrani-px:


pkg/log/log.go
```diff
@@ -20,7 +20,7 @@ func PodLog(pod *v1.Pod) *logrus.Entry {
 			"Namespace": pod.Namespace,
 		}
 		for _, owner := range pod.OwnerReferences {
-			if *owner.Controller {
+			if owner.Controller != nil && *owner.Controller {
```
harsh-px: This should be 

```suggestion
			if owner != nil && *owner.Controller {
```

disrani-px: Owner is not a pointer

PR #206, opened 2018-11-28
UTs for storkctl migration subcommand

disrani-px:


pkg/log/log_test.go
```diff
@@ -5,12 +5,22 @@ package log
...
+	appv1 "k8s.io/api/apps/v1"
+	appv1beta1 "k8s.io/api/apps/v1beta1"
+	appv1beta2 "k8s.io/api/apps/v1beta2"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestLog(t *testing.T) {
 	t.Run("podLogTest", podLogTest)
+	t.Run("deploymentLogTest", deploymentLogTest)
```
rodrigc: Oh, nice use of golang 1.7 subtests: https://blog.golang.org/subtests

I've been working a lot with px-test, which was started with an older version of go, so doesn't use these techniques for running a suite of tests.

PR #207, opened 2018-11-29
 Allow configuring an admin namespace that can migrate all other namespaces

disrani-px:
Also add rule CRD register which was removed in code refactor

### DIFF
--- a/pkg/rule/rule.go
+++ b/pkg/rule/rule.go
@@ -4,11 +4,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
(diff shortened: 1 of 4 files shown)


pkg/rule/rule.go
```diff
@@ -81,6 +88,32 @@ var ownerAPICallBackoff = wait.Backoff{
 	Steps:    20,
 }
 
+// Init initializes the rule executor
+func Init() error {
```
piyush-nimbalkar: Who calls this Init?

disrani-px: cmd/stork/stork.go

PR #210, opened 2018-11-30
Add namespace to clusterpair and migration specs

ram-infrac:
Jenkins job is still running on it, added it to get review till then

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -258,8 +258,9 @@ func createClusterPair(pairInfo map[string]string) error {
 	}
 	defer pairFile.Close()
 
-	cmd := storkctl.NewCommand(storkctl.NewFactory(), os.Stdin, pairFile, os.Stderr)
-	cmd.SetArgs([]string{"generate", "clusterpair", "--kubeconfig", remoteFilePath})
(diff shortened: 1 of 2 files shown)


test/integration_test/common_test.go
```diff
@@ -258,8 +258,9 @@ func createClusterPair(pairInfo map[string]string) error {
 	}
 	defer pairFile.Close()
 
-	cmd := storkctl.NewCommand(storkctl.NewFactory(), os.Stdin, pairFile, os.Stderr)
-	cmd.SetArgs([]string{"generate", "clusterpair", "--kubeconfig", remoteFilePath})
+	factory := storkctl.NewFactory()
+	cmd := storkctl.NewCommand(factory, os.Stdin, pairFile, os.Stderr)
+	cmd.SetArgs([]string{"generate", "clusterpair", "--kubeconfig", remoteFilePath, "-n", "mysql-1-pvc-singlemysql"})
```
disrani-px: You also need to pass the name to the command otherwise it'll fail. You can remove the lines below where you are adding the name to the spec.

ram-infrac: yes, since your PR is merged. Let me rebase it

PR #218, opened 2018-12-12
Migration tests

ram-infrac:
This PR add support for migrating postgress, wordpress and elasticsearch apps.

To add new app in existing test for migration follow -
* Create 3 folder for app migration
1. Folder with app specs named as ```<appname>``` (specs must be of format that torpedo support for eg. see https://github.com/portworx/torpedo/tree/master/drivers/scheduler/k8s/specs )
   ```e.g ./wordpress``` 
2. Empty folder with name``` <appname>-cluster-pair```
3. Migration spec folder ```<appname>-migration```
   ```eg. wordpress-migration```

This should contain migration specs with following format -
 **Templates:**
```
kind: Migration
metadata:
  name: <appname>-migration
  namespace: <appname>-integration-test
spec:
  # This should be the name of the cluster pair created above
  clusterPair: remoteclusterpair-<appname>-integration-test
  # If set to false this will migrate only the Portworx volumes. No PVCs, apps, etc will be migrated
  includeResources: true
  # If set to false, the deployments and stateful set replicas will be set to 0 on the destination. There will be an annotation with "stork.openstorage.org/migrationReplicas" to store the replica count from the source
  startApplications: true
  # List of namespaces to migrate
  namespaces:
  - <appname>-integration-test 
```
**e.g**  Template for **wordpress** - 
```
kind: Migration
metadata:
  name: wordpress-migration
  namespace: wordpress-integration-test
spec:
  # This should be the name of the cluster pair created above
  clusterPair: remoteclusterpair-wordpress-integration-test
  # If set to false this will migrate only the Portworx volumes. No PVCs, apps, etc will be migrated
  includeResources: true
  # If set to false, the deployments and stateful set replicas will be set to 0 on the destination. There will be an annotation with "stork.openstorage.org/migrationReplicas" to store the replica count from the source
  startApplications: true
  # List of namespaces to migrate
  namespaces:
  - wordpress-integration-test 
```
**Note** : You are free to change ```includeResources, startApplications``` fields

* Add app Name to migrateAppTest in ```test/integration_tests/common.go ```
e.g
   - var migrateAppTest = []string{"postgres", "elasticsearch", "wordpress", "mysql-1-pvc"}


harsh-px:
@ram-infrac / @disrani-px I'm guessing this PR can be closed since we already have merged into migration tests.

PR #222, opened 2018-12-21
 Add group volume snapshot CRD

harsh-px:
This change adds CRD and controller for Group local and cloud snapshots including 3D snaps.

This version also deprecates group snapshots using legacy volumesnapshots. Hence specs have been removed from integration tests.

Integration tests will be added for GroupVolumeSnapshots once this and https://github.com/portworx/sched-ops/pull/98 are merged.


pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +213,66 @@ type MigrationList struct {
...
+type GroupVolumeSnapshotList struct {
+	meta.TypeMeta `json:",inline"`
+	meta.ListMeta `json:"metadata,omitempty"`
+
+	Items []GroupVolumeSnapshot `json:"items"`
+}
+
+// GroupVolumeSnapshotStatus is status for the group snapshot
+type GroupVolumeSnapshotStatus struct {
+	Status          GroupVolumeSnapshotStatusType `json:"status"`
```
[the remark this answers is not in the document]
harsh-px: done

pkg/apis/stork/register.go
```diff
@@ -3,6 +3,4 @@ package stork
 const (
 	// GroupName is the group name of the CRD
 	GroupName = "stork.libopenstorage.org"
-	// Version is the version of the CRD
-	Version = "v1alpha1"
```
disrani-px: Why remove this?

pkg/migration/controllers/clusterpair.go
```diff
@@ -8,11 +8,11 @@ import (
 
 	"github.com/libopenstorage/stork/drivers/volume"
 	stork "github.com/libopenstorage/stork/pkg/apis/stork"
-	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
+	storkv1alpha1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
```
disrani-px: Let's not use alpha1 in the import name. If we update the version we'll need to update the references everywhere.

pkg/apis/stork/v1beta1/doc.go
```diff
@@ -0,0 +1,5 @@
+// +k8s:deepcopy-gen=package,register
```
disrani-px: Why is this beta1? Let's use the convention of starting these as alpha1, we can update them all later when it has gone through some releases.

pkg/apis/stork/register.go
harsh-px: Moved it inside pkg/apis/stork/v1alpha1. It's in the `SchemeGroupVersion` now.

It was better to keep the actual version inside the version package since when we have multiple versions, `pkg/apis/stork/register.go` will be generic.

pkg/migration/controllers/clusterpair.go
harsh-px: done. Renamed the import to stork_api

pkg/apis/stork/v1beta1/doc.go
harsh-px: v1beta1 should not be present in this PR. Removed the pkg.

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +213,81 @@ type MigrationList struct {
...
+	meta.ListMeta `json:"metadata,omitempty"`
+
+	Items []GroupVolumeSnapshot `json:"items"`
+}
+
+// GroupVolumeSnapshotStatus is status for the group snapshot
+type GroupVolumeSnapshotStatus struct {
+	Stage           GroupVolumeSnapshotStageType  `json:"stage"`
+	Status          GroupVolumeSnapshotStatusType `json:"status"`
+	VolumeSnapshots []*crdv1.VolumeSnapshot       `json:"volumeSnapshots"`
```
disrani-px: I don't think this needs the entire CRD here, just the name of the snapshots and their status should suffice.

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,616 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
+)
+
+const (
+	validateCRDInterval time.Duration = 5 * time.Second
+	validateCRDTimeout  time.Duration = 1 * time.Minute
+	resyncPeriod                      = 60 * time.Second
+
+	updateCRD     = true
```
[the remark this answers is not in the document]
harsh-px: it was done to make the returns more readable. Made change to just have one variable.

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,616 @@
...
+					backgroundChannel <- true
+					delete(m.bgChannelsForRules, snapUID)
+				}
+			}
+		case stork_api.GroupSnapshotStagePostSnapshot:
+			updateCRDForThisEvent, err = m.handlePostSnap(groupSnapshot)
+		case stork_api.GroupSnapshotStageFinal:
+			return m.handleFinal(groupSnapshot)
+		default:
+			log.GroupSnapshotLog(groupSnapshot).Errorf("Invalid stage for groupSnapshot: %v", groupSnapshot.Status.Stage)
```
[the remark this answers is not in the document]
harsh-px: done

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,616 @@
...
+		if updateErr != nil {
+			return updateErr
+		}
+
+		// Since we updated, bump the minimum resource version
+		// This is needed since the resync period can overlap with the time a handle
+		// event is already being processed. In such situation, the operator framework
+		// with provide a groupSnapshot which is the same version as the previous groupSnapshot
+		// If we reprocess an outdated object, this can throw off the status checks in the snapshot stage
+		m.minResourceVersion = groupSnapshot.ResourceVersion
```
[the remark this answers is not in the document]
harsh-px: I verified this by printing the actual objects between 2 updates. This was happening very regularly when the sync period was 15 seconds.

First print showed the I was updating the snapshots in the status of the group snapshot using sdk.Update().
Second print showed the new handle event coming in at the same timestamp. The groupSnapshot object in the event did not have the snapshots in status and had the same Metadata.ResourceVersion.

drivers/volume/portworx/portworx.go
```diff
@@ -1666,6 +1244,293 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
...
+// createGroupCloudSnapFromVolumes creates cloud group snapshots
+func (p *portworx) createGroupCloudSnapFromVolumes(
+	groupSnap *stork_crd.GroupVolumeSnapshot,
+	volNames []string,
+	options map[string]string) (
+	*storkvolume.GroupSnapshotCreateResponse, error) {
+
+	credID := getCredIDFromSnapshot(groupSnap.Spec.Options)
+
+	resp, err := p.volDriver.CloudBackupGroupCreate(&api.CloudBackupGroupCreateRequest{
```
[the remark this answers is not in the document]
harsh-px: You mean pass a task ID to the OSD API to allow it to resume from a previously triggered backup?

cmd/stork/stork.go
```diff
@@ -251,6 +252,14 @@ func runStork(d volume.Driver, recorder record.EventRecorder, c *cli.Context) {
 		if err := snapshotController.Start(); err != nil {
 			log.Fatalf("Error starting snapshot controller: %v", err)
 		}
+
+		groupsnapshotInst := groupsnapshot.GroupSnapshot{
+			Driver:   d,
+			Recorder: recorder,
+		}
+		if err := groupsnapshotInst.Init(); err != nil {
+			log.Fatalf("Error initializing groupsnapshot controller due to: %v", err)
```
disrani-px: Remove `due to`, keep it consistent with the other messages above.

drivers/volume/mock/mock.go
```diff
@@ -10,7 +10,7 @@ import (
 	storkvolume "github.com/libopenstorage/stork/drivers/volume"
 	"github.com/libopenstorage/stork/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
disrani-px: Not required

drivers/volume/volume.go
```diff
@@ -9,7 +9,7 @@ import (
 	stork_crd "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
disrani-px: Not required

drivers/volume/portworx/portworx.go
```diff
@@ -1666,6 +1295,311 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
...
+	*storkvolume.GroupSnapshotCreateResponse, error) {
+	snapType, err := getSnapshotType(snap.Spec.Options)
+	if err != nil {
+		return nil, err
+	}
+	switch snapType {
+	case crdv1.PortworxSnapshotTypeCloud:
+		return p.getGroupCloudSnapStatus(snap)
+	case crdv1.PortworxSnapshotTypeLocal:
+		return nil, fmt.Errorf("status API not supported for local group snapshots")
```
disrani-px: This can't return an error. The caller wouldn't know how to deal with a legit error as opposed to this error.

drivers/volume/portworx/portworx.go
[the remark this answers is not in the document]
disrani-px: Yes, we should pass in the same taskID for the same cloudsnap when we resume. The response should return an appropriate error if the same taskID has already been triggered and we can deal with it here.

drivers/volume/portworx/portworx.go
```diff
@@ -1666,6 +1295,311 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
...
+				SnapshotCloudCredID: credID,
+			},
+		}
+
+		var conditions []crdv1.VolumeSnapshotCondition
+
+		if csStatus.status == api.CloudBackupStatusDone {
+			conditions = getReadySnapshotConditions()
+			doneTasks = append(doneTasks, taskID)
+			snapIDsPendingRevert = append(snapIDsPendingRevert, csStatus.cloudSnapID)
```
disrani-px: Call this `snapIDsDone`. It's confusing to say that these are pending revert here when they are successful.

drivers/volume/portworx/portworx.go
```diff
@@ -1666,6 +1295,311 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
...
+		} else if csStatus.status != api.CloudBackupStatusNotStarted {
+			conditions = getPendingSnapshotConditions(csStatus.msg)
+			activeTasks = append(activeTasks, taskID)
```
disrani-px: What happens to these active/pending cloudsnaps? They need to be canceled and cleaned up too on failure

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,94 @@ type MigrationList struct {
...
+
+// GroupVolumeSnapshotStatus is status for the group snapshot
+type GroupVolumeSnapshotStatus struct {
+	Stage           GroupVolumeSnapshotStageType  `json:"stage"`
+	Status          GroupVolumeSnapshotStatusType `json:"status"`
+	VolumeSnapshots []*VolumeSnapshotStatus       `json:"volumeSnapshots"`
+}
+
+// VolumeSnapshotStatus captures the status of a volume snapshot operation
+type VolumeSnapshotStatus struct {
```
disrani-px: We should add the source PVC name here too.

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+	if err != nil {
+		return fmt.Errorf("Error getting apiextention client, %v", err)
+	}
+
+	err = m.createCRD()
+	if err != nil {
+		return err
+	}
+
+	discoveryClient := aeclient.Discovery()
```
disrani-px: This through line 94 isn't required

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,94 @@ type MigrationList struct {
...
+)
+
+// GroupVolumeSnapshotStageType is the stage of the group snapshot
+type GroupVolumeSnapshotStageType string
+
+const (
+	// GroupSnapshotStageInitial is when the group snapshot is just created
+	GroupSnapshotStageInitial GroupVolumeSnapshotStageType = ""
+	// GroupSnapshotStagePrechecks is when the group snapshot is going through prechecks
+	GroupSnapshotStagePrechecks GroupVolumeSnapshotStageType = "Prechecks"
```
disrani-px: PreChecks

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+func (m *GroupSnapshotController) handleSnap(groupSnap *stork_api.GroupVolumeSnapshot) (bool, error) {
+	var (
+		err      error
+		stage    stork_api.GroupVolumeSnapshotStageType
+		status   stork_api.GroupVolumeSnapshotStatusType
+		response *volume.GroupSnapshotCreateResponse
+	)
+
+	if len(groupSnap.Status.VolumeSnapshots) > 0 {
+		logrus.Infof("Group snapshot already active. Checking status")
```
disrani-px: Log with GroupSnapshotLog

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+		response, err = m.Driver.GetGroupSnapshotStatus(groupSnap)
+		if err != nil {
+			logrus.Errorf("group snapshot status returned err: %v", err)
```
disrani-px: Log with GroupSnapshotLog

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+			m.Recorder.Event(groupSnap,
+				v1.EventTypeWarning,
+				string(stork_api.GroupSnapshotFailed),
+				err.Error())
+
+			return !updateCRD, err
+		}
+	} else {
+		logrus.Infof("Creating new group snapshot")
```
disrani-px: Log with GroupSnapshotLog

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+		status   stork_api.GroupVolumeSnapshotStatusType
+		response *volume.GroupSnapshotCreateResponse
+	)
+
+	if len(groupSnap.Status.VolumeSnapshots) > 0 {
+		logrus.Infof("Group snapshot already active. Checking status")
+		response, err = m.Driver.GetGroupSnapshotStatus(groupSnap)
+		if err != nil {
+			logrus.Errorf("group snapshot status returned err: %v", err)
+			m.Recorder.Event(groupSnap,
```
disrani-px: This will log multiple events? Handle() is also recording events when this function returns an error.

Same for other places in this function where event is being recorded.

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -0,0 +1,615 @@
...
+
+func (m *GroupSnapshotController) handleFinal(groupSnap *stork_api.GroupVolumeSnapshot) error {
+	// currently nothing to do
+	return nil
+}
+
+func (m *GroupSnapshotController) handleDelete(groupSnap *stork_api.GroupVolumeSnapshot) error {
+	err := m.Driver.DeleteGroupSnapshot(groupSnap)
+	if err != nil {
+		m.Recorder.Event(groupSnap,
```
disrani-px: I think the object is gone at this time, so the event won't really show up.

pkg/log/log.go
```diff
@@ -7,7 +7,7 @@ import (
 	appv1 "k8s.io/api/apps/v1"
 	appv1beta1 "k8s.io/api/apps/v1beta1"
 	appv1beta2 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
disrani-px: Not required

test/integration_test/snapshot_test.go
```diff
@@ -33,132 +30,6 @@ func simpleSnapshotTest(t *testing.T) {
 	destroyAndWait(t, ctx)
 }
 
-func groupSnapshotTest(t *testing.T) {
```
disrani-px: We need to have similar tests for the new groupsnapshot CRD

disrani-px:
Also need to add sub commands to storkctl

harsh-px:
> Also need to add sub commands to storkctl

Created https://github.com/libopenstorage/stork/issues/226 for this. storkctl support was absent for prior groupsnaps too. We are already too late for the 2.0.2 release. So out of scope for new changes.

harsh-px:
Have addressed the review comments.

specs/groupvolumesnapshots/cassandra-e2e/cassandra-restore-pvcs.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
```
disrani-px: specs is for the deployments. Can you move these under /examples so that they are easier to find.

drivers/volume/portworx/portworx.go
disrani-px: My point was it's confusing that we are saying that completed cloudsnaps need to be reverted. The decision that it needs to be reverted is being made somewhere else. This should just keep a track of snaps that are done and the variable name should reflect that.

harsh-px:
Addressed last 2 comments.

drivers/volume/portworx/portworx.go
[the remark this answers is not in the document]
disrani-px: Can you file a bug to track this?

drivers/volume/portworx/portworx.go
harsh-px: Done

https://github.com/libopenstorage/stork/issues/229
https://portworx.atlassian.net/browse/PWX-7385

drivers/volume/portworx/portworx.go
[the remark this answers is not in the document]
harsh-px: Done

https://github.com/libopenstorage/stork/issues/229
https://portworx.atlassian.net/browse/PWX-7385

specs/groupvolumesnapshots/cassandra-e2e/cassandra-restore-pvcs.yaml
harsh-px: moved

PR #223, opened 2019-01-03
Add a StorageCluster CRD to manage cluster

piyush-nimbalkar:
- Added a controller which does nothing as of now except for watching over StorageCluster object
- The controller starts by default with stork, can be disabled using params

Signed-off-by: Piyush Nimbalkar <piyush@portworx.com>


pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+	Env []core.EnvVar `json:"env"`
+	// CSIEndpoint is the csi endpoint for the driver
+	CSIEndpoint string `json:"csiEndpoint"`
+	// CallHome send cluster information for analytics
+	CallHome bool `json:"callHome"`
+	// RuntimeOpts is a map of options with extra configs for storage driver
+	RuntimeOpts map[string]string `json:"runtimeOptions"`
+	// Nodes node level configurations that will override the ones at cluster
+	// level. These configurations can be grouped based on label selectors.
+	Nodes []NodeSpec `json:"nodes"`
```
harsh-px: This is missing a placement spec that we will use for the daemonset. By default, we use the following that install.portworx.com returns. But we will need users to override that else we get stuck on an install which requires custom placement.

```
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: px/enabled
                operator: NotIn
                values:
                - "false"
              - key: node-role.kubernetes.io/master
                operator: DoesNotExist
```

https://rook.io/docs/rook/v0.5/cluster-crd.html#placement-configuration-settings is a good reference for placement spec.

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+	// Internal flag indicates whether to use internal kvdb or an external one
+	Internal bool `json:"internal"`
+	// Endpoints to access the kvdb
+	Endpoints []string `json:"endpoints"`
+	// BasicAuthSecret is name of kubernetes secret containing the username
+	// and password for basic auth
+	BasicAuthSecret string `json:"accessSecret"`
+	// CertificateSecret is name of the kubernetes secret containing the
+	// cert files required for kvdb authentication
+	CertificateSecret string `json:"certificateSecret"`
```
harsh-px: Do we assume the certificate filenames?

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+}
+
+// NodeStatus status of the storage cluster node
+type NodeStatus struct {
+	// NodeName name of the node
+	NodeName string `json:"nodeName"`
+	// NodeUUID uuid of the node
+	NodeUUID string `json:"nodeUuid"`
+	// Status status of the node
+	Status string `json:"status"`
```
harsh-px: Status needs to be an enum. Have a StatusMessage for arbitrary string status descriptions.

harsh-px: For node conditions, I prefer how k8s does it.

https://github.com/portworx/porx/blob/master/vendor/k8s.io/api/core/v1/types.go#L3950

A node status can have a list of conditions, each with it's own state and msg. Hence a PX node can express multiple conditions (e.g Storage down, Control plane up, Maintenance mode etc.) at the same time.

A single status will be limiting.

cmd/stork/stork.go
```diff
@@ -107,12 +108,17 @@ func main() {
 			Name:  "migration-admin-namespace",
 			Usage: "Namespace to be used by a cluster admin which can migrate all other namespaces (default: none)",
 		},
+		cli.BoolTFlag{
+			Name:  "cluster-controller",
```
disrani-px: Should we call it storage controller? Cluster is too generic

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+	// Kvdb is the information of kvdb that storage driver uses
+	Kvdb KvdbSpec `json:"kvdb"`
+	// Network is the network information for storage driver
+	Network NetworkSpec `json:"network"`
+	// Storage details of storage used by the driver
+	Storage StorageSpec `json:"storage"`
+	// CloudStorage details of storage in cloud environment.
+	// Keep this only at cluster level for now until support to change
+	// cloud storage at level is added.
+	CloudStorage CloudStorageSpec `json:"cloudStorage"`
```
disrani-px: Should CloudStorage be a part of the StorageSpec? StorageSpec should take care of whether storage is from local disks or cloud disks

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+	// UseAll use all available, unformatted, unpartioned devices.
+	// This will be ignored if Devices is not empty.
+	UseAll bool `json:"useAll"`
+	// UseAllWithPartitions use all available unformatted devices
+	// including partitions. This will be ignored if Devices is not empty.
+	UseAllWithPartitions bool `json:"useAllWithPartitions"`
+	// Devices list of devices to be used by storage driver
+	Devices []string `json:"devices"`
+	// StorageMdDevices metadata device for storage pools
+	StorageMdDevices []string `json:"storageMetadataDevices"`
```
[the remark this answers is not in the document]
piyush-nimbalkar: Had talked to @jvinod, he said we need it.

pkg/client/clientset/versioned/typed/stork/v1alpha1/storagecluster.go
```diff
@@ -0,0 +1,174 @@
...
+		ns:     namespace,
+	}
+}
+
+// Get takes name of the storageCluster, and returns the corresponding storageCluster object, and an error if there is any.
+func (c *storageClusters) Get(name string, options v1.GetOptions) (result *v1alpha1.StorageCluster, err error) {
+	result = &v1alpha1.StorageCluster{}
+	err = c.client.Get().
+		Namespace(c.ns).
+		Resource("storageclusters").
```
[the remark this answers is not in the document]
piyush-nimbalkar: This is autogenerated.

pkg/cluster/controller.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// createCRD creates the CRD for StorageCluster object
+func (c *Controller) createCRD() error {
+	resource := k8s.CustomResource{
+		Name:    storkv1.StorageClusterResourceName,
+		Plural:  storkv1.StorageClusterResourcePlural,
+		Group:   stork.GroupName,
+		Version: stork.Version,
+		Scope:   apiextensionsv1beta1.ClusterScoped,
```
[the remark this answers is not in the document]
piyush-nimbalkar: `sc` is already being used by storage-class. I can probably use `stc`

pkg/apis/stork/v1alpha1/types.go
piyush-nimbalkar: Yes. We can doc it.

disrani-px: That should be on the implementation to figure out from the secret.

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -212,3 +217,193 @@ type MigrationList struct {
...
+	// JournalDevice device for journaling
+	JournalDevice string `json:"journalDevice"`
+	// SystemMdDevice device that will be used to store system metadata
+	SystemMdDevice string `json:"systemMetadataDevice"`
+	// DataStorageType type of data storage used
+	DataStorageType string `json:"dataStorageType"`
```
disrani-px: Needs more info on what this means.

pkg/client/clientset/versioned/typed/stork/v1alpha1/storagecluster.go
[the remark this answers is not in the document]
disrani-px: @adityadani  This is generated code.

pkg/apis/stork/v1alpha1/types.go
piyush-nimbalkar: How would we know what is the key that is used for client cert or CA cert? 
By default when you create a k8s secret it  takes the name of the file as key. We need to document and make sure people use a known names for different secret keys.

cmd/stork/stork.go
piyush-nimbalkar: Storage controller assumes all storage operations.
I will call it storage-cluster-controller.

pkg/apis/stork/v1alpha1/types.go
piyush-nimbalkar: Had consciously separated them because there are different options in both them that don't make sense when combined with the other.
Like max_storage_nodes or max_storage_nodes_per_zone in storage spec,
and use_all, use_all_with_partitions in cloud spec.
Would be overall easier for people to understand what they are doing. They will specify only one of them.

pkg/apis/stork/v1alpha1/types.go
piyush-nimbalkar: @jvinod Do we need to expose this field to users? If it's for internal testing, can we just put it in RuntimeOpts and add necessary flags in oci mon?

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -308,3 +315,253 @@ const (
...
+	// including partitions. This will be ignored if Devices is not empty.
+	UseAllWithPartitions *bool `json:"useAllWithPartitions"`
+	// Devices list of devices to be used by storage driver
+	Devices *[]string `json:"devices"`
+	// JournalDevice device for journaling
+	JournalDevice *string `json:"journalDevice"`
+	// SystemMdDevice device that will be used to store system metadata
+	SystemMdDevice *string `json:"systemMetadataDevice"`
+	// DataStorageType backing store type for managing drives and pools
+	DataStorageType *string `json:"dataStorageType"`
```
disrani-px: We should enums for Types

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -308,3 +315,253 @@ const (
...
+	// cloud storage at node level is added.
+	CloudStorage *CloudStorageSpec `json:"cloudStorage"`
+	// SecretsProvider is the name of secret provider that driver will connect to
+	SecretsProvider string `json:"secretsProvider"`
+	// CSIEndpoint is the csi endpoint for the driver
+	CSIEndpoint string `json:"csiEndpoint"`
+	// CallHome send cluster information for analytics
+	CallHome bool `json:"callHome"`
+	// CommonSpec that is present at both cluster and node level
+	CommonSpec
```
disrani-px: Don't need to call everything a spec. Only things with status need a spec.
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status

pkg/apis/stork/v1alpha1/types.go
piyush-nimbalkar: These cannot be well defined enums if we have to keep the CRD provider agnostic. An example of data storage type would be `pxv`. Every storage provider can come have their custom data store types.

disrani-px:
Can you remove the internal PWX numbers from the commits

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -308,3 +315,253 @@ const (
...
+
+// ConditionStatus is the enum type for node condition statuses
+type ConditionStatus string
+
+// These are valid statuses of different node conditions.
+const (
+	// NodeOnline means the node condition is online and healthy
+	NodeOnline ConditionStatus = "Online"
+	// NodeInit means the node condition is in intializing state
+	NodeInit ConditionStatus = "Intializing"
```
disrani-px: `Initializing`

pkg/apis/stork/v1alpha1/types.go
```diff
@@ -308,3 +315,253 @@ const (
...
+	StorageState NodeConditionType = "StorageState"
+)
+
+// ConditionStatus is the enum type for node condition statuses
+type ConditionStatus string
+
+// These are valid statuses of different node conditions.
+const (
+	// NodeOnline means the node condition is online and healthy
+	NodeOnline ConditionStatus = "Online"
```
disrani-px: Can be made in a subsequent change but these should be called `ConditionOnline` etc.
Also need more Statuses like `StorageFull`, `DiskError`, etc

PR #227, opened 2019-01-14
Update Rule CRD spec field

disrani-px:
Spec should point to a struct and shouldn't be an array type

This is a breaking change and will be listed in the Release Note.


pkg/apis/stork/v1alpha1/types.go
```diff
@@ -34,7 +34,12 @@ const (
 type Rule struct {
 	meta.TypeMeta   `json:",inline"`
 	meta.ObjectMeta `json:"metadata,omitempty"`
-	Spec            []RuleItem `json:"spec"`
+	Spec            RuleSpec `json:"spec"`
```
harsh-px: How will this work with existing 2.0.2 users who already have the old rule spec?

disrani-px: It's a breaking change like mentioned above, so the old rules won't work. But we need to fix this sooner than later.

disrani-px: The field shouldn't be called a `spec` then. For all objects that have `spec` it refers to another structure. For things like ClusterRole, the field is called `rules`.

Either way, the CRD needs to be fixed.

harsh-px: In that case, let's merge this into 2.0.2 since we don't have anyone using rules in prod now.

https://github.com/libopenstorage/stork/tree/master/examples/groupvolumesnapshots need to be updated with this change.

disrani-px: Actually going to release the groupsnapshot as 2.1 since there is a new CRD. This change will go in that

disrani-px:
Updated examples. Removed the example with the error case.

PR #232, opened 2019-01-16
 Fixes for rule recovery

harsh-px:
Fixes the following

1. Pods running rule commands were not being tracked in snapshot annotations
2. Rule recovery was not getting invoked for group snapshots
3. Existing pods in tracker were getting overridden by new pods


Gopkg.lock
```diff
@@ -571,30 +571,44 @@
...
   ]
   pruneopts = "UT"
-  revision = "51b66a2b633767dac6b89254a6dd745e2f3390af"
+  revision = "d64ebc777d0c67885c842bc7f03a4c131cf38bfa"
 
 [[projects]]
   branch = "master"
-  digest = "1:cdff4b38209b651b78fb982e5cbd9ca0520ea41e6f3464139ae6a8876ffdfa05"
+  digest = "1:cfe8d419eb553560ba1e297dc0430a161be3239372a9fd60111e63da3b75fdb8"
+  name = "github.com/portworx/talisman"
```
disrani-px: Why is this dependency required?

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -205,24 +208,25 @@ func (m *GroupSnapshotController) handleInitial(groupSnap *stork_api.GroupVolume
...
 		groupSnap.Status.Stage = stork_api.GroupSnapshotStageSnapshot
-		return updateCRD, nil
+		return groupSnap, updateCRD, nil
 	}
 
 	log.GroupSnapshotLog(groupSnap).Infof("Running pre-snapshot rule: %s", ruleName)
 	r, err := k8s.Instance().GetRule(ruleName, groupSnap.Namespace)
 	if err != nil {
-		return !updateCRD, err
+		return groupSnap, !updateCRD, err
```
disrani-px: This isn't idiomatic golang. If err is set the other return values shouldn't be used by the caller.

pkg/rule/rule.go
```diff
@@ -59,21 +59,21 @@ const (
...
 	err error
 }
 
-// commandTask tracks pods where commands for a taskID might still be running
-type commandTask struct {
-	taskID string
-	pods   []pod
+// CommandTask tracks pods where commands for a taskID might still be running
+type CommandTask struct {
+	TaskID string `json:"task_id"`
```
disrani-px: taskID for the json tag to keep in line with k8s tags

pkg/rule/rule.go
```diff
@@ -150,46 +150,31 @@ func terminateCommandInPods(owner runtime.Object, pods []v1.Pod, taskID string)
...
-					}
-
-					log.RuleLog(nil, owner).Warnf("Failed to get pod with uid: %s due to: %v", pod.uid, err)
-					continue // best effort
+				if err == k8s.ErrPodsNotFound {
+					continue
 				}
 
-				backgroundPodList = append(backgroundPodList, *p)
+				log.RuleLog(nil, owner).Warnf("Failed to get pod with uid: %s due to: %v", pod.UID, err)
```
disrani-px: Raise an event here.

pkg/rule/rule.go
```diff
@@ -316,12 +301,48 @@ func executeCommandAction(
...
+					if err == k8s.ErrPodsNotFound {
+						continue
+					}
+
+					return err
+				}
+
+				// Check duplicates
+				skipPod := false
+				for _, pod := range podsForAction {
```
disrani-px: Store the pods in a map with  UID->*podInfo so that you don't need this loop within a loop. After the loop put it in an array.

pkg/groupsnapshot/controllers/groupsnapshot.go
harsh-px: fixed

pkg/rule/rule.go
harsh-px: done

pkg/rule/rule.go
harsh-px: done

Gopkg.lock
harsh-px: sched-ops master has this in it's vendor. So dep pulled it here too

pkg/rule/rule.go
harsh-px: done

harsh-px:
Addressed the comments

pkg/rule/rule.go
disrani-px: I don't see the event?

PR #250, opened 2019-02-13
 Add CRDs for SchedulePolicy and MigrationSchedule

disrani-px:


pkg/apis/stork/v1alpha1/schedulepolicy.go
```diff
@@ -0,0 +1,181 @@
...
+const (
+	// SchedulePolicyResourceName is name for "schedulepolicy" resource
+	SchedulePolicyResourceName = "schedulepolicy"
+	// SchedulePolicyResourcePlural is plural for "schedulepolicy" resource
+	SchedulePolicyResourcePlural = "schedulepolicies"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
```
disrani-px: Going to make this non-namespaced since this is something that should be created by admins and used by other users similar to a storageclass

pkg/apis/stork/v1alpha1/schedulepolicy.go
```diff
@@ -0,0 +1,182 @@
...
+	// the specified time
+	Weekly *WeeklyPolicy `json:"weekly"`
+	// Monthly policy that will be triggered on the specified date of the month
+	// at the specified time
+	Monthly *MonthlyPolicy `json:"monthly"`
+}
+
+// IntervalPolicy contains the interval at which an action should be triggered
+type IntervalPolicy struct {
+	IntervalMinutes int `json:"intervalMinutes"`
```
harsh-px: How do I set an interval of every 15 seconds?

pkg/apis/stork/v1alpha1/schedulepolicy.go
```diff
@@ -0,0 +1,182 @@
...
+	if _, present := Days[w.Day]; !present {
+		return fmt.Errorf("Invalid day of the week (%v) in Weekly policy", w.Day)
+	}
+	return nil
+}
+
+// MonthlyPolicy contains the date and time in a month when an action should be
+// executed
+type MonthlyPolicy struct {
+	Date int    `json:"date"`
```
harsh-px: Instead of `Date`, this should be called `Day`.  Cron follows this naming scheme: http://www.nncron.ru/help/EN/working/cron-format.htm

pkg/apis/stork/v1alpha1/schedulepolicy.go
disrani-px: You can't, that's why it explicitly mentions Minutes. I don't see a use case of running something at a granularity lower than a minute.

pkg/apis/stork/v1alpha1/schedulepolicy.go
disrani-px: Day or Date isn't part of Cron's scheme as such, they don't use any names.

The problem with using Day here is that it becomes confusing for a user to figure out that Day here is date of the month but in the WeeklyPolicy it is Day of the week.

pkg/apis/stork/v1alpha1/schedulepolicy.go
harsh-px: Sure. If there is a need, we can add a seconds field later on.

pkg/apis/stork/v1alpha1/schedulepolicy.go
```diff
@@ -0,0 +1,182 @@
...
+func (i *IntervalPolicy) Validate() error {
+	if i.IntervalMinutes < 1 {
+		return fmt.Errorf("Invalid intervalMinutes (%v) in Interval policy", i.IntervalMinutes)
+	}
+	return nil
+}
+
+// DailyPolicy contains the time in the day where an action should be executed
+type DailyPolicy struct {
+	Time string `json:"time"`
```
harsh-px: What format does `Time` follow? Please add that to the comment.

disrani-px: Done. Added comments for the other members too.

PR #252, opened 2019-02-13
Add controller for migration schedule

disrani-px:
- The MigrationSchedule object takes the same parameters as Migration Spec
- It also takes in a SchedulePolicyName
- The reconciler checks every minute if any new migrations need to be triggered
- The status for each migration is stored in the object. Only one successful
status is stored
- Only one migration can be triggered at a time
- The schedule package is used to check is a migration should be triggered for
each type of policy.
- Includes UTs for the schedule package to make sure triggers will fire
correctly

Added subcommands to storkctl for SchedulePolicy and MigrationSchedule
- Added create, get and delete for MigrationSchedule
- Added get for SchedulePolicy
- Added UTs for both resources

Output for schedulePolicy:
```
ubuntu@kb1:~$ storkctl get schedulepolicy
NAME      INTERVAL-MINUTES   DAILY     WEEKLY             MONTHLY
daily     10                 7:41PM    Thursday@10:17PM   14@8:05PM
```

Output for migrationSchedule:
```
ubuntu@kb1:~$ storkctl get migrationschedules -n mysql
NAME                     POLICYNAME   CLUSTERPAIR     LAST-SUCCESS-TIME
mysqlmigrationschedule   daily        remotecluster   14 Feb 19 22:32 UTC
```


pkg/apis/stork/v1alpha1/migrationschedule.go
```diff
@@ -13,8 +13,14 @@ const (
 
 // MigrationScheduleSpec is the spec used to schedule migrations
 type MigrationScheduleSpec struct {
-	MigrationSpec      `json:",inline"`
-	SchedulePolicyName string `json:"schedulePolicyName"`
+	Template           MigrationTemplateSpec `json:"template"`
```
harsh-px: any reason we need the additional template field?

disrani-px: As mentioned in the commit message
```
Update MigrationSchedule CRD to add Migration under Template.Spec

This is similar to how deployments, statefulsets, etc have pod spec defined
```

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func (m *MigrationScheduleController) shouldStartMigration(migrationSchedule *stork_api.MigrationSchedule) (stork_api.SchedulePolicyType, bool, error) {
+	// Don't trigger a new migration if one is already in progress
+	for _, policyType := range stork_api.GetValidSchedulePolicyTypes() {
+		policyMigration, present := migrationSchedule.Status.Items[policyType]
+		if present {
+			for _, migration := range policyMigration {
+				if !m.isMigrationComplete(migration.Status) {
+					return stork_api.SchedulePolicyTypeInvalid, false, nil
```
harsh-px: Why should this return SchedulePolicyTypeInvalid?

We found that a migration is not yet complete for `policyType`. So it should be

```
return policyType, false, nil
```

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return sdk.Update(migrationSchedule)
+}
+
+func (m *MigrationScheduleController) deleteMigrations(migrationSchedule *stork_api.MigrationSchedule) error {
+	for _, policyMigration := range migrationSchedule.Status.Items {
+		for _, migration := range policyMigration {
+			err := k8s.Instance().DeleteMigration(migration.Name, migrationSchedule.Namespace)
+			if err != nil {
+				log.MigrationScheduleLog(migrationSchedule).Warnf("Error deleting %v: %v", migration.Name, err)
```
harsh-px: save atleast the last non-nil error and return at the end.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func (m *MigrationScheduleController) createCRD() error {
+	resource := k8s.CustomResource{
+		Name:    stork_api.MigrationScheduleResourceName,
+		Plural:  stork_api.MigrationScheduleResourcePlural,
```
harsh-px: You could pass in `ShortNames` as `mis`, `migsched` etc.

pkg/migration/controllers/migrationschedule.go
disrani-px: The purpose of the function is to check if a migration should be started. If yes, then it should also return the policy type that needs to be triggered. It doesn't return the type of policy that has been triggered. I'll add comments to the function to make it clear.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+
+				if pendingMigration.Status.Status == stork_api.MigrationStatusInitial {
+					pendingMigration.Status.Status = stork_api.MigrationStatusPending
+				}
+
+				// Check again and update the status if it is completed
+				migration.Status = pendingMigration.Status.Status
+				if m.isMigrationComplete(migration.Status) {
+					migration.FinishTimestamp = meta.Now()
+					if pendingMigration.Status.Status == stork_api.MigrationStatusSuccessful {
```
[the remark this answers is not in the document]
disrani-px: It is being treated as a failed case. I've updated the event to print the status so it can be reported.

pkg/migration/controllers/migrationschedule.go
disrani-px: Short names need to be intuitive, otherwise, no one will use them.

pkg/migration/controllers/migrationschedule.go
disrani-px: Returning the last error after Ignoring "NotFound" errors

harsh-px:
Also update https://github.com/libopenstorage/stork/blob/master/specs/stork-deployment.yaml#L58 with permissions for migration schedule

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+		for _, migration := range policyMigration {
+			// Get the updated status it we see it as not completed
+			if !m.isMigrationComplete(migration.Status) {
+				pendingMigration, err := k8s.Instance().GetMigration(migration.Name, migrationSchedule.Namespace)
+				if err != nil {
+					m.Recorder.Event(migrationSchedule,
+						v1.EventTypeWarning,
+						string(stork_api.MigrationStatusFailed),
+						fmt.Sprintf("Error getting status of migration %v: %v", migration.Name, err))
+					pendingMigration.Status.Status = stork_api.MigrationStatusFailed
```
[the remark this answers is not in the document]
disrani-px: Fixed

pkg/migration/controllers/migrationschedule.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+	return nil
+}
+
+func (m *MigrationScheduleController) updateMigrationStatus(migrationSchedule *stork_api.MigrationSchedule) error {
+	updated := false
+	for _, policyMigration := range migrationSchedule.Status.Items {
+		for _, migration := range policyMigration {
+			// Get the updated status it we see it as not completed
```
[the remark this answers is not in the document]
disrani-px: Fixed a typo `it`->`if`

specs/stork-deployment.yaml
```diff
@@ -76,8 +76,8 @@ rules:
     resources: ["rules"]
     verbs: ["get", "list"]
   - apiGroups: ["stork.libopenstorage.org"]
-    resources: ["clusterpairs", "migrations", "groupvolumesnapshots", "storageclusters"]
-    verbs: ["get", "list", "watch", "update", "patch"]
+    resources: ["clusterpairs", "migrations", "groupvolumesnapshots", "storageclusters", "schedulepolicies", "migrationschedules"]
```
harsh-px: While writing the integration test, I also run into more missing permissions in the stork-role. configmaps, services, secrets, namespaces will need list permissions. 
```
  Warning  Failed      2m                stork  Error migrating resources: configmaps is forbidden: User "system:serviceaccount:kube-system:stork-account" cannot list configmaps in the namespace "mysql-1-pvc-mysql-migration"
  Warning  Failed      1m                stork  Error migrating resources: secrets is forbidden: User "system:serviceaccount:kube-system:stork-account" cannot list secrets in the namespace "mysql-1-pvc-mysql-migration"
  Warning  Failed      13s (x3 over 1m)  stork  Error migrating resources: services is forbidden: User "system:serviceaccount:kube-system:stork-account" cannot list services in the namespace "mysql-1-pvc-mysql-migration"
```

harsh-px: Actually I was missing the below in my stork-role. This is already present on install.portworx.com.

```
  - apiGroups: ["*"]
    resources: ["*"]
    verbs: ["list", "get"]
```

PR #253, opened 2019-02-13
propagate group snap annotations to child volumesnapshots

harsh-px:
This is required so that we can use snapshots across namespaces ( https://docs.portworx.com/portworx-install-with-kubernetes/storage-operations/create-snapshots/snaps-single-pvc/#creating-snapshots-across-namespaces)


harsh-px:
> Add this as a field in the GroupSnapshot spec instead of having it as an annotation there. If the field is set set the annotation in the individual snapshots

done

harsh-px:
> Please squash into one commit

done

PR #254, opened 2019-02-13
 Add --all-namespaces parameter to storkctl for get subcommands

disrani-px:


disrani-px:
`codecov/patch` is low because snapshot APIs don't have the mocks generated from the external project, so can't write UTs for some of that code. Validated that manually.

pkg/storkctl/common_test.go
```diff
@@ -46,8 +46,9 @@ func resetTest() {
 	}
 	testFactory = NewTestFactory()
 	testFactory.setOutputFormat(outputFormatTable)
-	tf := testFactory.TestFactory.WithNamespace("test")
-	testFactory.setNamespace("test")
+	tf := testFactory.TestFactory
+	//tf := testFactory.TestFactory.WithNamespace("test")
+	//testFactory.setNamespace("test")
```
harsh-px: commented code can be removed

disrani-px: Removed

PR #259, opened 2019-02-15
Return status for failed cloudsnapshots

harsh-px:
If any cloudsnap has failed, the get status API should not fail. Rather it should return the failed tasks in the response so controller can log an event and then reset and retry the group snapshot.

The PR also handles all failed cloud snapshot statues (stopped and aborted).

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -761,7 +761,7 @@ func (p *portworx) DescribeSnapshot(snapshotData *crdv1.VolumeSnapshotData) (*[]
 		}
 
 		csStatus := p.getCloudSnapStatus(api.CloudBackupOp, pv.Spec.PortworxVolume.VolumeID)
-		if csStatus.status == api.CloudBackupStatusFailed {
+		if isCloudsnapStatusFailed(csStatus.status) {
 			err = fmt.Errorf(csStatus.msg)
 			return getErrorSnapshotConditions(err), false, err
 		}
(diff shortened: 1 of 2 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1518,7 +1508,17 @@ func (p *portworx) generateStatusReponseFromTaskIDs(taskIDs []string, credID str
 
 		var conditions []crdv1.VolumeSnapshotCondition
 
-		if csStatus.status == api.CloudBackupStatusDone {
+		if isCloudsnapStatusFailed(csStatus.status) {
+			logrus.Errorf(csStatus.msg)
```
disrani-px: Pass in the groupsnapshot object to `generateStatusReponseFromTaskIDs()` and log with `GroupSnapshotLog()` so that there is more context in the log.

harsh-px: done

pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -305,8 +305,14 @@ func (m *GroupSnapshotController) handleSnap(groupSnap *stork_api.GroupVolumeSna
 		return !updateCRD, err
 	}
 
-	if isAnySnapshotFailed(response.Snapshots) {
-		log.GroupSnapshotLog(groupSnap).Infof("Some snapshots in group have failed")
+	if isFailed, failedTasks := isAnySnapshotFailed(response.Snapshots); isFailed {
+		err = fmt.Errorf("Some snapshots in group have failed: %s."+
+			" Resetting group snapshot to retry.", failedTasks)
+		log.GroupSnapshotLog(groupSnap).Infof(err.Error())
```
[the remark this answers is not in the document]
harsh-px: done

PR #261, opened 2019-02-15
For groupsnap scale test, first create all and then verify

harsh-px:

### DIFF
--- a/test/integration_test/snapshot_test.go
+++ b/test/integration_test/snapshot_test.go
@@ -114,6 +114,10 @@ func groupSnapshotTest(t *testing.T) {
 	ctxs := createGroupsnaps(t)
 	ctxsToDestroy = append(ctxsToDestroy, ctxs...)
 
+	for _, ctx := range ctxs {
+		verifyGroupSnapshot(t, ctx, defaultWaitTimeout)
+	}
+
 	// Negative
 	ctxs, err := schedulerDriver.Schedule(generateInstanceID(t, ""),
 		scheduler.ScheduleOptions{AppKeys: []string{"mysql-snap-group-fail"}})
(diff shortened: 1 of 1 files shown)


test/integration_test/snapshot_test.go
```diff
@@ -134,13 +138,18 @@ func groupSnapshotTest(t *testing.T) {
 }
 
 func groupSnapshotScaleTest(t *testing.T) {
-	ctxsToDestroy := make([]*scheduler.Context, 0)
+	alllContexts := make([]*scheduler.Context, 0)
```
disrani-px: `allContexts`

test/integration_test/snapshot_test.go
```diff
@@ -134,13 +138,18 @@ func groupSnapshotTest(t *testing.T) {
...
 	// Triggers 2 snaps, so use half the count in the loop
 	for i := 0; i < snapshotScaleCount/2; i++ {
 		ctxs := createGroupsnaps(t)
-		ctxsToDestroy = append(ctxsToDestroy, ctxs...)
+		alllContexts = append(alllContexts, ctxs...)
 	}
-	destroyAndWait(t, ctxsToDestroy)
+
+	for _, ctx := range alllContexts {
+		verifyGroupSnapshot(t, ctx, defaultWaitTimeout)
```
disrani-px: The timeout needs to be increased if there are a large number of snapshots, similar to `snapshotScaleTest`

harsh-px: done I'm also doing that when it's equal to 10 since I just had a run where the default timeout wasn't sufficient for scale factor = 10

test/integration_test/snapshot_test.go
harsh-px: done

PR #266, opened 2019-02-21
Add support for configurable retry count

harsh-px:
The PR adds a new field maxRetries in the group volume snapshot spec that allows users to set a max retry count if a group snapshot fails. 

By default, the max retries is 0 and hence the controller will mark the group snapshot as failed if it's fails after it was triggered.

```release-note
Don't retry by default if a group snapshot fails after it is triggered. Allows users to set max retry count.
```

### DIFF
--- a/pkg/groupsnapshot/controllers/groupsnapshot.go
+++ b/pkg/groupsnapshot/controllers/groupsnapshot.go
@@ -328,16 +328,34 @@ func (m *GroupSnapshotController) handleSnap(groupSnap *stork_api.GroupVolumeSna
 	}
 
 	if isFailed, failedTasks := isAnySnapshotFailed(response.Snapshots); isFailed {
-		err = fmt.Errorf("Some snapshots in group have failed: %s."+
-			" Resetting group snapshot to retry.", failedTasks)
+		errMsgPrefix := fmt.Sprintf("Some snapshots in group have failed: %s", failedTasks)
+
+		if groupSnap.Status.NumRetries < groupSnap.Spec.MaxRetries {
+			groupSnap.Status.NumRetries++
+
+			err = fmt.Errorf("%s. Resetting group snapshot for retry: %d",
+				errMsgPrefix, groupSnap.Status.NumRetries)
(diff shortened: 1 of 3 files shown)


pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -328,16 +328,34 @@ func (m *GroupSnapshotController) handleSnap(groupSnap *stork_api.GroupVolumeSna
...
-		err = fmt.Errorf("Some snapshots in group have failed: %s."+
-			" Resetting group snapshot to retry.", failedTasks)
+		errMsgPrefix := fmt.Sprintf("Some snapshots in group have failed: %s", failedTasks)
+
+		if groupSnap.Status.NumRetries < groupSnap.Spec.MaxRetries {
+			groupSnap.Status.NumRetries++
+
+			err = fmt.Errorf("%s. Resetting group snapshot for retry: %d",
+				errMsgPrefix, groupSnap.Status.NumRetries)
+			response.Snapshots = nil // so that snapshots are retried
```
disrani-px: This is not enough. If the storage driver is still backing up some of the snapshots then those need to be canceled and deleted.

harsh-px: That's already being done in the driver: https://github.com/libopenstorage/stork/tree/master/https://github.com/libopenstorage/stork/blob/master/drivers/volume/portworx/portworx.go#L1513

This PR doesn't change that behavior and just disables the default retries. `response.Snapshots` was getting set to nil before this PR when the controller was always retrying.

disrani-px: Ok, then that should be moved to the controller here from the driver. Status() APIs should be read-only, they shouldn't modify any state. We should add an API to the volume driver interface to cancel group snapthots that should be called from the controller in that case. 

Doesn't need to be done with this change. Can you file a bug for 2.1.1 to fix this?

harsh-px: #268 created for that.

PR #272, opened 2019-02-28
Add support to use labels to select resources during migration

disrani-px:
**What type of PR is this?**
feature

Adds support to use label selectors during migration

The label selector field present in the CRD is implemented

```release-note
Added support to use label selectors during migration
```

No

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -18,28 +18,41 @@ func testMigration(t *testing.T) {
 	t.Run("disallowedNamespaceTest", migrationDisallowedNamespaceTest)
 	t.Run("failingPreExecRuleTest", migrationFailingPreExecRuleTest)
 	t.Run("failingPostExecRuleTest", migrationFailingPostExecRuleTest)
+	t.Run("labelSelectorTest", migrationLabelSelectorTest)
 }
 
 func triggerMigrationTest(
 	t *testing.T,
 	instanceID string,
 	appKey string,
+	additionalAppKeys []string,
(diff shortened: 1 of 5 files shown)


test/integration_test/migration_test.go
```diff
@@ -55,22 +68,30 @@ func triggerMigrationTest(
...
 
-		// wait on cluster 2 to get mysql pod running
+		// wait on cluster 2 for the app to be running
 		err = setRemoteConfig(remoteFilePath)
 		require.NoError(t, err, "Error setting remote config")
 		err = schedulerDriver.WaitForRunning(preMigrationCtx, defaultWaitTimeout, defaultWaitInterval)
 		require.NoError(t, err, "Error waiting for pod to get to running state on remote cluster after migration")
-		// destroy mysql app on cluster 2
+		if !migrateAllAppsExpected {
+			err = schedulerDriver.WaitForRunning(allAppsCtx, defaultWaitTimeout/2, defaultWaitInterval)
```
harsh-px: This assumes that `WaitForRunning` will fail because of absence of the apps which shouldn't be migrated. But we could have a case where the apps are migrated but are crashing. In that case, the test will not catch the migration issue.

We should rather check specifically for absence of the app which doesn't match the label selector. If that's too specific, we can even parse the returned err for a not found error.

disrani-px: Yeah torpedo isn't flexible enough to do that. I'm actually going to use sched-ops package directly in these tests to get more flexibility. But that is future work.

PR #288, opened 2019-03-12
Update bool in migration specs to pointers so that we can set defaults

disrani-px:
**What type of PR is this?**
Bug

The `includeVolumes` field was defaulting to `false` if not specified in the spec

Yes

No

Yes, 2.1

### DIFF
--- a/pkg/storkctl/migration.go
+++ b/pkg/storkctl/migration.go
@@ -26,6 +26,7 @@ func newCreateMigrationCommand(cmdFactory Factory, ioStreams genericclioptions.I
 	var startApplications bool
 	var preExecRule string
 	var postExecRule string
+	var includeVolumes bool
 
 	createMigrationCommand := &cobra.Command{
 		Use:     migrationSubcommand,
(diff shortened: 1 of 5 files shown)


pkg/storkctl/migration.go
```diff
@@ -70,6 +72,7 @@ func newCreateMigrationCommand(cmdFactory Factory, ioStreams genericclioptions.I
 	createMigrationCommand.Flags().StringSliceVarP(&namespaceList, "namespaces", "", nil, "Comma separated list of namespaces to migrate")
 	createMigrationCommand.Flags().StringVarP(&clusterPair, "clusterPair", "c", "", "ClusterPair name for migration")
 	createMigrationCommand.Flags().BoolVarP(&includeResources, "includeResources", "r", true, "Include resources in the migration")
+	createMigrationCommand.Flags().BoolVarP(&includeVolumes, "includeVolumes", "r", true, "Include volumees in the migration")
 	createMigrationCommand.Flags().BoolVarP(&startApplications, "startApplications", "a", true, "Start applications on the destination cluster after migration")
```
adityadani: startApplications default value should be false here

disrani-px: For manually triggered the default is true, for scheduled migrations it is false.

PR #292, opened 2019-03-15
[Portworx] Don't rely on storage class to determine ownersip of PVC

disrani-px:
For statically created PVCs there might not be a storage class or annotations
with the provisioner. Looks at the volume source in that case to figure out
the owner

**What type of PR is this?**
Bug

To check for PVC ownership the Portworx volume driver queries the storage class to figure out who provisioned the PVC. This causes issues when PVCs are manually bounded or if the storage class is deleted

No

```release-note
[Portworx] Fixed checked for ownership for statically created PVCs
```

Yes, 2.1

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -64,6 +64,8 @@ const (
 
 	// pvcProvisionerAnnotation is the annotation on PVC which has the provisioner name
 	pvcProvisionerAnnotation = "volume.beta.kubernetes.io/storage-provisioner"
+	// pvProvisionedByAnnotation is the annotation on PV which has the provisioner name
+	pvProvisionedByAnnotation = "pv.kubernetes.io/provisioned-by"
 
 	// pvcNameLabel is the key of the label used to store the PVC name
 	pvcNameLabel = "pvc"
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -360,10 +362,26 @@ func (p *portworx) OwnsPVC(pvc *v1.PersistentVolumeClaim) bool {
 	} else {
 		storageClass, err := k8s.Instance().GetStorageClass(storageClassName)
 		if err != nil {
```
harsh-px: For pre-provisioned volumes created using https://docs.portworx.com/portworx-install-with-kubernetes/storage-operations/create-pvcs/using-preprovisioned-volumes/#using-the-portworx-volume

Wouldn't `OwnsPVC` returns from line 354 itself?

I have a statically provisioned PVC using PX that doesn't have annotations and a storage class. It only has pv.Spec.PortworxVolume.

disrani-px: I was reproing similar to what a user had hit where they were populating a random UUID for the storageclass. Will fix for the above case too

disrani-px: Updated

PR #296, opened 2019-03-19
Convert from dep to go modules

rodrigc:
**What type of PR is this?**
cleanup

The following Portworx projects use go modules:

https://github.com/portworx/sched-ops/blob/master/go.mod
https://github.com/portworx/torpedo/blob/master/go.mod
https://github.com/portworx/spawn/blob/master/go.mod
Kubernetes 1.15 and higher use go modules:

https://github.com/kubernetes/kubernetes/blob/release-1.15/go.mod
It will be easier to interoperate between this projects and take vendor imports of Kubernetes if libopenstorage/stork also uses go modules.

No

Yes

No


disrani-px:
Will take a look after branching master for 2.2.0

chrboe:
Ping – any updates here?

chrboe:
Another ping... I feel like this would really ease development

@disrani-px?

disrani-px:
@chrboe With the various dependency requirements during 2.4.0 it was a little difficult to switch to this. We'll try to do this in 2.5.0, or earlier if we get cycles

briangordon:
Hi, I'm arriving via @rodrigc's [comment](https://github.com/golang/go/issues/30515#issuecomment-478727300) on golang/go#30515.

I'm in the same boat as you, trying to write a Makefile that's able to install some tool dependencies without producing noisy changes to `go.mod`. These are what I concluded are the best available options given what's currently supported:

* If you want everyone to use a specific version of the tools then you can pin the versions in `go.mod` but then your Makefile should use `go install` rather than `go get` to install them, because `go get` won't respect what you have in `go.mod`. In fact, it's worse - `go get` will *overwrite* your `go.mod` with the latest version and people will be constantly committing changes to it. `go install` will respect the versions that are in `go.mod`.
* If you don't care what specific version of the tools that people use, then you can create a temporary file (e.g. `$(shell mktemp -d)/temp.mod`) and supply it in the -modfile argument to `go get`. That way some other mod file gets clobbered instead of the one under version control.

Hope this helps.

PR #307, opened 2019-03-22
Allow configuring portworx service in stork

adityadani:

**What type of PR is this?**
>bug

- Get the service name and namespace from env variables.
- Find the ports which PX uses from kubernetes service.

No

```release-note
- Portworx's service name and namespace are configurable through environment variables to stork.
- Added support for Stork to work with Portworx running on custom ports configuration.
```

No

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -4,6 +4,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"math"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1138,7 +1188,7 @@ func (p *portworx) CreatePair(pair *stork_crd.ClusterPair) (string, error) {
 		return "", err
 	}
 
-	port := uint64(apiPort)
+	port := uint64(p.restPort)
```
disrani-px: This is for the destination cluster, so it should be set to `defaultAPIPort`, it can be overwritten from the clusterpair spec below.

PR #313, opened 2019-03-27
Migrate additional resources

disrani-px:
**What type of PR is this?**
Feature

Adds support for migrating additional types of resources

No

```release-note
Added support to migrate following resources 
* DaemonSets
* ServiceAccounts (except default)
* ClusterRoles (if used in the namespace)
* ClusterRoleBindings (if used in the namespace)
```

No

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -534,7 +534,11 @@ func resourceToBeMigrated(migration *stork_api.Migration, resource metav1.APIRes
 		"StatefulSet",
 		"ConfigMap",
 		"Service",
-		"Secret":
+		"Secret",
+		"DaemonSet",
+		"ServiceAccount",
+		"ClusterRole",
+		"ClusterRoleBinding":
 		return true
 	default:
 		return false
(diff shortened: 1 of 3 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -631,12 +635,39 @@ func (m *MigrationController) objectToBeMigrated(
 			return false, nil
 		}
 		return true, nil
-	case "Secret":
```
piyush-nimbalkar: Why is secret object removed from migration?

piyush-nimbalkar:
Where do we migrate daemon set?

pkg/migration/controllers/migration.go
disrani-px: It's not removed. This function just returns whether an object should be migrated. Previously we were not returning secrets for secret accounts. By removing this it returns `true,nil` below for all secrets.

disrani-px:
@piyush-nimbalkar The check is in `resourceToBeMigrated()` that now returns true for DaemonSets. `objectToBeMigrated()` wil always return true for all DaemonSets

PR #302, opened 2019-03-20
Add authentication to Stork

pault84:
Signed-off-by: Paul <paul@portworx.com>

**What type of PR is this?**
feature

**What this PR does / why we need it**:
This PR adds Authentication support to STORK for the OSD SDK

**Does this PR change a user-facing CRD or CLI?**:
no

**Is a release note needed?**:
yes

```release-note
Authentication support added, will read user-context from CRD annotations
```

**Does this change need to be cherry-picked to a release branch?**:
yes
2.1

PR #324, opened 2019-04-01
Authorization support for Portworx 

lpabon:
**What type of PR is this?**
feature

This change is a continuation of @pault84 's PR #302
Moved here from #323 to bring the branch from a fork to `origin`.

Stork/Portworx driver must have the appropriate token to communicate with Px. Some functions require stork to generate its own token based on a shared secret JWT. Most calls have stork extract the token from a secret pointed by the annotations in the CRD.

Also this PR has the following updates (saved as separate commits):
* TLS support in gRPC
* Authorization support for gRPC calls like the ClusterDomain calls.
    * For this, the volume driver interface had to be edited to pass in the crd object. That way we could get the annotations which point to the secret containing the token.

Yes, This requires only that a secret is saved in the annotations.

Not sure. Maybe regular documentations will be needed in the docs for Portworx

I think so, but I am not sure which branch would be needed for Portworx release.


drivers/volume/portworx/portworx.go
```diff
@@ -164,6 +175,16 @@ type portworxGrpcConnection struct {
 	endpoint    string
 }
 
+func isTLSEnabled() bool {
+	tls := strings.ToLower(os.Getenv(pxEnableTLS))
+	if len(tls) != 0 ||
```
harsh-px: There is a bug here. This will return true for `PX_ENABLE_TLS=false`.

drivers/volume/portworx/portworx.go
```diff
@@ -246,27 +280,54 @@ func (p *portworx) initPortworxClients() error {
...
+	logrus.Infof("Using %s as the endpoint", p.endpoint)
+	clnt, err := clusterclient.NewClusterClient(p.endpoint, "v1")
 	if err != nil {
 		return err
 	}
 	p.clusterManager = clusterclient.ClusterManager(clnt)
 
-	clnt, err = volumeclient.NewDriverClient(endpointPort, "pxd", "", "stork")
+	p.stopChannel = make(chan struct{})
+	err = p.startNodeCache()
```
harsh-px: looks like startNodeCache is already being done in Init() after initPortworxClients is being called. Any reason to do this here too?

harsh-px:
I see vendor updates in the PR but no change in Gopkg.lock.

drivers/volume/portworx/portworx.go
```diff
@@ -550,12 +623,104 @@ func (p *portworx) getSnapshotName(tags *map[string]string) string {
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Set the token expiration
+		options := &auth.Options{
+			Expiration: time.Now().Add(time.Hour).Unix(),
+		}
+
+		token, err := auth.Token(claims, signature, options)
```
disrani-px: Why generate this token for every call? We should store it and regenerate when it expires.

drivers/volume/portworx/portworx.go
```diff
@@ -1332,7 +1538,12 @@ func (p *portworx) getMigrationTaskID(migration *stork_crd.Migration, volumeInfo
 }
 
 func (p *portworx) GetMigrationStatus(migration *stork_crd.Migration) ([]*stork_crd.VolumeInfo, error) {
-	status, err := p.volDriver.CloudMigrateStatus(nil)
+	var err error
```
disrani-px: Don't need this

drivers/volume/portworx/portworx.go
```diff
@@ -1382,9 +1593,14 @@ func (p *portworx) GetMigrationStatus(migration *stork_crd.Migration) ([]*stork_
 }
 
 func (p *portworx) CancelMigration(migration *stork_crd.Migration) error {
+	var err error
```
disrani-px: Don't need this

drivers/volume/portworx/portworx.go
```diff
@@ -1524,7 +1740,7 @@ func (p *portworx) GetClusterDomains() (*stork_crd.ClusterDomains, error) {
 	return clusterDomainsInfo, nil
 }
 
-func (p *portworx) ActivateClusterDomain(clusterDomainName string) error {
+func (p *portworx) ActivateClusterDomain(cdu *stork_crd.ClusterDomainUpdate, clusterDomainName string) error {
```
disrani-px: You don't need to pass in the name if you have the object, you can get it from there. Same for deactivate.

drivers/volume/portworx/portworx.go
```diff
@@ -1556,8 +1784,12 @@ func (p *portworx) DeactivateClusterDomain(clusterDomainName string) error {
 
 func (p *portworx) createGroupLocalSnapFromPVCs(groupSnap *stork_crd.GroupVolumeSnapshot, volNames []string, options map[string]string) (
 	*storkvolume.GroupSnapshotCreateResponse, error) {
-
-	resp, err := p.volDriver.SnapshotGroup("", nil, volNames)
+	var err error
```
disrani-px: Don't need this

drivers/volume/portworx/portworx.go
```diff
@@ -1633,10 +1865,15 @@ func (p *portworx) createGroupCloudSnapFromVolumes(
 	volNames []string,
 	options map[string]string) (
 	*storkvolume.GroupSnapshotCreateResponse, error) {
+	var err error
```
disrani-px: Don't need this

drivers/volume/portworx/portworx.go
```diff
@@ -1733,6 +1976,13 @@ func (p *portworx) generateStatusReponseFromTaskIDs(
 
 // revertPXCloudSnaps deletes all cloudsnaps with given IDs
 func (p *portworx) revertPXCloudSnaps(cloudSnapIDs []string, credID string) {
+	var err error
```
disrani-px: Don't need this

drivers/volume/portworx/portworx.go
```diff
@@ -246,27 +280,54 @@ func (p *portworx) initPortworxClients() error {
...
+	p.endpoint = fmt.Sprintf("%s://%v:%v", scheme, endpoint, p.restPort)
+	logrus.Infof("Using %s as the endpoint", p.endpoint)
+	clnt, err := clusterclient.NewClusterClient(p.endpoint, "v1")
 	if err != nil {
 		return err
 	}
 	p.clusterManager = clusterclient.ClusterManager(clnt)
 
-	clnt, err = volumeclient.NewDriverClient(endpointPort, "pxd", "", "stork")
+	p.stopChannel = make(chan struct{})
```
disrani-px: stopChannel is already being init'd in Init(). I think you added these by mistake during a merge

drivers/volume/portworx/portworx.go
```diff
@@ -164,6 +175,16 @@ type portworxGrpcConnection struct {
 	endpoint    string
 }
 
+func isTLSEnabled() bool {
+	tls := strings.ToLower(os.Getenv(pxEnableTLS))
+	if len(tls) != 0 ||
+		tls == "true" || tls == "yes" ||
```
disrani-px: You don't need to check for all these combinations. Just use `strconv.ParseBool()`
https://golang.org/pkg/strconv/#ParseBool

drivers/volume/portworx/portworx.go
lpabon: This is a good point, but that would mean tracking tokens with time. If a token is expired generated a new one, and so on. I think we can do that, but that complexity I would like to move to another PR for later, if you don't mind. Currently generating a token each time is very fast.

drivers/volume/portworx/portworx.go
lpabon: oops! nice catch

drivers/volume/portworx/portworx.go
lpabon: must be a rebase issue. i'll fix this.

drivers/volume/portworx/portworx.go
lpabon: Good point. i will fix that

drivers/volume/portworx/portworx.go
lpabon: oh wow, i didn't know about that API. Thanks!

drivers/volume/portworx/portworx.go
disrani-px: Generating a token is expensive if done very often, they involve crypto calls. Please file a bug to track that.
Example of how kubernetes deals with OIDC tokens to alleviate this: https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/oidc/oidc.go#L332

lpabon: OIDC is different because it is an RPC connection to get the public auth key. We only generate it. But I'm not against it. I can change it to do this, I just suggested we do this after.

lpabon: Filed #325

PR #330, opened 2019-04-10
ClusterDomains controller changes.

adityadani:
**What type of PR is this?**
>bug

- cluster.Uuid is not implemented on Portworx REST client. Use cluster.Enumerate to get the clusterID
- Ignore already exists error while creating ClusterDomainsStatus

No

No

No

### DIFF
--- a/pkg/clusterdomains/controllers/clusterdomainstatus.go
+++ b/pkg/clusterdomains/controllers/clusterdomainstatus.go
@@ -127,7 +127,7 @@ func (c *ClusterDomainsStatusController) createClusterDomainsStatusObject() {
 				Name: clusterID,
 			},
 		}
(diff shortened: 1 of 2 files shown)


pkg/clusterdomains/controllers/clusterdomainstatus.go
```diff
@@ -17,6 +17,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
harsh-px: the import already exists above

adityadani: done

PR #332, opened 2019-04-10
Allow for small time skew between authorized systems

ggriffiths:
**What type of PR is this?**
bug

Adds the IATSubtract option for auth token generation. This allows for a small amount of time skew between nodes.

No

No

2.2


disrani-px:
@ggriffiths Can you squash the vendor changes into one commit and add `Portworx:` prefix to the driver commit message

ggriffiths:
Updated the commits @disrani-px

ggriffiths:
> Doesn't need to be cherry-picked to 2.1 This is only supported in 2.2 (which is master right now). Please update the PR description.

Got it, just updated the PR description

PR #337, opened 2019-04-13
Normalize the clusterID before using it as a kubernetes resource name.

adityadani:
**What type of PR is this?**
>bug

ClusterDomainsStatus CRD uses clusterID as Name. 
This PR removes any upper case letters or `_` from it so that Kubernetes will allow creating the resource with such a name

No

No

No

### DIFF
--- a/pkg/clusterdomains/controllers/clusterdomainstatus.go
+++ b/pkg/clusterdomains/controllers/clusterdomainstatus.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/libopenstorage/stork/drivers/volume"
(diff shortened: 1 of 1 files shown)


pkg/clusterdomains/controllers/clusterdomainstatus.go
```diff
@@ -30,6 +32,10 @@ const (
 	createCdsRetryInterval               = 10 * time.Second
 )
 
+var (
+	clusterIDRegex = regexp.MustCompile("[^a-zA-Z0-9-.]+")
```
disrani-px: Make this a const?

adityadani: You can't assign regexp.MustCompile to a constant.
Assigning it to a global variable, means we only compile the regex once.

Issue #282, opened 2019-02-28
Add support to migrate CRDs


disrani-px:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Feature request

**What happened**:
Right now only a few types of resources are migrated. 

**What you expected to happen**:
Need to be able to migrate CRDs too.
Will need to qualify some well known CRDs with integration tests.

PR #344, opened 2019-04-16
Move resource collection logic from migration controller

disrani-px:
Can be used by other modules to get unstructured objects from a namepsace
and matching label selectors

**What type of PR is this?**
Cleanup

Refactored code so that same code can be shared by multiple modules

No

No

No

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -8,32 +8,28 @@ import (
 	"strings"
 	"time"
 
-	"github.com/heptio/ark/pkg/discovery"
 	"github.com/heptio/ark/pkg/util/collections"
 	"github.com/libopenstorage/stork/drivers/volume"
 	"github.com/libopenstorage/stork/pkg/apis/stork"
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/controller"
 	"github.com/libopenstorage/stork/pkg/log"
+	"github.com/libopenstorage/stork/pkg/resourcecollector"
 	"github.com/libopenstorage/stork/pkg/rule"
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
 	"github.com/portworx/sched-ops/k8s"
 	"github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubernetes/pkg/registry/core/service/portallocator"
 )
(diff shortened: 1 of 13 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -730,159 +766,29 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration) e
...
-				m.updateResourceStatus(
-					migration,
-					o,
-					stork_api.MigrationStatusFailed,
-					fmt.Sprintf("Error preparing Application resource: %v", err))
-				continue
+				return fmt.Errorf("Error preparing %v resource %v: %v", o.GetObjectKind().GroupVersionKind().Kind, metadata.GetName(), err)
 			}
 			o = updatedObject
-		case "Service":
```
sharma-tapas: It this resource type not needed any more for migrations?

pkg/migration/controllers/migration.go
```diff
@@ -730,159 +766,29 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration) e
...
-
 func (m *MigrationController) prepareResources(
 	migration *stork_api.Migration,
 	objects []runtime.Unstructured,
 ) error {
 	for _, o := range objects {
-		content := o.UnstructuredContent()
-		// Status shouldn't be migrated between clusters
-		delete(content, "status")
+		metadata, err := meta.Accessor(o)
```
sharma-tapas: Whats the difference between delete(content, "status") and meta,Accessor(o) ? How does a call to Accessor(o) removes the need to have a delete? Some code comment here, to understand what meta.Accessor(o) does and why it eliminated the need to have an explicit delete call.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,356 @@
...
+					resourceMap[metadata.GetUID()] = true
+				}
+			}
+		}
+	}
+
+	return allObjects, nil
+}
+
+func (r *ResourceCollector) objectToBeCollected(
```
sharma-tapas: A comment on top, as to what this functions purpose is.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,356 @@
...
+		name := metadata.GetName()
+		if name == "default" {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+func (r *ResourceCollector) preparePVResource(
```
sharma-tapas: comment on top, as to what this function does exactly.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return nil, err
+	}
+	delete(spec, "claimRef")
+	delete(spec, "storageClassName")
+
+	return object, nil
+}
+
+func (r *ResourceCollector) prepareServiceResource(
```
sharma-tapas: Add a comment to all the perpareXXXResource functions, as to what they are doing and why.

pkg/resourcecollector/resourcecollector.go
disrani-px: Done

pkg/resourcecollector/resourcecollector.go
disrani-px: Added comments inline for the preparation being done for PVs

pkg/resourcecollector/resourcecollector.go
disrani-px: `prepareServiceResource` does exactly one thing and there is a comment inline

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,378 @@
...
+
+func resourceToBeCollected(resource metav1.APIResource) bool {
+	// Deployment is present in "apps" and "extensions" group, so ignore
+	// "extensions"
+	if resource.Group == "extensions" && resource.Kind == "Deployment" {
+		return false
+	}
+
+	switch resource.Kind {
+	case "PersistentVolumeClaim",
```
harsh-px: we should allow users to configure this list as this CRDs, we would skip those leading to a partial migration.

Also I think we discussed this before but this needs to have storage classes. A statefulset scale up will fail post migration, which means we aren't doing a complete statefulset migration.

disrani-px: There is already an open issue for migrating CRs and CRDs (#282). There is more effort required to support that than just allowing them here since the resources created by the CRs don't need to be migrated themselves. Also, need to be able to handle cases where the CRD isn't registered on the destination.

Storageclasses are cluster specific, admins will create them during cluster install. Although a good to have I don't see that as a blocker for anything.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,378 @@
...
+		if err != nil {
+			return false, err
+		}
+		pvcName := metadata.GetName()
+		pvc, err := k8s.Instance().GetPersistentVolumeClaim(pvcName, namespace)
+		if err != nil {
+			return false, err
+		}
+		// Only collect Bound PVCs
+		if pvc.Status.Phase != v1.ClaimBound {
```
harsh-px: how about a PVC which is stuck in pending because the driver cannot satisfy the request on the source cluster?

There might still be a use case for the PVC spec to get migration so that it can be bound on the destination cluster.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -0,0 +1,378 @@
...
+// Init initializes the resource collector
+func (r *ResourceCollector) Init() error {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return fmt.Errorf("error getting cluster config: %v", err)
+	}
+
+	aeclient, err := apiextensionsclient.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("error getting apiextention client, %v", err)
```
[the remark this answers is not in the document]
disrani-px: Fixed

pkg/resourcecollector/resourcecollector.go
disrani-px: There is no way to know whether the PVC is stuck while being bound or if it was just created and will be bound soon. Users might expect different things in both cases. So I'm just being safe and ignoring these for now

Issue #342, opened 2019-04-15
Add operator to clone applications (resources + volumes) to other namespaces


disrani-px:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What happened**:
We should allow entire applications to be cloned between namespaces

**What you expected to happen**:
Namespaced and clusterscoped resources should be usable from the destination cluster

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

disrani-px:
Done with #373

PR #348, opened 2019-04-17
#342 define CRD for application cloning

sharma-tapas:
**What type of PR is this?**
>bug
feature #342 
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

Creates a new CRD for the application cloning object

```release-note

```


pkg/apis/stork/v1alpha1/applicationclone.go
```diff
@@ -0,0 +1,86 @@
...
+	meta.TypeMeta   `json:",inline"`
+	meta.ObjectMeta `json:"metadata,omitempty"`
+	Spec            ApplicationCloneSpec   `json:"spec"`
+	Status          ApplicationCloneStatus `json:"status,omitempty"`
+}
+
+//ApplicationCloneSpec defines the spec to create an application clone
+type ApplicationCloneSpec struct {
+	//NamespaceMapping to store the target:destination namespaces
+	NamespaceMapping map[string]string `json:"namespace_mapping"`
```
disrani-px: json tag should be namespaceMapping, that corresponds to the field that users use in the yaml spec.

pkg/apis/stork/v1alpha1/applicationclone.go
```diff
@@ -0,0 +1,86 @@
...
+	//Selectors for label on objects
+	Selectors    map[string]string `json:"selectors"`
+	PreExecRule  string            `json:"preExecRule"`
```
disrani-px: We need to add a `ReplacePolicy` here to govern what happens if there are volumes/resources with the same name in the destination namespace. ReplacePolicy can be `Delete`,`Retain`, `Fail`

pkg/apis/stork/v1alpha1/applicationclone.go
```diff
@@ -0,0 +1,86 @@
...
+	ApplicationCloneResourceName = "applicationclone"
+	//ApplicationCloneResourcePlural is the name in plural for the application clone resources
+	ApplicationCloneResourcePlural = "applicationclones"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+//ApplicationClone represents the cloning of application in different namespaces
+type ApplicationClone struct {
```
disrani-px: Also need to update `register.go` in this directory to add `ApplicationClone` and `ApplicationCloneList` to the schema

pkg/apis/stork/v1alpha1/applicationclone.go
```diff
@@ -0,0 +1,86 @@
...
+	NamespaceMapping map[string]string `json:"namespace_mapping"`
+	//Selectors for label on objects
+	Selectors    map[string]string `json:"selectors"`
+	PreExecRule  string            `json:"preExecRule"`
+	PostExecRule string            `json:"postExecRule"`
+}
+
+//ApplicationCloneStatus defines the status of the clone
+type ApplicationCloneStatus struct {
+	//Status of the cloning process
```
disrani-px: We need to add more info here about what volumes and resources are being cloned and the status for each, similar to the Migration spec. It'll help users know what was successful and what failed.

pkg/apis/stork/v1alpha1/applicationclone.go
```diff
@@ -0,0 +1,86 @@
+package v1alpha1
+
+import (
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	//ApplicationCloneResourceName is the name for the application clone resource
```
piyush-nimbalkar: nit: space after //, everywhere

pkg/apis/stork/v1alpha1/applicationclone.go
piyush-nimbalkar: @disrani-px , @sharma-tapas ApplicationClone sounds little misleading. It feels like it is a clone of an application. While this object is basically carrying out the migration between namespaces. Could it be called NamespaceMigration?

disrani-px: It will be cloning an application between namespaces, that's what users will want to do from their perspective. 
I want a consistent naming for the other new CRDs we are adding for application operations (ApplicationBackup and ApplicationRestore)

pkg/apis/stork/v1alpha1/applicationclone.go
piyush-nimbalkar: @sharma-tapas, Dinesh and I were discussing offline about the mapping. We feel we should have a sourceNamespace and destinationNamespace and clone only one namespace at at time.

piyush-nimbalkar: We can probably have the sourceNamespace optional in the code. If specified we will should make sure that it is the same as ApplicationClone namespace unless the AppClone is in admin namespace.

pkg/apis/stork/v1alpha1/applicationclone.go
sharma-tapas: done

pkg/apis/stork/v1alpha1/applicationclone.go
sharma-tapas: I have added volumeInfo and resourceInfo please check and let me know if we need to add something more to the info structs.

pkg/apis/stork/v1alpha1/applicationclone.go
sharma-tapas: Changed the spec to have SourceNamespace and DestinationNamespace, please check the syntax for making the SourceNamespace optional.

pkg/apis/stork/v1alpha1/applicationclone.go
sharma-tapas: Keeping it as ApplicationClone to be inline with what @disrani-px said, we can change it later on if it becomes confusing.

PR #369, opened 2019-05-14
Add debug package to dump profiles

disrani-px:
SIGUSR1 can be used to dump memory and goroutine info
SIGUSR2 can be used to toggle collection of cpuprofile

**What type of PR is this?**
Enhancement

No

```release-note
Add signal handlers to dump memory info for debugging
```

### DIFF
--- a/pkg/dbg/dbg.go
+++ b/pkg/dbg/dbg.go
@@ -1,0 +1,106 @@
+package dbg
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/signal"
+	"path"
+	"runtime"
+	"runtime/pprof"
+	"syscall"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	fnameFmt       = "2006-01-02T15:04:05.999999-0700MST"
+	stackTraceSize = 5120 * 1024 // 5MB
+)
+
+// Init Initializes the debug handlers
+func Init(name string, dir string) {
+	installSignalHandlers(name, dir)
+}
+
+func installSignalHandlers(name string, dir string) {
+	dumpChannel := make(chan os.Signal, 1)
+	cpuProfChannel := make(chan os.Signal, 1)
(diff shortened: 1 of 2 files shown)


pkg/dbg/dbg.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/signal"
+	"runtime"
+	"runtime/pprof"
+	"syscall"
+	"time"
+
```
harsh-px: nit: remove empty line number 12 so imports for formatted.

pkg/dbg/dbg.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func generateFilename(name string, dir string, suffix string) string {
+	return path.Join(dir, name+"."+time.Now().Format(fnameFmt)+suffix)
+}
+
+func startCPUProfileDump(filename string) {
+	f, err := os.Create(filename)
+	if err != nil {
+		logrus.Errorf("could not create memory profile: %v", err)
```
harsh-px: this should be "could not create CPU profile"

pkg/dbg/dbg.go
```diff
@@ -0,0 +1,96 @@
...
+	"path"
+)
+
+const (
+	fnameFmt       = "2006-01-02T15:04:05.999999-0700MST"
+	stackTraceSize = 5120 * 1024 // 5MB
+)
+
+// Initialize the debug handlers
+func Init(name string, dir string) {
```
harsh-px: function comment needs fixing

pkg/dbg/dbg.go
disrani-px: Done

pkg/dbg/dbg.go
disrani-px: Fixed

pkg/dbg/dbg.go
disrani-px: Fixed

PR #353, opened 2019-05-02
Add support for In-place snapshot restore

ram-infrac:
**What type of PR is this?**
feature

**What this PR does / why we need it**:
This PR add in-place restore support for volume snapshots. Snapshot now can be restored to origin volumes

**Does this PR change a user-facing CRD or CLI?**:
We have new CRD for In-place volume snapshot restore.
eg.
```
apiVersion: stork.libopenstorage.org/v1alpha1
kind: SnapshotRestore
metadata:
    creationTimestamp: null
    name: mysql-group-snap-test123
    namespace: default
spec:
  sourcetype: "local"
  sourcename: "mysql-group-snap"
  sourcenamespace: "default"
  groupsnapshot: true
```
sourcetype: snapshot type either local or cloud
sourcename: name of snapshot to be restore
groupsnapshot: true if snapshot is `volumegroupsnapshot`
**Is a release note needed?**:

**Does this change need to be cherry-picked to a release branch?**:

PR #377, opened 2019-05-20
In-Place Volume Snapshot restore  (2)

ram-infrac:
**What type of PR is this?**

>feature

Continuing PR for in place restore, should be reviewed after #353  
base PR https://github.com/libopenstorage/stork/pull/353


pkg/extender/extender.go
```diff
@@ -35,6 +36,8 @@ const (
 	defaultScore = 5
 
 	schedulingFailureEventReason = "FailedScheduling"
+	annotationPrefix             = "stork.libopenstorage.org/"
+	volumerestore                = annotationPrefix
```
disrani-px: This should be ` annotationPrefix + "restore-in-progress"`?

disrani-px: Actually, this is defined in `pkg/snapshot/controllers/snapshotrestore.go` and should be used from there

pkg/extender/extender.go
```diff
@@ -123,6 +126,20 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
 	}
 
 	pod := args.Pod
+	for _, vol := range pod.Spec.Volumes {
+		// if any of pvc has restore annotation skip scheduling pod
+		pvc, err := k8s.Instance().GetPersistentVolumeClaim(vol.PersistentVolumeClaim.ClaimName, pod.Namespace)
+		if err != nil {
+			msg := "Unable to find PVC claim"
```
disrani-px: Add the PVC name to the message. And no need to add `claim`

pkg/extender/extender.go
```diff
@@ -123,6 +126,20 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
+			storklog.PodLog(pod).Warnf(msg)
```
disrani-px: Need to respond to the request with an error with `http.Error(w, ...)` and return

pkg/extender/extender.go
```diff
@@ -123,6 +126,20 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
+		} else if pvc.Annotations != nil && pvc.Annotations[volumerestore] == "true" {
+			msg := "Volume restore is in progress for pvc: " + pvc.Name
+			storklog.PodLog(pod).Warnf(msg)
+			e.Recorder.Event(pod, v1.EventTypeWarning, schedulingFailureEventReason, msg)
+			return
```
disrani-px: Need to respond with error before returning

pkg/log/log.go
```diff
@@ -204,3 +204,14 @@ func PVCLog(pvc *v1.PersistentVolumeClaim) *logrus.Entry {
 
 	return logrus.WithFields(logrus.Fields{})
 }
+
+// VolumeSnapshotRestoreLog formats a log message with volumesnapshotrestore information
+func VolumeSnapshotRestoreLog(snapRestore *storkv1.VolumeSnapshotRestore) *logrus.Entry {
```
disrani-px: Add UT for this in `log_test.go`

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +53,165 @@ func (c *SnapshotRestoreController) Init() error {
...
+	)
+
 	switch o := event.Object.(type) {
 	case *stork_api.VolumeSnapshotRestore:
-		snapRestore := o
-		if snapRestore.Spec.SourceName == "" || snapRestore.Spec.SourceType == "" {
-			return fmt.Errorf("empty Snapshot name/type or namespace")
+		snapRestore = o
+		if snapRestore.Spec.SourceName == "" || snapRestore.Spec.SourceNamespace == "" {
+			return fmt.Errorf("empty Snapshot name or namespace")
```
disrani-px: Also create an event. Namespace cannot be empty, so skip that check

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +53,165 @@ func (c *SnapshotRestoreController) Init() error {
...
+	if err != nil {
+		return err
+	}
 
-		err := c.Driver.VolumeSnapshotRestore(snapRestore)
+	return nil
+}
+
+func (c *SnapshotRestoreController) handleInitial(snapRestore *stork_api.VolumeSnapshotRestore) error {
+	log.VolumeSnapshotRestoreLog(snapRestore).Infof("handled initial")
```
disrani-px: Don't need this log

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +53,165 @@ func (c *SnapshotRestoreController) Init() error {
...
 
-		err = sdk.Update(snapRestore)
+	// Do driver volume snapshot restore here
+	err = c.Driver.VolumeSnapshotRestore(snapRestore, restoreVolumes)
+	if err != nil {
+		snapRestore.Status.Status = stork_api.VolumeSnapshotRestoreStatusError
+		return fmt.Errorf("failed to restore pvc details %v", err)
+	}
+
+	err = unmarkPVCForRestore(pvcList)
```
disrani-px: This needs to be done even on error on line 153

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +53,165 @@ func (c *SnapshotRestoreController) Init() error {
...
+	// snapshot is list of snapshots
+	snapshotList := []*snap_v1.VolumeSnapshot{}
+
+	snapName := snapRestore.Spec.SourceName
+	snapNamespace := snapRestore.Spec.SourceNamespace
+	if snapRestore.Spec.GroupSnapshot {
+		logrus.Infof("GroupVolumeSnapshot In-place restore request for %v", snapName)
+		snapshotList, err = k8s.Instance().GetSnapshotsForGroupSnapshot(snapName, snapNamespace)
+		if err != nil {
+			logrus.Errorf("unable to get group snapshot details %v", err)
```
disrani-px: Update all logs to use `log.VolumeSnapshotRestoreLog(snapRestore)` to help with debugging

pkg/apis/stork/v1alpha1/snapshotrestore.go
```diff
@@ -29,18 +27,34 @@ type VolumeSnapshotRestoreSpec struct {
...
+	// VolumeSnapshotRestoreStatusError for when restore failed
+	VolumeSnapshotRestoreStatusError SnapshotRestoreStatusType = "Error"
+)
+
+// VolumeSnapshotRestoreStageType is stage of volume in-place restore
+type VolumeSnapshotRestoreStageType string
+
+const (
+	// VolumeSnapshotRestoreStageInitial is the initial stage when snapshot restore is initiated
+	VolumeSnapshotRestoreStageInitial VolumeSnapshotRestoreStageType = ""
```
disrani-px: The stage and status have same values. Maybe we don't need stage for this. Just the following status: Initial, Pending, InProgress, Failed, Successful

pkg/extender/extender.go
disrani-px: Still missing `http.Error(w, ...)` here?

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +54,172 @@ func (c *SnapshotRestoreController) Init() error {
...
+		if event.Deleted {
+			return c.handleDelete(snapRestore)
+		}
+
+		switch snapRestore.Status.Stage {
+		case stork_api.VolumeSnapshotRestoreStageInitial:
+			err = c.handleInitial(snapRestore)
+		case stork_api.VolumeSnapshotRestoreStageRestore:
+			err = c.handleFinal(snapRestore)
+			c.Recorder.Event(snapRestore,
```
disrani-px: This event needs to be created only if `err == nil`

pkg/apis/stork/v1alpha1/snapshotrestore.go
```diff
@@ -29,18 +27,34 @@ type VolumeSnapshotRestoreSpec struct {
...
+	VolumeSnapshotRestoreStageRestore VolumeSnapshotRestoreStageType = "Restore"
+	// VolumeSnapshotRestoreStageReady for when restore is completed
+	VolumeSnapshotRestoreStageReady VolumeSnapshotRestoreStageType = "Ready"
+	// VolumeSnapshotRestoreStageError for when restore failed
+	VolumeSnapshotRestoreStageError VolumeSnapshotRestoreStageType = "Error"
 )
 
 // RestoreStatus of volume
 type RestoreStatus struct {
+	// Stage of volume snapshot restore
```
disrani-px: We should add a `Reason` string in `RestoreVolumeInfo`. On success it should have `Volume restored successfully` and on failure it should have the error message.

pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -45,34 +54,172 @@ func (c *SnapshotRestoreController) Init() error {
...
+	// TODO: Need to mark deployment/daemonset as stork scheduler
+	err = markPVCForRestore(pvcList)
+	if err != nil {
+		return err
+	}
+
+	// Do driver volume snapshot restore here
+	err = c.Driver.VolumeSnapshotRestore(snapRestore, restoreVolumes)
+	if err != nil {
+		snapRestore.Status.Status = stork_api.VolumeSnapshotRestoreStatusError
```
disrani-px: We shouldn't mark if failed here. Return an error here so that it retries.

pkg/extender/extender.go
ram-infrac: if pod does not have valid volumes claim should we stop scheduling or just warn in log and go ahead

PR #379, opened 2019-05-21
Application backup schedule

disrani-px:
**What type of PR is this?**
Feature

Adds new CRD for ApplicationBackupSchedule

```release-note
Yes
TODO: Add CRD spec for schedule
```

No


pkg/applicationmanager/controllers/applicationbackupschedule.go
```diff
@@ -0,0 +1,342 @@
...
+		for _, backup := range policyApplicationBackup {
+			// Get the updated status if we see it as not completed
+			if !s.isApplicationBackupComplete(backup.Status) {
+				pendingApplicationBackupStatus, err := getApplicationBackupStatus(backup.Name, backupSchedule.Namespace)
+				if err != nil {
+					s.Recorder.Event(backupSchedule,
+						v1.EventTypeWarning,
+						string(stork_api.ApplicationBackupStatusFailed),
+						fmt.Sprintf("Error getting status of backup %v: %v", backup.Name, err))
+				}
```
piyush-nimbalkar: Should we return err if we could not get the backup status? 
Currently we are marking the backup as failed if the status cannot be retrieved.

disrani-px: Added a continue for errors other than `NotFound` because we want to mark those as failed and still check status for other backups

PR #380, opened 2019-05-21
[Portworx] Add idempotency for local snapshots

harsh-px:

**What type of PR is this?**
>bug

If stork crashes after invoking the driver to create a snapshot but before returning success back to the snapshot provisioner, then the snapshot is created in the backend. But stork will keep retrying to bind the snapshot and will fail since the snapshot with the same ID already exists.

no

```release-note
Add idempotency to local snapshot calls to handle crashes while a snapshot is being taken.
```

no

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -844,7 +844,23 @@ func (p *portworx) SnapshotCreate(
 		}
 		snapshotID, err = volDriver.Snapshot(volumeID, true, locator, true)
 		if err != nil {
-			return nil, getErrorSnapshotConditions(err), err
+			// Check already exists error and return existing snapshot if found
+			if isAlreadyExistsError(err) {
+				snapInfo, err := p.inspectVolume(volDriver, snapName)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


disrani-px:
Add a `[Portworx]` or `Portworx` prefix to the commit message since this is a driver only change.

harsh-px:
> Add a [Portworx] or Portworx prefix to the commit message since this is a driver only change.


Done

drivers/volume/portworx/portworx.go
```diff
@@ -844,7 +844,23 @@ func (p *portworx) SnapshotCreate(
 		}
 		snapshotID, err = volDriver.Snapshot(volumeID, true, locator, true)
 		if err != nil {
-			return nil, getErrorSnapshotConditions(err), err
+			// Check already exists error and return existing snapshot if found
+			if isAlreadyExistsError(err) {
```
disrani-px: File a bug to return a proper typed error from `volDriver.Snapshot()`.

harsh-px: Done PWX-8959

PR #385, opened 2019-05-24
Create some default schedule policies

disrani-px:
Also update strokctl to use a default policy for migration schedule

**What type of PR is this?**
Enhancement

No

```release-note
* Added some default schedule policies:
  * default-migration-policy: Triggers after every minute
  * default-interval-policy: Triggers after every hour. Retains last 10
  * default-daily-policy: Triggers every day at 12:00am. Retains last 7
  * default-weekly-policy: Triggers every Sunday@12:00am. Retains last 4
  * default-monthly-policy: Triggers every 15th day of the month at 12:00am. Retains last 12.
```

2.2

### DIFF
--- a/pkg/schedule/schedule.go
+++ b/pkg/schedule/schedule.go
@@ -290,6 +290,77 @@ func Init() error {
 		}
 	}
 
+	return createDefaultPolicy()
+}
+
+func createDefaultPolicy() error {
+	_, err := k8s.Instance().CreateSchedulePolicy(&stork_api.SchedulePolicy{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "default-migration-policy",
+		},
+		Policy: stork_api.SchedulePolicyItem{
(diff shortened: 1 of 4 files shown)


pkg/schedule/schedule.go
```diff
@@ -290,6 +290,77 @@ func Init() error {
 		}
 	}
 
+	return createDefaultPolicy()
+}
+
+func createDefaultPolicy() error {
+	_, err := k8s.Instance().CreateSchedulePolicy(&stork_api.SchedulePolicy{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "default-migration-policy",
```
piyush-nimbalkar: should we call it migration policy or keep it generic?

disrani-px: This is used specifically as the default for migrations when creating a schedule using `storkctl`

PR #394, opened 2019-05-29
Integration test: Use remote cluster to activate/deactivate domains

adityadani:
**What type of PR is this?**
>integration-test

Always use the remote cluster to activate/deactivate a cluster domain as in our tests the remote cluster is never stopped/deactivated

No

No

No


test/integration_test/clusterdomain_test.go
```diff
@@ -45,12 +46,15 @@ func triggerClusterDomainUpdate(
 	domain string,
 	active bool,
 ) {
-	err := setRemoteConfig("")
+	// run the command on the remote cluster as currently
+	// we do not stop/deactivate the remote cluster
+	err := setRemoteConfig(remoteFilePath)
```
disrani-px: Add a `defer` to reset the remote config

PR #397, opened 2019-05-30
Fail migrations if domain is inactive

disrani-px:
**What type of PR is this?**
Enhancement

No

```release-note
* Migrations will be marked as Failed if the local cluster domain is inactive
* Updated migration behavior for PVs
  * Setting reclaim policy to Retain if volumes aren't being migrated
  * Updating PV if it already exists instead of deleting and creating
```

No, this is for 2.2 branch. Will send separate PR for master since code has changed there.

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -183,6 +183,33 @@ func (m *MigrationController) Handle(ctx context.Context, event sdk.Event) error
 
 		var terminationChannels []chan bool
 		var err error
+		var clusterDomains *stork_api.ClusterDomains
+		for i := 0; i < domainsMaxRetries; i++ {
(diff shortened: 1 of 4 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -184,6 +184,27 @@ func (m *MigrationController) Handle(ctx context.Context, event sdk.Event) error
 		var terminationChannels []chan bool
 		var err error
 
+		// Fail the migration if the current domain is inactive
+		clusterDomains, err := m.Driver.GetClusterDomains()
```
adityadani: Can we retry this 4-5 times with a delay of 2-3 seconds? So that even if we hit the window where PX is restarting you will eventually get the cluster domains?

PR #400, opened 2019-06-07
Storkctl subcommands for applications crds

disrani-px:
**What type of PR is this?**
Enhancement

Adds new subcommands to `storkctl`

Include with the feature release notes

No

### DIFF
--- a/pkg/storkctl/applicationclone_test.go
+++ b/pkg/storkctl/applicationclone_test.go
@@ -1,0 +1,227 @@
+// +build unittest
+
+package storkctl
+
+import (
+	"testing"
+	"time"
+
+	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
+	"github.com/portworx/sched-ops/k8s"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestGetClonesNoClone(t *testing.T) {
+	cmdArgs := []string{"get", "applicationclones"}
+
+	var cloneList storkv1.ApplicationCloneList
+	expected := "No resources found.\n"
+	testCommon(t, cmdArgs, &cloneList, expected, false)
+}
+
+func createApplicationCloneAndVerify(
+	t *testing.T,
+	name string,
+	namespace string,
+	sourceNamespace string,
+	destinationNamespace string,
+	preExecRule string,
+	postExecRule string,
+) {
+	cmdArgs := []string{"create", "clones", "-n", namespace, "--sourceNamespace", sourceNamespace, "--destinationNamespace", destinationNamespace, name}
+	if preExecRule != "" {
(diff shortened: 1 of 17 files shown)


pkg/storkctl/applicationclone_test.go
```diff
@@ -0,0 +1,227 @@
...
+	name := "dummy-name"
+	cmdArgs := []string{"create", "clones", "-n", srcNamespace, "--sourceNamespace", srcNamespace,
+		"--destinationNamespace", dstNamespace, name, "-w"}
+
+	expected := "ApplicationClone dummy-name started successfully\n" +
+		"STAGE\t\tSTATUS              \n" +
+		"\t\t                    \n" +
+		"Volumes\t\tSuccessful          \n" +
+		"ApplicationClone dummy-name completed successfully\n"
+	go setApplicationCloneStatus(name, srcNamespace, false, t)
```
ram-infrac: I'm planning to customise retryTimeout and retryInterval for unit test purpose 
eg. `storkctl create migrations -w 2,10 (2 seconds retry, 10 Min wait)

pkg/storkctl/backuplocation.go
```diff
@@ -0,0 +1,221 @@
...
+
+			s3BackupLocations := &storkv1.BackupLocationList{}
+			azureBackupLocations := &storkv1.BackupLocationList{}
+			googleBackupLocations := &storkv1.BackupLocationList{}
+			unknownBackupLocations := &storkv1.BackupLocationList{}
+			for _, bl := range backupLocations.Items {
+				switch bl.Location.Type {
+				case storkv1.BackupLocationS3:
+					if !showSecrets && bl.Location.S3Config != nil {
+						bl.Location.S3Config.SecretAccessKey = hiddenString
```
ram-infrac: do we also need to <hidde>  access key -id as well
Same for other cloud provider Storage-Account-Name and Project-ID

pkg/storkctl/applicationrestore.go
```diff
@@ -0,0 +1,249 @@
...
+		Use:     applicationRestoreSubcommand,
+		Aliases: applicationRestoreAliases,
+		Short:   "Start an applicationRestore",
+		Run: func(c *cobra.Command, args []string) {
+			if len(args) != 1 {
+				util.CheckErr(fmt.Errorf("exactly one name needs to be provided for applicationrestore name"))
+				return
+			}
+			if backupLocation == "" {
+				util.CheckErr(fmt.Errorf("need to provide BackupLocation to use for backup"))
```
ram-infrac: for restore*

pkg/storkctl/applicationclone_test.go
disrani-px: Unit tests can change private variables so we can just reduce the variables for the poll times. I already reduced it for the migration UT

pkg/storkctl/backuplocation.go
disrani-px: Only trying to hide the secret information here.

PR #401, opened 2019-06-11
Add inplace restore support for cloudsnapshots

ram-infrac:
**What type of PR is this?**
>enhancement

Add inlace restore support for Portworx driver cloud snapshot  restore
 
No

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -1105,7 +1105,7 @@ func (p *portworx) checkAndUpdateHaLevel(volID string, params map[string]string)
 		logrus.Warnf("Volume is not ready yet %v", restoreVol.GetLocator().GetName())
 		return false, nil
 	}
-	logrus.Infof("Restore Volume Status %v \t state %v", restoreVol.Status, restoreVol.State)
+	logrus.Infof("Restore Volume: %v \t Status %v \t state %v", restoreVol.Id, restoreVol.Status, restoreVol.State)
 	// find replica nodes
 	var nodes []string
 	rsSets := parentVols[0].GetReplicaSets()
@@ -1113,7 +1113,7 @@ func (p *portworx) checkAndUpdateHaLevel(volID string, params map[string]string)
 		nodes = append(nodes, rs.GetNodes()...)
 	}
 	if restoreVol.GetSpec().GetHaLevel() != parentVols[0].GetSpec().GetHaLevel() {
-		logrus.Infof("Updating HA Level of volume %v", parentVols[1].GetLocator().GetName())
+		logrus.Infof("Updating HA Level of volume %v", restoreVol.GetLocator().GetName())
 		spec := &api.VolumeSpec{
 			HaLevel:    restoreVol.GetSpec().GetHaLevel() + 1,
 			ReplicaSet: &api.ReplicaSet{Nodes: nodes},
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -957,28 +957,43 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 	volumeInfos := make([]*stork_crd.RestoreVolumeInfo, 0)
 	volRestores, snapType, credID, err := processRestoreVolumes(snapRestore.Status.RestoreVolumes)
 	if err != nil {
-		logrus.Errorf("Invalid snapshot data %v", err)
+		log.VolumeSnapshotRestoreLog(snapRestore).Errorf("Invalid snapshot data %v", err)
 		return err
 	}
+
 	switch snapType {
-	case "", string(crdv1.PortworxSnapshotTypeLocal):
-		// for local snapshot return vollist as nil
+	case "", crdv1.PortworxSnapshotTypeLocal:
+		snapRestore.Status.Volumes = volumeInfos
 		return nil
-	case string(crdv1.PortworxSnapshotTypeCloud):
+	case crdv1.PortworxSnapshotTypeCloud:
 		// Get volume client from user context
 		volDriver, err := p.getUserVolDriver(snapRestore.Annotations)
 		if err != nil {
 			return err
 		}
 		for volID, snapID := range volRestores {
-			taskID := restoreTaskPrefix + volID
+			// Get Volume Info
+			vols, err := volDriver.Inspect([]string{volID, restoreNamePrefix + volID})
+			if err != nil {
+				return &ErrFailedToInspectVolume{
+					ID:    volID,
+					Cause: fmt.Sprintf("Volume inspect returned err: %v", err),
+				}
+			} else if len(vols) == 0 {
+				return &ErrFailedToInspectVolume{
+					ID:    volID,
+					Cause: fmt.Sprintf("Volume not found err: %v", err),
+				}
+			}
+			replNodes := vols[0].GetReplicaSets()[0]
+			taskID := restoreTaskPrefix + snapID + "-" + volID
 			restoreName := restoreNamePrefix + volID
-			_, err := volDriver.CloudBackupRestore(&api.CloudBackupRestoreRequest{
+			_, err = volDriver.CloudBackupRestore(&api.CloudBackupRestoreRequest{
 				Name:              taskID,
 				ID:                snapID,
 				RestoreVolumeName: restoreName,
 				CredentialUUID:    credID,
-				// Need node id to restore
+				NodeID:            replNodes.GetNodes()[0],
 			})
 			if err != nil {
 				if _, ok := err.(*ost_errors.ErrExists); !ok {
@@ -988,6 +1003,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 			volInfo := &stork_crd.RestoreVolumeInfo{}
 			volInfo.RestoreStatus = stork_crd.VolumeSnapshotRestoreStatusInProgress
 			volInfo.Volume = volID
+			volInfo.Snapshot = snapID
 			volInfo.Reason = "Volume restore in progress"
 			volumeInfos = append(volumeInfos, volInfo)
 		}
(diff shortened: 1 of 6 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1004,25 +1082,21 @@ func (p *portworx) localSnapshotRestore(
...
+			snapID = restoreNamePrefix + volID
+		}
 		err = volDriver.Restore(volID, snapID)
 		if err != nil {
 			logrus.Errorf("Unable to restore volume %v with ID %v, err %v",
 				volID, snapID, err)
 			return err
 		}
 		logrus.Infof("Completed restore for volume %v with Snapshotshot %v", volID, snapID)
+		// TODO: Delete restore objects
```
ram-infrac: @disrani-px  should we delete restore object here ? or have separate api to cleanup restore objects.
So handleDelete will also delete pending restore objects
Restore objects here are - 
volumes with prefix `in-place-restore`

disrani-px: Once the restore is complete to the original volume the temporary volume should be deleted. The restore should not be marked succesful until that is deleted otherwise it could leave orphaned volumes

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
 }
 
-// VolumeSnapshotRestore does in-place restore of snapshot to it's parent volume
-func (p *portworx) VolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnapshotRestore, restoreVolumes map[string]string) error {
+// StartVolumeSnapshotRestore will prepare volume for restore
+func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnapshotRestore) error {
+	volumeInfos := make([]*stork_crd.RestoreVolumeInfo, 0)
+	volRestores, snapType, credID, err := processRestoreVolumes(snapRestore.Status.RestoreVolumes)
+	if err != nil {
+		logrus.Errorf("Invalid snapshot data %v", err)
```
disrani-px: Use `VolumeSnapshotRestoreLog()`

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
 				snapDataName, err)
 		}
 		// Let's verify if source snapshotdata is complete
 		err = k8s.Instance().ValidateSnapshotData(snapshotData.Metadata.Name, false, validateSnapshotTimeout, validateSnapshotRetryInterval)
 		if err != nil {
-			return volRestore, "", fmt.Errorf("snapshot: %s is not complete. %v", snapshotData.Metadata.Name, err)
+			return volRestore, "", "", fmt.Errorf("snapshot: %s is not complete. %v", snapshotData.Metadata.Name, err)
 		}
 		snapID := snapshotData.Spec.PortworxSnapshot.SnapshotID
 		snapType = string(snapshotData.Spec.PortworxSnapshot.SnapshotType)
```
disrani-px: Remove this type casting and return as `crdv1.PortworxSnapshotType`

pkg/apis/stork/v1alpha1/snapshotrestore.go
```diff
@@ -47,6 +50,12 @@ type RestoreStatus struct {
 	Status SnapshotRestoreStatusType `json:"status"`
 	// Volumes list of restore inforamtion
 	Volumes []*RestoreVolumeInfo `json:"volumes"`
+	// RestoreVolume map of snapID and volID to restore
+	RestoreVolumes map[string]string `json:"restoreVolumes"`
+	// List of PVC associated with snapshot restore
+	PVCs []*v1.PersistentVolumeClaim `json:"pvcs"`
+	// RestoreType restore type of volumes
+	RestoreType string `json:"restoreType"`
```
disrani-px: This shouldn't be in the CRD. This is PX specific. Store this as an annotation in the spec which PX can refer to

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+	volRestores, snapType, credID, err := processRestoreVolumes(snapRestore.Status.RestoreVolumes)
+	if err != nil {
+		logrus.Errorf("Invalid snapshot data %v", err)
+		return err
+	}
+
+	switch snapType {
+	case "", string(crdv1.PortworxSnapshotTypeLocal):
+		snapRestore.Status.Volumes = volumeInfos
+		snapRestore.Status.RestoreType = string(crdv1.PortworxSnapshotTypeLocal)
```
disrani-px: The type shouldn't be stored in the spec status

pkg/apis/stork/v1alpha1/snapshotrestore.go
disrani-px: Actually, it doesn't need to be stored anywhere. We should always get it from the snapshot information

pkg/apis/stork/v1alpha1/snapshotrestore.go
```diff
@@ -37,6 +38,8 @@ const (
 	VolumeSnapshotRestoreStatusSuccessful SnapshotRestoreStatusType = "Successful"
 	// VolumeSnapshotRestoreStatusInProgress for when restore is in progress
 	VolumeSnapshotRestoreStatusInProgress SnapshotRestoreStatusType = "InProgress"
+	// VolumeSnapshotRestoreStatusResyncInProgress for when restore is in sync state
+	VolumeSnapshotRestoreStatusResyncInProgress SnapshotRestoreStatusType = "Resync"
```
disrani-px: This is PX specific, stork CRDs shouldn't know about it. Store it as an annotation that PX can understand

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+		return nil
+	case string(crdv1.PortworxSnapshotTypeCloud):
+		// Get volume client from user context
+		volDriver, err := p.getUserVolDriver(snapRestore.Annotations)
+		if err != nil {
+			return err
+		}
+		for volID, snapID := range volRestores {
+			// Get Volume Info
+			vols, err := volDriver.Inspect([]string{volID, restoreNamePrefix + volID})
```
disrani-px: Why is `restoreNamePrefix + volID` being used in the Inspect()?

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+					Cause: fmt.Sprintf("Volume inspect returned err: %v", err),
+				}
+			} else if len(vols) == 0 {
+				return &ErrFailedToInspectVolume{
+					ID:    volID,
+					Cause: fmt.Sprintf("Volume not found err: %v", err),
+				}
+			}
+			replNodes := vols[0].GetReplicaSets()[0]
+			taskID := restoreTaskPrefix + volID
```
disrani-px: Use `snapRestore.UID` as the prefix so that it is unique. Otherwise, the restore will work only once, it'll return ErrExists for the next restore

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+			Cause: fmt.Sprintf("Volume inspect returned err: %v", err),
+		}
+	} else if len(vols) == 0 {
+		return false, &ErrFailedToInspectVolume{
+			ID:    volID,
+			Cause: fmt.Sprintf("Volume not found err: %v", err),
+		}
+	}
+	logrus.Debugf("volume info %v \t %v", vols[0].GetLocator().GetName(), vols[1].GetLocator().GetName())
+	restoreVol := vols[1]
```
disrani-px: The order of the returned volumes should not be assumed. Make sure the first is the original and the second is the restored volume

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,201 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+	if vols[1].Status != api.VolumeStatus_VOLUME_STATUS_UP {
+		logrus.Warnf("Volume is not ready yet %v", restoreVol.GetLocator().GetName())
+		return false, nil
+	}
+	logrus.Infof("Restore Volume Status %v \t state %v", restoreVol.Status, restoreVol.State)
+	if restoreVol.GetSpec().GetHaLevel() != vols[0].GetSpec().GetHaLevel() {
+		logrus.Infof("Updating HA Level of volume %v", vols[1].GetLocator().GetName())
+		spec := &api.VolumeSpec{
+			HaLevel:    restoreVol.GetSpec().GetHaLevel() + 1,
+			ReplicaSet: &api.ReplicaSet{},
```
disrani-px: Need to provide the same nodes as the original volumes in the replica set

drivers/volume/portworx/portworx.go
ram-infrac: we are downloading restore object to volume ```restoreNamePrefix + volID```

drivers/volume/portworx/portworx.go
```diff
@@ -949,53 +952,227 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+	}
+	logrus.Infof("Restore Volume Status %v \t state %v", restoreVol.Status, restoreVol.State)
+	// find replica nodes
+	var nodes []string
+	rsSets := parentVols[0].GetReplicaSets()
+	for _, rs := range rsSets {
+		nodes = append(nodes, rs.GetNodes()...)
+	}
+	if restoreVol.GetSpec().GetHaLevel() != parentVols[0].GetSpec().GetHaLevel() {
+		logrus.Infof("Updating HA Level of volume %v", parentVols[1].GetLocator().GetName())
```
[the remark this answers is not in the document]
ram-infrac: volDriver.Inspect() returns array of volumes for given volume ID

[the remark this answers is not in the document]
ram-infrac: good catch , thanks

PR #402, opened 2019-06-12
Add controller to sync backups

disrani-px:
**What type of PR is this?**
Enhancement

No
```release-note
* Backups will be synced from a BackupLocation if not already present on the cluster
```

No


pkg/applicationmanager/controllers/backupsync.go
```diff
@@ -0,0 +1,156 @@
...
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			if object.IsDir {
+				data, err := bucket.ReadAll(context.TODO(), filepath.Join(object.Key, metadataObjectName))
+				if err != nil {
+					continue
```
ram-infrac: we may want to log errors before `continue` here

disrani-px: Done

PR #414, opened 2019-07-10
Allow users to specify annotation in pod if only local node should be…

disrani-px:
… preferred

**What type of PR is this?**
Improvement

No

```release-note
Users can now set the annotation `stork.libopenstorage.org/preferLocalNodeOnly` to `true` in their pods if they want pods to be scheduled only on nodes where data is located.
```

2.2

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -34,9 +35,10 @@ const (
 	// same region as a node which has data for the volume
 	regionPriorityScore = 10
 	// defaultScore Score assigned to a node which doesn't have data for any volume
-	defaultScore = 5
-
+	defaultScore                 = 5
 	schedulingFailureEventReason = "FailedScheduling"
+	// annotation to check if only local nodes should be used to schedule a pod
+	preferLocalNodeOnlyAnnotation = "stork.libopenstorage.org/preferLocalNodeOnly"
 )
 
 // Extender Scheduler extender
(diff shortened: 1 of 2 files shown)


pkg/extender/extender.go
```diff
@@ -176,30 +178,58 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
+					storklog.PodLog(pod).Errorf("No online storage nodes have replica for volume, returning error")
 					msg := "No online node found with volume replica"
 					e.Recorder.Event(pod, v1.EventTypeWarning, schedulingFailureEventReason, msg)
 					http.Error(w, msg, http.StatusBadRequest)
 					return
 				}
 			}
 
+			preferLocalOnly := false
+			if value, ok := pod.Annotations[preferLocalNodeOnlyAnnotaion]; ok {
```
harsh-px: safe to nil check `pod.Annotations`

pkg/extender/extender.go
```diff
@@ -34,9 +35,10 @@ const (
 	// same region as a node which has data for the volume
 	regionPriorityScore = 10
 	// defaultScore Score assigned to a node which doesn't have data for any volume
-	defaultScore = 5
-
+	defaultScore                 = 5
 	schedulingFailureEventReason = "FailedScheduling"
+	// annotation to check if only local nodes should be used to schedule a pod
+	preferLocalNodeOnlyAnnotaion = "stork.libopenstorage.org/preferLocalNodeOnly"
```
harsh-px: type in `preferLocalNodeOnlyAnnotaion`

pkg/extender/extender.go
disrani-px: Done

pkg/extender/extender.go
disrani-px: Done

PR #422, opened 2019-07-23
Validate name and namespace when generating clusterpair

disrani-px:
**What type of PR is this?**
Bug

Yes, adds validation for name and namespace for `storkctl generate clusterpair` command

```release-note
* Added validation for `name` and `namespace` in the clusterpair generated with `storkctl generate clusterpair`
```

2.2


disrani-px:
Added unit tests

PR #424, opened 2019-07-25
Bug fixes for In-place volumesnapshot restore controller

ram-infrac:
**What type of PR is this?**
>bug

 - Fixes HA update failure for restore objects
 - Added restoreAPICallBackoff  in case time require to detach volume is more
 - Use updated PVC for marking/unmarking annotation
 - Handle restore failed stage
No

2.0.3

### DIFF
--- a/pkg/snapshot/controllers/snapshotrestore.go
+++ b/pkg/snapshot/controllers/snapshotrestore.go
@@ -216,7 +216,7 @@ func markPVCForRestore(pvcList []*v1.PersistentVolumeClaim) ([]*v1.PersistentVol
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -224,11 +230,12 @@ func markPVCForRestore(pvcList []*v1.PersistentVolumeClaim) ([]*v1.PersistentVol
 				log.PodLog(&pod).Errorf("Error deleting pod %v: %v", pod.Name, err)
 				return updatedPvc, err
 			}
-
+			log.PodLog(&pod).Infof("Deleted before wait pod %v", pod.Name)
```
disrani-px: Can we change this and the next one to `Debugf()`

PR #426, opened 2019-07-26
Update permissions for stork-role

disrani-px:
Needed for application restore

**What type of PR is this?**
Permissions update

No

No

No

### DIFF
--- a/specs/stork-daemonset.yaml
+++ b/specs/stork-daemonset.yaml
@@ -53,57 +53,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
    name: stork-role
 rules:
-  - apiGroups: [""]
-    resources: ["pods", "pods/exec"]
-    verbs: ["get", "list", "delete", "create"]
-  - apiGroups: [""]
-    resources: ["persistentvolumes"]
-    verbs: ["get", "list", "watch", "create", "delete"]
-  - apiGroups: [""]
-    resources: ["persistentvolumeclaims"]
-    verbs: ["get", "list", "watch", "update"]
-  - apiGroups: ["storage.k8s.io"]
-    resources: ["storageclasses"]
-    verbs: ["get", "list", "watch"]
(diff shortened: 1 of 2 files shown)


specs/stork-deployment.yaml
```diff
@@ -57,57 +57,9 @@ apiVersion: rbac.authorization.k8s.io/v1
...
-    verbs: ["list", "get", "watch", "patch", "update", "initialize"]
-  - apiGroups: ["*"]
-    resources: ["statefulsets", "statefulsets/extensions"]
-    verbs: ["list", "get", "watch", "patch", "update", "initialize"]
-  - apiGroups: ["extensions"]
-    resources: [" thirdpartyresources"]
-    verbs: [ "get", "list", "watch","create","delete"]
   - apiGroups: ["*"]
     resources: ["*"]
-    verbs: ["get", "list"]
```
Rohit-PX: do we also need `create` and `update` here?

specs/stork-daemonset.yaml
```diff
@@ -53,57 +53,9 @@ apiVersion: rbac.authorization.k8s.io/v1
...
-    verbs: ["list", "get", "watch", "patch", "update", "initialize"]
-  - apiGroups: ["*"]
-    resources: ["statefulsets", "statefulsets/extensions"]
-    verbs: ["list", "get", "watch", "patch", "update", "initialize"]
-  - apiGroups: ["extensions"]
-    resources: [" thirdpartyresources"]
-    verbs: [ "get", "list", "watch","create","delete"]
   - apiGroups: ["*"]
     resources: ["*"]
-    verbs: ["get", "list"]
```
Rohit-PX: do we also need `create` and `update` here?

specs/stork-deployment.yaml
disrani-px: I've added `*` instead of the specific verbs, so it'll have all permissions.

PR #427, opened 2019-07-30
Cluster domain test

Rohit-PX:
**What type of PR is this?**
integration-test

no

no

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strconv"
 	"testing"
 	"time"
 
@@ -52,6 +53,8 @@ const (
 	defaultWaitTimeout       time.Duration = 5 * time.Minute
 	groupSnapshotWaitTimeout time.Duration = 15 * time.Minute
 	defaultWaitInterval      time.Duration = 10 * time.Second
+
+	enableClusterDomainTests = "ENABLE_CLUSTER_DOMAIN_TESTS"
 )
 
 var nodeDriver node.Driver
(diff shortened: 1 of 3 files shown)


test/integration_test/common_test.go
```diff
@@ -98,9 +101,11 @@ func TestMain(t *testing.T) {
 	}
 	// we can either run cluster domain tests or the other tests
 	// the underlying setup of clusters is different
-	if os.Getenv("ENABLE_CLUSTER_DOMAIN_TESTS") != "" {
+	if _, err := strconv.ParseBool(os.Getenv(enableClusterDomainTests)); err == nil {
```
disrani-px: You also have to check the value of the bool ie the first return value from `ParseBool`

test/integration_test/test-deploy.sh
```diff
@@ -5,6 +5,7 @@ snapshot_scale=10
 image_name="stork:master"
 test_image_name="stork_test:latest"
 remote_config_path=""
+run_cluster_domain_test=true
```
disrani-px: Set to `false` by default

test/integration_test/common_test.go
Rohit-PX: Fixed.

test/integration_test/test-deploy.sh
Rohit-PX: Fixed.

test/integration_test/common_test.go
```diff
@@ -98,9 +101,13 @@ func TestMain(t *testing.T) {
 	}
 	// we can either run cluster domain tests or the other tests
 	// the underlying setup of clusters is different
-	if os.Getenv("ENABLE_CLUSTER_DOMAIN_TESTS") != "" {
-		t.Run("ClusterDomains", testClusterDomains)
+	if enabled, err := strconv.ParseBool(os.Getenv(enableClusterDomainTests)); err == nil {
+		if enabled {
```
disrani-px: If enabled is false this won't run any test

Rohit-PX: Fixed.

PR #431, opened 2019-08-02
Validate snapshot before inplace restore

ram-infrac:
**What type of PR is this?**
>bug

Add check to validate whether given snapshot is complete before performing
 in place restore

### DIFF
--- a/pkg/snapshot/controllers/snapshotrestore.go
+++ b/pkg/snapshot/controllers/snapshotrestore.go
@@ -26,7 +26,9 @@ const (
 	annotationPrefix   = "stork.libopenstorage.org/"
 	storkSchedulerName = "stork"
 	// RestoreAnnotation for pvc which has in-place resotre in progress
-	RestoreAnnotation = annotationPrefix + "restore-in-progress"
+	RestoreAnnotation            = annotationPrefix + "restore-in-progress"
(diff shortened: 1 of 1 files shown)


pkg/snapshot/controllers/snapshotrestore.go
```diff
@@ -155,6 +157,12 @@ func (c *SnapshotRestoreController) handleInitial(snapRestore *stork_api.VolumeS
 			return fmt.Errorf("unable to get get snapshot  details %s: %v",
 				snapName, err)
 		}
+		if err := k8s.Instance().ValidateSnapshot(snapName,
```
disrani-px: We shouldn't retry here. Just check if the snapshot is bound. If not raise an event and return nil, it'll be retried in the next reconcile loop

ram-infrac: bound meaning whether spec.SnapshotDataName is not empty ?

disrani-px: Set the 3rd param for `ValidateSnaphshot()` to false to not retry

ram-infrac: set `retry` to false

PR #434, opened 2019-08-06
Cleanup restore objects upon CRD delete

ram-infrac:
**What type of PR is this?**
>bug

This PR does cleanup of restore objects upon restore CRD failure 
  - fixes issue where restore fails if haUpdate fails for restore vol
  - make task add more unique
  - cleanup restore object if restore failed


drivers/volume/portworx/portworx.go
```diff
@@ -992,6 +992,31 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+// CleanupSnapshotRestoreObjects deletes restore objects if any
+func (p *portworx) CleanupSnapshotRestoreObjects(snapRestore *stork_crd.VolumeSnapshotRestore) error {
+	logrus.Infof("Cleaning up in-place restore objects")
+	volRestores, _, _, err := processRestoreVolumes(snapRestore.Status.RestoreVolumes)
+	if err != nil {
+		log.VolumeSnapshotRestoreLog(snapRestore).Errorf("Unable to process volume information %v", err)
+		return err
+	}
+	for volID := range volRestores {
+		logrus.Infof("Deleting volume %v", volID)
```
disrani-px: Use `log.VolumeSnapshotRestoreLog()`

drivers/volume/portworx/portworx.go
```diff
@@ -1026,7 +1051,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 				}
 			}
 			replNodes := vols[0].GetReplicaSets()[0]
-			taskID := restoreTaskPrefix + snapID + "-" + volID
+			taskID := restoreTaskPrefix + snapID + "-" + volID + string(snapRestore.GetUID())
```
disrani-px: Generate the taskID and volume name in a separate function and use them here and in `CleanupSnapshotRestoreObjects`

drivers/volume/portworx/portworx.go
```diff
@@ -1026,7 +1051,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
...
 			restoreName := restoreNamePrefix + volID
```
disrani-px: `restoreName` also needs to have the UID from the restore object to be unique.

drivers/volume/portworx/portworx.go
disrani-px: Don't need the `snapID` in `taskID`, it is pretty long. `volID + restoreUID` should make it unique

drivers/volume/portworx/portworx.go
```diff
@@ -992,6 +992,31 @@ func (p *portworx) SnapshotDelete(snapDataSrc *crdv1.VolumeSnapshotDataSource, _
...
+		return err
+	}
+	for volID := range volRestores {
+		logrus.Infof("Deleting volume %v", volID)
+		// Get volume client from user context
+		volDriver, err := p.getUserVolDriver(snapRestore.Annotations)
+		if err != nil {
+			return err
+		}
+		// Delete restore volume
```
disrani-px: Also need to cancel any cloudsnap restores that might be in progress.

drivers/volume/portworx/portworx.go
```diff
@@ -1074,7 +1099,7 @@ func (p *portworx) GetVolumeSnapshotRestoreStatus(snapRestore *stork_crd.VolumeS
 		return nil
 	}
 	for _, volInfo := range snapRestore.Status.Volumes {
-		taskID := restoreTaskPrefix + volInfo.Snapshot + "-" + volInfo.Volume
+		taskID := restoreTaskPrefix + volInfo.Snapshot + "-" + volInfo.Volume + string(snapRestore.GetUID())
```
disrani-px: The return from `checkAndUpdateHaLevel` still needs to be fixed. If it returns err we shouldn't set the status to Failed. It should return err so that the controller can retry.

PR #436, opened 2019-08-07
Disable CSRestore integration tests

ram-infrac:
Signed-off-by: Ram <ram.suradkar@portworx.com>

**What type of PR is this?**
>integration-test

Disables CS restore integration test

```release-note

```


test/integration_test/snapshot_restore_test.go
```diff
@@ -13,8 +13,10 @@ import (
 func testSnapshotRestore(t *testing.T) {
 	t.Run("simpleSnapshotRestoreTest", simpleSnapshotRestoreTest)
 	t.Run("groupSnapshotRestoreTest", groupSnapshotRestoreTest)
```
ram-infrac: Should I just add `test.Ski`p rather than commenting out  test cases that way static check will not cause issue ?

disrani-px: Yes, please use `test.Skip`

PR #439, opened 2019-08-12
Disable CS inplace restore feature

ram-infrac:

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -1002,6 +1002,16 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1002,6 +1002,12 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 		return err
 	}
 
+	if snapType == crdv1.PortworxSnapshotTypeCloud {
+		log.VolumeSnapshotRestoreLog(snapRestore).Errorf("CloudSnapshot restore not supported")
+		snapRestore.Status.Status = stork_crd.VolumeSnapshotRestoreStatusFailed
+		return fmt.Errorf("not supported")
```
disrani-px: Return `ErrNotSupported` with `Feature` and `Reason` set

PR #441, opened 2019-08-14
Cancel and fail backup if starting volume backup fails

disrani-px:
**What type of PR is this?**
Bug

No

No

No

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"path/filepath"
 	"reflect"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -266,6 +266,22 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 				v1.EventTypeWarning,
 				string(stork_api.ApplicationBackupStatusFailed),
 				message)
+
+			// Cancel any backups that might have started and mark the backup as failed
+			err := a.Driver.CancelBackup(backup)
+			if err != nil {
+				message := fmt.Sprintf("Error cancelling ApplicationBackup for volumes: %v", err)
+				log.ApplicationBackupLog(backup).Errorf(message)
```
ram-infrac: if driver is unable to cancel backup do we need to retry here to cancel backup and then mark backup as failed

disrani-px: Added exponential backoff for cancel

PR #447, opened 2019-08-21
Provisioner flag

Rohit-PX:
**What type of PR is this?**
integration-test

Enable integration tests to use configurable provisioner for volumes

no

no

no

### DIFF
--- a/test/integration_test/test-deploy.sh
+++ b/test/integration_test/test-deploy.sh
@@ -7,6 +7,7 @@ test_image_name="stork_test:latest"
 remote_config_path=""
 run_cluster_domain_test=false
 environment_variables=""
+storage_provisioner="portworx"
 for i in "$@"
 do
 case $i in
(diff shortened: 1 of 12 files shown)


test/integration_test/test-deploy.sh
```diff
@@ -7,6 +7,7 @@ test_image_name="stork_test:latest"
 remote_config_path=""
 run_cluster_domain_test=false
 environment_variables=""
+storage_provisioner=""
```
disrani-px: Initialize this to portworx, so you don't need the `if/else` below

Rohit-PX: Fixed.

test/integration_test/test-deploy.sh
```diff
@@ -129,6 +136,10 @@ else
 	sed -i 's/'enable_cluster_domain'/'""'/g' /testspecs/stork-test-pod.yaml
 fi
 
+if [ "$storage_provisioner" != "portworx" ] ; then
```
disrani-px: You shouldn't use the `if` here, just always set it to the variable, otherwise, if it is set to `portworx` it'll end up not updating the spec

Rohit-PX: Fixed.

PR #460, opened 2019-09-10
Migrate image pull secrets associated with default service account

ram-infrac:
**What type of PR is this?**
>bug

We don't maintain link between secrets while migrating `default` service account. This PR check and update imagepullSecrets for default service account

No

yes

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -721,14 +721,25 @@ func (m *MigrationController) checkAndUpdateDefaultSA(
 		return nil
 	}
 
-	log.MigrationLog(migration).Infof("Updating default service account with image pull secrets")
+	log.MigrationLog(migration).Infof("Updating default service account(namespace : %v) with image pull secrets", sourceSA.GetNamespace())
 	// merge service account resource for default namespaces
 	destSA, err := adminClient.CoreV1().ServiceAccounts(sourceSA.GetNamespace()).Get(sourceSA.GetName(), metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
-	destSA.ImagePullSecrets = append(destSA.ImagePullSecrets, sourceSA.ImagePullSecrets...)
+	for _, s := range sourceSA.ImagePullSecrets {
+		found := false
+		for _, d := range destSA.ImagePullSecrets {
+			if d.Name == s.Name {
+				found = true
+				break
+			}
+		}
+		if !found {
+			destSA.ImagePullSecrets = append(destSA.ImagePullSecrets, s)
+		}
+	}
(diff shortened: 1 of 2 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -676,6 +676,66 @@ func (m *MigrationController) updateResourceStatus(
...
+		if err := adminClient.CoreV1().ServiceAccounts(sourceSA.GetNamespace()).Delete(sourceSA.GetName(), &metav1.DeleteOptions{}); err != nil {
+			return err
+		}
+		if _, err := adminClient.CoreV1().ServiceAccounts(sourceSA.GetNamespace()).Create(&sourceSA); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	log.MigrationLog(migration).Infof("Updating default service account with image pull secrets")
```
disrani-px: Print the namespace of the service account too

pkg/migration/controllers/migration.go
```diff
@@ -676,6 +676,66 @@ func (m *MigrationController) updateResourceStatus(
...
+	// merge service account resource for default namespaces
+	destSA, err := adminClient.CoreV1().ServiceAccounts(sourceSA.GetNamespace()).Get(sourceSA.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	destSA.ImagePullSecrets = append(destSA.ImagePullSecrets, sourceSA.ImagePullSecrets...)
```
disrani-px: Can you confirm that this works fine even if the destination service account already has imagePullSecrets with the same name.

ram-infrac: it works even though destination service account already has imagePullSecrets, it just add duplicate entry under image pull secrets 
eg. 
```
Name:                default
Namespace:           default
Labels:              <none>
Annotations:         <none>
Image pull secrets:  test-dr
                     test-dr
Mountable secrets:   default-token-44krf
Tokens:              default-token-44krf
Events:              <none>
```

So I added commit to not duplicated already present `imagepullsecrets`

pkg/migration/controllers/migration.go
```diff
@@ -676,6 +676,77 @@ func (m *MigrationController) updateResourceStatus(
...
+	}
+
+	log.MigrationLog(migration).Infof("Updating default service account(namespace : %v) with image pull secrets", sourceSA.GetNamespace())
+	// merge service account resource for default namespaces
+	destSA, err := adminClient.CoreV1().ServiceAccounts(sourceSA.GetNamespace()).Get(sourceSA.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	found := false
```
disrani-px: `found` needs to be initialized to `false` inside the first loop below

ram-infrac: yeah, missed it

PR #464, opened 2019-09-16
Correct volume count for storkctl snaprestore output

ram-infrac:
**What type of PR is this?**
>bug

Fix volume count for volumesnapshotrestore 

Yes
Earlier volumesnapshotrestore output always show volume to be restore as zero. This PR changes volume count as per volume which are undergoing in place restore

eg.
```
[#vsrestore]# ./storkctl get volumesnapshotrestore
NAME                         SOURCE-SNAPSHOT             SOURCE-SNAPSHOT-NAMESPACE   STATUS       VOLUMES   CREATED
mysql-snap-group-inrestore   mysql-group-localsnapshot   default                     Successful   2         16 Sep 19 04:38 PDT
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -777,14 +777,14 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:d60592e3ef7b03350f15217c6cee5ed591a23730dc194cdec679c7b35fd8fb5d"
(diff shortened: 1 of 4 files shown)


disrani-px:
@ram-infrac Can you add a unit test for this?

PR #466, opened 2019-09-17
Use cassandra instead of mysql for clusterdomain migration.

Rohit-PX:

**What type of PR is this?**
integration-test

Switch cluster domain migration test to cassandra from mysql

no

no

no

### DIFF
--- a/test/integration_test/clusterdomain_test.go
+++ b/test/integration_test/clusterdomain_test.go
@@ -101,10 +101,10 @@ func failoverAndFailbackClusterDomainTest(t *testing.T) {
 	// Migrate the resources
 	ctxs, preMigrationCtx := triggerMigration(
 		t,
-		"mysql-clusterdomain-migration",
-		"mysql-1-pvc",
+		"cassandra-clusterdomain-migration",
+		"cassandra",
 		nil,
-		[]string{"mysql-clusterdomain-migration"},
+		[]string{"cassandra-clusterdomain-migration"},
 		true,
 		true,
 	)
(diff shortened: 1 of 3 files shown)


disrani-px:
The scale factor is being set to 1, it should be whatever was at the source

Rohit-PX:
> The scale factor is being set to 1, it should be whatever was at the source

Creating a copy of the old scale factor map and restoring from that now.

test/integration_test/clusterdomain_test.go
```diff
@@ -146,6 +146,10 @@ func testClusterDomainsFailover(
 	// Reduce the replicas on cluster 1
 	scaleFactor, err := schedulerDriver.GetScaleFactorMap(ctxs[0])
 	require.NoError(t, err, "Unexpected error on GetScaleFactorMap")
+
+	// Copy the old scale factor map
+	oldScaleFactor := scaleFactor
```
disrani-px: This won't make a copy of the map, so oldScaleFactor will also have 0s for the scale after the for loop below

Rohit-PX: Fixed.

test/integration_test/clusterdomain_test.go
```diff
@@ -156,11 +163,11 @@ func testClusterDomainsFailover(
 	err = setRemoteConfig(remoteFilePath)
 	require.NoError(t, err, "Error setting remote config")
 
-	// Increase the replicas on cluster 1
+	// Set scale factor to it's orignal values on cluster 1
 	scaleFactor, err = schedulerDriver.GetScaleFactorMap(preMigrationCtx)
 	require.NoError(t, err, "Unexpected error on GetScaleFactorMap")
 	for k := range scaleFactor {
```
disrani-px: Don't need this loop, just pass in oldScaleFactor to scale up

Rohit-PX: Fixed.

PR #470, opened 2019-09-19
Order snapshot migration tests

Rohit-PX:
**What type of PR is this?**
integration-test

Cloudsnaps tests fail if they are run after the migration tests.

no

no

no

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -65,6 +65,12 @@ var storkVolumeDriver storkdriver.Driver
 
 var snapshotScaleCount int
 
+func TestSnapshotMigration(t *testing.T) {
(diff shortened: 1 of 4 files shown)


test/integration_test/common_test.go
```diff
@@ -65,6 +65,12 @@ var storkVolumeDriver storkdriver.Driver
 
 var snapshotScaleCount int
 
+func TestSnapshotMigration(t *testing.T) {
+	t.Run("testSnapshot", testSnapshot)
+	t.Run("testSnapshot", testSnapshotRestore)
```
disrani-px: `testSnapshotRestore`

Rohit-PX: fixed.

PR #472, opened 2019-09-20
Delete VolumeAttachments when node driver is offline

ggriffiths:
**What type of PR is this?**
>bug

*  If a Driver Node is down, or a Pod is found to be in an Unknown state (Node is offline), we will delete it's volume attachments before attempting to delete the pod.
* ~This PR requires https://github.com/portworx/sched-ops/pull/152 to be merged first. After that, I'll update the vendor here correctly.~ done

No

No

Yes, 2.3


pkg/monitor/monitor.go
```diff
@@ -177,6 +183,12 @@ func (m *Monitor) driverMonitor() {
 							continue
 						}
 
+						// delete volume attachments if the node is down for this pod
+						err = m.cleanupVolumeAttachments(&pod)
```
disrani-px: This will delete all the pods in the cluster. Needs be inside the if condition below which checks if the pod is on the node where PX is offline.

disrani-px: Actually, this should be outside the loop for the pods. If the storage driver is offline we should go through all the volumeattachments and delete the ones that are attached to that node. So even if someone deletes a pod manually from a node where the storage is offline this will clean it up.

disrani-px:
Can you also add a UT.

Create multiple pods on different nodes, some with volumeattachments, some without. Stop the driver on the node with the attachment and make sure only that pod and volumeattachment are deleted.

ggriffiths:
> Can you also add a UT.
> 
> Create multiple pods on different nodes, some with volumeattachments, some without. Stop the driver on the node with the attachment and make sure only that pod and volumeattachment are deleted.

Yep will do.

pkg/monitor/monitor.go
ggriffiths: Makes sense - will change

ggriffiths: Changed to two different functions-
1. For podMonitor - `cleanupVolumeAttachmentsByPod`, which deletes all VolumeAttachment for a given pod.
2. For driverMonitor - `cleanupVolumeAttachmentsByNode`, which deletes all VolumeAttachments for a given node.

ggriffiths: I ran this lastest code on my cluster and I noticed that the VolumeAttachments weren't continuously being deleted/recreated on every `driverMonitor` loop as they were before.

So, this fix handles both cases (driverMonitor and podMonitor)

pkg/monitor/monitor_test.go
```diff
@@ -277,3 +283,62 @@ func testOfflineStorageNodeDuplicateIP(t *testing.T) {
...
+	// Kill N1
+	err = driver.UpdateNodeStatus(1, volume.NodeOffline)
+	require.NoError(t, err, "Error setting node status to Offline")
+	defer func() {
+		err = driver.UpdateNodeStatus(0, volume.NodeOnline)
+		require.NoError(t, err, "Error setting node status to Online")
+	}()
+
+	// VolumeAttachments (VA) for N1 should be deleted, but VA for N0 should remain.
+	// HOWEVER, we can't currently test this because I didn't add CreateVolumeAttachment to sched-ops
```
disrani-px: Can you add that? The test isn't really testing the fix without that

ggriffiths: Yeah, in the process of adding it now

ggriffiths:
Updated UTs to make them test VolumeAttachments @disrani-px

disrani-px:
@ggriffiths looks like coverage is missing for the code path from the node monitor

ggriffiths:
Looks like it's actually missing from cleanupVolumeAttachmentsByPod (podMonitor).

I'll add another two cases:

Pod in unknown state (attached)
Pod in unknown state (detached)

ggriffiths:
Added the above cases. Monitor pkg is now at 78.7% cov (was 63% before) on my machine:

PASS
coverage: 78.7% of statements

PR #473, opened 2019-09-20
Fix CSI unstructured object access

lpabon:
**What type of PR is this?**
Bug

Received error:

```
Error preparing resource: error preparing PV resource pvc-1678c28f-da63-11e9-94db-000c2910432d: .spec.csi accessor error: contains non-string key in the map: map[attached:ATTACH_STATE_INTERNAL_SWITCH state:VOLUME_STATE_DETACHED readonly:false secure:false shared:false sharedv4:false storage.kubernetes.io/csiProvisionerIdentity:1568735171590-8081-pxd.portworx.com error: parent:] is of the type map[string]interface {}, expected string
```

No

No

Not sure, I do not think the CSI support has been released yet.

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2051,25 +2051,22 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
 	}
 
 	// Get access to the csi section of the PV
-	csiSpec, found, err := unstructured.NestedStringMap(object.UnstructuredContent(), "spec", "csi")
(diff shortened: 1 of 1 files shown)


lpabon:
Tested migration of two Kubernetes 1.14.9 clusters setup with Portworx CSI on both.

```
$ storkctl get migration -n default
NAME             CLUSTERPAIR     STAGE     STATUS       VOLUMES   RESOURCES   CREATED               ELAPSED
mysqlmigration   remotecluster   Final     Successful   1/1       3/3         20 Sep 19 08:26 PDT   1m7s
```

PR #494, opened 2019-10-10
Groupsnapshot: If objects exist reuse them if they are the same

disrani-px:
Else delete and re-create them

**What type of PR is this?**
Bug

No

```release-note
- Fix an issue with groupsnapshot where a failure during creating snapshot objects could lead to a stuck groupsnapshot
```

2.3


pkg/groupsnapshot/controllers/groupsnapshot.go
```diff
@@ -477,6 +523,11 @@ func (m *GroupSnapshotController) createSnapAndDataObjects(
 
 		snap, err = k8s.Instance().CreateSnapshot(snap)
 		if err != nil {
+			// Try to replace the snapshot if it already exists
+			if errors.IsAlreadyExists(err) {
+				err = m.replaceSnapshot(snap)
```
ram-infrac: why delete snapshot data if we are able to replace snapshot on line 532

disrani-px: Need to skip the delete below if replace succeeds. Thanks for catching that.

disrani-px: @ram-infrac Please take a look again

PR #496, opened 2019-10-14
Migration scale test.

Rohit-PX:

**What type of PR is this?**
integration-test

We need to test scaled migrations across namespaces

no

no

no


test/integration_test/migration_test.go
```diff
@@ -568,3 +575,87 @@ func migrationScheduleTest(
...
+	defer func() {
+		err = setRemoteConfig("")
+		require.NoError(t, err, "Error resetting remote config")
+	}()
+	migrationKey := "mysql-migration"
+	migrationAppKey := "mysql-1-pvc"
+	includeResourcesFlag := true
+
+	for i := 1; i <= migrationScaleCount; i++ {
+		currCtxs, err := schedulerDriver.Schedule(migrationKey+"-"+strconv.Itoa(i),
```
disrani-px: Can you use `triggerMigration()` here? You should be able to do the same thing with that.

test/integration_test/migration_test.go
```diff
@@ -568,3 +575,87 @@ func migrationScheduleTest(
...
+
+		ctxs = append(ctxs, currCtxs...)
+
+		currMigNamespace := migrationAppKey + "-" + migrationKey + "-" + strconv.Itoa(i)
+		currMig, err := createMigration(t, migrationKey, currMigNamespace, "remoteclusterpair", currMigNamespace, &includeResourcesFlag, true)
+		logrus.Infof("Successfully created migration %s:%s", currMig.Name, currMig.Namespace)
+		allMigrations = append(allMigrations, currMig)
+
+	}
+	err = WaitForMigration(allMigrations)
```
disrani-px: You should be able to loop on all the migrations and call `validateAndDestroyMigration()` which will also take care of deleting the migrations after they succeed.

Rohit-PX:
```time="2019-10-15T22:31:41Z" level=info msg="[mysql-1-pvc] Destroyed storage class: px-mysql-sc"
--- PASS: TestSnapshotMigration (625.56s)
    --- PASS: TestSnapshotMigration/testMigration (625.56s)
        --- PASS: TestSnapshotMigration/testMigration/scaleTest (622.16s)
PASS

DONE 3 tests in 627.087s```

test/integration_test/migration_test.go
Rohit-PX: As discussed, trigger migrations will cause same migration to be created everytime, so need a separate function to create a new migration everytime.

test/integration_test/migration_test.go
Rohit-PX: Validation also expects same migration to be present, since we have diff migrations we cannot use that here.

PR #501, opened 2019-10-28
Update API version to v1.

Rohit-PX:

**What type of PR is this?**
api-change

Update api versions from v1beta to v1 since v1beta will be deprecated

no

no

no

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,9 @@ spec:
         version: "1"
     spec:
       schedulerName: stork
+      selector:
+        matchLabels:
+          name: stork-scheduler
       containers:
       - image: mysql:5.6
         name: mysql
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ provisioner: kubernetes.io/portworx-volume
 parameters:
    repl: "2"
 ---
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: mysql
--- a/README.md
+++ b/README.md
@@ -205,6 +205,9 @@ spec:
       maxUnavailable: 1
     type: RollingUpdate
   replicas: 1
+  selector:
+    matchLabels:
+      app: mysql
   template:
     metadata:
       labels:
@@ -212,9 +215,6 @@ spec:
         version: "1"
     spec:
       schedulerName: stork
-      selector:
-        matchLabels:
-          name: stork-scheduler
       containers:
       - image: mysql:5.6
         name: mysql
--- a/examples/groupvolumesnapshots/cassandra-e2e/cassandra-ss.yaml
+++ b/examples/groupvolumesnapshots/cassandra-e2e/cassandra-ss.yaml
@@ -1,6 +1,6 @@
 ##### Portworx storage class
 kind: StorageClass
-apiVersion: storage.k8s.io/v1beta1
+apiVersion: storage.k8s.io/v1
 metadata:
     name: portworx-repl2
 provisioner: kubernetes.io/portworx-volume
@@ -22,7 +22,7 @@ spec:
 
 ---
 
-apiVersion: "apps/v1beta1"
+apiVersion: "apps/v1"
 kind: StatefulSet
 metadata:
   name: cassandra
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -194,7 +194,7 @@ provisioner: kubernetes.io/portworx-volume
 parameters:
    repl: "2"
 ---
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
```
disrani-px: You'll also need the selector under spec.

Rohit-PX: Added selector for labels.

rodrigc: The selector you added is in the wrong place.  If I try to apply this yaml I get:

```
persistentvolumeclaim/mysql-data created
storageclass.storage.k8s.io/px-mysql-sc created
error: error validating "a.yml": error validating data: [ValidationError(Deployment.spec.template.spec): unknown field "selector" in io.k8s.api.core.v1.PodSpec, ValidationError(Deployment.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec]; if you choose to ignore these errors, turn validation off with --validate=false
```

rodrigc: You need to do something like this:

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
   name: mysql-data
   annotations:
     volume.beta.kubernetes.io/storage-class: px-mysql-sc
spec:
   accessModes:
     - ReadWriteOnce
   resources:
     requests:
       storage: 2Gi
---
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
    name: px-mysql-sc
provisioner: kubernetes.io/portworx-volume
parameters:
   repl: "2"
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mysql
spec:
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  replicas: 1
  selector:
    matchLabels:
      app: mysql  # needs to match spec.template.metadata.labels
  template:
    metadata:
      labels:
        app: mysql
        version: "1"
    spec:
      schedulerName: stork
      containers:
      - image: mysql:5.6
        name: mysql
        env:
        - name: MYSQL_ROOT_PASSWORD
          value: password
        ports:
        - containerPort: 3306
        volumeMounts:
        - name: mysql-persistent-storage
          mountPath: /var/lib/mysql
      volumes:
      - name: mysql-persistent-storage
        persistentVolumeClaim:
          claimName: mysql-data
```

rodrigc: See this example: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components

The indentation is wrong in this example, but the comments give you an idea of what you need to do when adding a spec.selector

rodrigc: This is the same example, but with the indentation fixed:
https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/workloads/controllers/statefulset.md#components

Rohit-PX: This has been fixed.

rodrigc: With the latest changes, this file now works with `kubectl apply`

specs/stork-daemonset.yaml
```diff
@@ -83,7 +83,7 @@ spec:
       port: 8099
       targetPort: 8099
 ---
-apiVersion: extensions/v1beta1
```
rodrigc: This file is missing a spec selector and does not apply:

```
error: error validating "stork-daemonset.yml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec
```

examples/groupvolumesnapshots/cassandra-e2e/cassandra-ss.yaml
```diff
@@ -22,7 +22,7 @@ spec:
 
 ---
 
-apiVersion: "apps/v1beta1"
```
rodrigc: This yaml needs a spec selector also:

```
error: error validating "cassandra-ss.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "selector" in io.k8s.api.apps.v1.StatefulSetSpec;
```

specs/stork-daemonset.yaml
Rohit-PX: Fixed.

examples/groupvolumesnapshots/cassandra-e2e/cassandra-ss.yaml
Rohit-PX: Fixed.

PR #502, opened 2019-10-30
Wait backup completion

Rohit-PX:
**What type of PR is this?**documentation
integration-test

In backup sync controller test, wait for backup to finish before proceeding with resync.

no

### DIFF
--- a/test/integration_test/applicationbackup_test.go
+++ b/test/integration_test/applicationbackup_test.go
@@ -746,6 +746,10 @@ func applicationBackupSyncControllerTest(t *testing.T) {
 	firstBackup, err := createApplicationBackupWithAnnotation(t, appKey+"-backup-sync", appCtx.GetID(), backupLocation)
 	require.NoError(t, err, "Error creating app backups")
 
+	// Wait for backup completion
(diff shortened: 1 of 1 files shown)


test/integration_test/applicationbackup_test.go
```diff
@@ -859,6 +863,23 @@ func applicationBackupSyncControllerTest(t *testing.T) {
...
 }
 
+func waitForAppBackupCompletion(name, namespace string) error {
+	getAppBackup := func() (interface{}, bool, error) {
+		appBackup, err := k8s.Instance().GetApplicationBackup(name, namespace)
+		if err != nil {
+			return "", false, err
+		}
+
+		if appBackup.Status.Status == storkv1.ApplicationBackupStatusSuccessful {
```
disrani-px: Can stop retrying if it goes to Failed state

Rohit-PX: I got the condition wrong there it should be 
`if appBackup.Status.Status != storkv1.ApplicationBackupStatusSuccessful` {
.............
.............
}`

PR #505, opened 2019-10-31
update ha level of volume using pool uuid instead of nodes

ram-infrac:
**What type of PR is this?**
>feature

Add support for cs inplace restore feature
 - pass pooluuid instead of node id to ha increase

No

```release-note

```

yes, 2.3.0


drivers/volume/portworx/portworx.go
```diff
@@ -1055,16 +1053,6 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 		return err
 	}
 
-	if snapType == crdv1.PortworxSnapshotTypeCloud {
```
disrani-px: We should add a version check here to only support this from PX 2.3.0 onwards

ram-infrac: this one was unnecessary , we have that on https://github.com/libopenstorage/stork/pull/505/files/bed5a31a466da0fde2427037658b028ea9c787d1#diff-0aa9aa57d00bdd00603aae645329fb04R1061

disrani-px: Ok, that should be changed to 2.3.0

ram-infrac: ok

drivers/volume/portworx/portworx.go
```diff
@@ -1118,7 +1105,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 				ID:                snapID,
 				RestoreVolumeName: restoreName,
 				CredentialUUID:    credID,
-				NodeID:            replNodes.GetNodes()[0],
+				NodeID:            replNodes.GetPoolUuids()[0],
```
disrani-px: Should be `PoolID`?

drivers/volume/portworx/portworx.go
```diff
@@ -1055,35 +1053,24 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
...
-			Reason:  "Feature not supported",
-		}
-		return err
-	}
-
 	switch snapType {
 	case "", crdv1.PortworxSnapshotTypeLocal:
 		return nil
 	case crdv1.PortworxSnapshotTypeCloud:
 		// refactor once cloudsnap is supported
```
disrani-px: Stale comment?

drivers/volume/portworx/portworx.go
talakad: No.  Should be PoolUUID

disrani-px: Yeah, I meant it shouldn't be NodeID

ram-infrac: @disrani-px , in CloudRestore() it is expected to pass `pooluuid` in nodeID field

drivers/volume/portworx/portworx.go
ram-infrac: yes

drivers/volume/portworx/portworx.go
disrani-px: @talakad Passing in poolID for the nodeID seems wrong. Can we fix that? Don't need to hold up this PR for that though.

PR #514, opened 2019-11-08
fixes for groupcsrestore

ram-infrac:
- correct error msg for restore failure
  - pass only one pool while doing haupdate
  - add debugging for correct  poolids

Signed-off-by: Ram <ram.suradkar@portworx.com>

**What type of PR is this?**
>bug

2.3.0

No
no

2.3.0

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -1057,7 +1057,6 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 	case "", crdv1.PortworxSnapshotTypeLocal:
 		return nil
 	case crdv1.PortworxSnapshotTypeCloud:
-		// refactor once cloudsnap is supported
 		ok, msg, err := p.ensureNodesHaveMinVersion("2.3.0")
 		if err != nil {
 			return err
@@ -1085,7 +1084,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 
 			// Get Volume Info
 			uid := getUidforRestore(volID, string(snapRestore.GetUID()))
-			vols, err := volDriver.Inspect([]string{volID, restoreNamePrefix + uid})
+			vols, err := volDriver.Inspect([]string{volID})
 			if err != nil {
 				return &ErrFailedToInspectVolume{
 					ID:    volID,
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -1100,6 +1100,7 @@ func (p *portworx) StartVolumeSnapshotRestore(snapRestore *stork_crd.VolumeSnaps
 			replNodes := vols[0].GetReplicaSets()[0]
 			taskID := restoreTaskPrefix + uid
 			restoreName := restoreNamePrefix + uid
+			logrus.Infof("Restoring cloudsnapshot to volume %v on pool %v", restoreName, replNodes.GetPoolUuids()[0])
```
disrani-px: Use VolumeSnapshotRestoreLog for logs

disrani-px:
Can you add [Portworx] to the first line of the commit message since this is a driver only fix

ram-infrac:
yes, @disrani-px  will do

drivers/volume/portworx/portworx.go
```diff
@@ -1212,41 +1212,42 @@ func (p *portworx) checkAndUpdateHaLevel(volID, uid string, params map[string]st
...
+			replPool = pool
+			break
 		}
 	}
 
 	if restoreVol.GetSpec().GetHaLevel() != parentVols[0].GetSpec().GetHaLevel() {
-		logrus.Infof("Updating HA Level of volume %v", restoreVol.GetLocator().GetName())
+		logrus.Infof("Updating HA Level of volume %v to pool %v", restoreVol.GetLocator().GetName(), replPool)
 		spec := &api.VolumeSpec{
 			HaLevel:    restoreVol.GetSpec().GetHaLevel() + 1,
```
Rohit-PX: If HA level of parent of volume is 3, would updating HA level to HA level + 1  work ? (it would make it 2 not 3)

ram-infrac: `checkAndUpdateHALevel` is called till ha level of restore volume does not match parent volume. At a time you can only update ha level by 1 of px volume

https://github.com/libopenstorage/stork/pull/514/files/9af000cba70565967f4e0458a6b257d2335c422e#diff-0aa9aa57d00bdd00603aae645329fb04R1176

ram-infrac: https://github.com/libopenstorage/stork/blob/9af000cba70565967f4e0458a6b257d2335c422e/drivers/volume/portworx/portworx.go#L1155

PR #515, opened 2019-11-11
Wait for snapshot to complete in in-place restore tests

Rohit-PX:

**What type of PR is this?**
integration-test

Currently, inplace restore tests don't wait on volume snaps to complete before proceeding to
volume restore

no

no

no

### DIFF
--- a/test/integration_test/snapshot_restore_test.go
+++ b/test/integration_test/snapshot_restore_test.go
@@ -79,6 +79,7 @@ func cloudSnapshotRestoreTest(t *testing.T) {
 		scheduler.ScheduleOptions{AppKeys: []string{"mysql-cloudsnap-restore"}})
 	require.NoError(t, err, "Error scheduling task")
 	require.Equal(t, 1, len(snapCtx), "Only one task should have started")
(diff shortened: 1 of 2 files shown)


test/integration_test/snapshot_restore_test.go
```diff
@@ -79,6 +79,7 @@ func cloudSnapshotRestoreTest(t *testing.T) {
 		scheduler.ScheduleOptions{AppKeys: []string{"mysql-cloudsnap-restore"}})
 	require.NoError(t, err, "Error scheduling task")
 	require.Equal(t, 1, len(snapCtx), "Only one task should have started")
+	verifyGroupSnapshot(t, snapCtx[0], defaultWaitTimeout)
```
disrani-px: This is not a group snapshot

Rohit-PX: The API name needs to be changed here, it just inspects the volumes from the context without differentiating between group or single volume snapshots.

Rohit-PX: Changed method to validate snapshot in this case.

PR #522, opened 2019-11-18
Cleanup deleted objects from destination during migration

ram-infrac:
**What type of PR is this?**\
>feature

With this user can able to cleanup deleted objects from destination during migration

yes, we have `AllowCleaningResources` flag to cleanup resources
no

yes, 2.30

### DIFF
--- a/pkg/apis/stork/v1alpha1/migration.go
+++ b/pkg/apis/stork/v1alpha1/migration.go
@@ -13,16 +13,16 @@ const (
 
 // MigrationSpec is the spec used to migrate apps between clusterpairs
 type MigrationSpec struct {
-	ClusterPair            string            `json:"clusterPair"`
-	AdminClusterPair       string            `json:"adminClusterPair"`
-	Namespaces             []string          `json:"namespaces"`
-	IncludeResources       *bool             `json:"includeResources"`
-	IncludeVolumes         *bool             `json:"includeVolumes"`
-	StartApplications      *bool             `json:"startApplications"`
-	AllowCleaningResources *bool             `json:"allowCleaningResources"`
-	Selectors              map[string]string `json:"selectors"`
-	PreExecRule            string            `json:"preExecRule"`
-	PostExecRule           string            `json:"postExecRule"`
+	ClusterPair           string            `json:"clusterPair"`
+	AdminClusterPair      string            `json:"adminClusterPair"`
+	Namespaces            []string          `json:"namespaces"`
+	IncludeResources      *bool             `json:"includeResources"`
+	IncludeVolumes        *bool             `json:"includeVolumes"`
+	StartApplications     *bool             `json:"startApplications"`
+	PurgeDeletedResources *bool             `json:"purgeDeletedResources"`
+	Selectors             map[string]string `json:"selectors"`
+	PreExecRule           string            `json:"preExecRule"`
+	PostExecRule          string            `json:"postExecRule"`
 }
 
 // MigrationStatus is the status of a migration operation
@@ -81,8 +81,8 @@ const (
 	MigrationStatusPartialSuccess MigrationStatusType = "PartialSuccess"
 	// MigrationStatusSuccessful for when migration has completed successfully
 	MigrationStatusSuccessful MigrationStatusType = "Successful"
-	// MigrationStatusCleaned for when migration objects has been deleted
-	MigrationStatusCleaned MigrationStatusType = "Cleaned"
+	// MigrationStatusPurged for when migration objects has been deleted
+	MigrationStatusPurged MigrationStatusType = "Purged"
 )
 
 // MigrationStageType is the stage of the migration
--- a/pkg/apis/stork/v1alpha1/migration.go
+++ b/pkg/apis/stork/v1alpha1/migration.go
@@ -13,15 +13,16 @@ const (
 
 // MigrationSpec is the spec used to migrate apps between clusterpairs
 type MigrationSpec struct {
-	ClusterPair       string            `json:"clusterPair"`
-	AdminClusterPair  string            `json:"adminClusterPair"`
-	Namespaces        []string          `json:"namespaces"`
-	IncludeResources  *bool             `json:"includeResources"`
-	IncludeVolumes    *bool             `json:"includeVolumes"`
-	StartApplications *bool             `json:"startApplications"`
-	Selectors         map[string]string `json:"selectors"`
-	PreExecRule       string            `json:"preExecRule"`
-	PostExecRule      string            `json:"postExecRule"`
+	ClusterPair            string            `json:"clusterPair"`
+	AdminClusterPair       string            `json:"adminClusterPair"`
+	Namespaces             []string          `json:"namespaces"`
+	IncludeResources       *bool             `json:"includeResources"`
+	IncludeVolumes         *bool             `json:"includeVolumes"`
+	StartApplications      *bool             `json:"startApplications"`
+	AllowCleaningResources *bool             `json:"allowCleaningResources"`
+	Selectors              map[string]string `json:"selectors"`
+	PreExecRule            string            `json:"preExecRule"`
+	PostExecRule           string            `json:"postExecRule"`
 }
 
 // MigrationStatus is the status of a migration operation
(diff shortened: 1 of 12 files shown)


pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -13,15 +13,16 @@ const (
...
-	Selectors         map[string]string `json:"selectors"`
-	PreExecRule       string            `json:"preExecRule"`
-	PostExecRule      string            `json:"postExecRule"`
+	ClusterPair            string            `json:"clusterPair"`
+	AdminClusterPair       string            `json:"adminClusterPair"`
+	Namespaces             []string          `json:"namespaces"`
+	IncludeResources       *bool             `json:"includeResources"`
+	IncludeVolumes         *bool             `json:"includeVolumes"`
+	StartApplications      *bool             `json:"startApplications"`
+	AllowCleaningResources *bool             `json:"allowCleaningResources"`
```
disrani-px: `PurgeDeletedResources`

pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -80,6 +81,8 @@ const (
 	MigrationStatusPartialSuccess MigrationStatusType = "PartialSuccess"
 	// MigrationStatusSuccessful for when migration has completed successfully
 	MigrationStatusSuccessful MigrationStatusType = "Successful"
+	// MigrationStatusCleaned for when migration objects has been deleted
+	MigrationStatusCleaned MigrationStatusType = "Cleaned"
```
disrani-px: Let's call this `Purged`, cleaned is not clear enough.

pkg/migration/controllers/migration.go
```diff
@@ -38,6 +38,9 @@ const (
 	// StorkMigrationReplicasAnnotation is the annotation used to keep track of
 	// the number of replicas for an application when it was migrated
 	StorkMigrationReplicasAnnotation = "stork.libopenstorage.org/migrationReplicas"
+	// StorkMigrationAnnotation is the annotation used to keep track of resources
+	// migrated by stork
+	StorkMigrationAnnotation = "stork.libopenstorage.org/storkMigration"
```
disrani-px: `stork.libopenstorage.org/migrated` should be good. Don't need `stork` in the final key, it is in the first part of the annotation already.

pkg/migration/controllers/migration.go
```diff
@@ -614,6 +705,20 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration) e
 	if err != nil {
 		return err
 	}
+	logrus.Infof("Cleaning up migrated resources")
```
disrani-px: Don't think we need this. If you want to keep it use `log.MigrationLog(migration).Infof(...)`

pkg/migration/controllers/migration.go
```diff
@@ -614,6 +705,20 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration) e
...
+	// delete resources on destination which are not present on source
+	// TODO: what should be idle location for this
```
disrani-px: This should be before calling `sdk.Update()` above. Otherwise if stork crashes after updating the object the cleanup will never get called on restart

pkg/migration/controllers/migration.go
```diff
@@ -884,6 +989,10 @@ func (m *MigrationController) applyResources(
 		if !ok {
 			return fmt.Errorf("unable to cast object to unstructured: %v", o)
 		}
+
+		unstructured.SetAnnotations(map[string]string{StorkMigrationAnnotation: "true"})
```
disrani-px: This will wipe out all the other annotations in the objects. You have the `metadata` for the object above. Get the annotations for the objects and add this to those.

As a separate change, we should also add an annotation with the timestamp and name of the migration.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -62,6 +69,9 @@ func (r *ResourceCollector) Init() error {
 	if err != nil {
 		return err
 	}
+
+	// reset k8s instance to given cluster config
+	k8s.Instance().SetConfig(config)
```
disrani-px: We can't do this since that updates the singleton and it'll cause all the sched-ops APIs to go to the remote cluster if config is passed in.

We need to use `k8s.NewInstance()` instead and use that in GetResources, and initialize the clients in `GetResources`

pkg/resourcecollector/resourcecollector.go
```diff
@@ -187,12 +205,24 @@ func (r *ResourceCollector) objectToBeCollected(
...
+	destCluster bool,
 ) (bool, error) {
 	metadata, err := meta.Accessor(object)
 	if err != nil {
 		return false, err
 	}
 
+	// check migration resources
+	if destCluster {
+		if val, ok := metadata.GetAnnotations()[storkMigrationAnnotation]; ok {
```
disrani-px: I think the `ResourceCollector` package shouldn't know about this, it should just return all the objects and the migration controller can decide which ones to filter out.

pkg/migration/controllers/migration.go
ram-infrac: I had sdk.Update seperately in `cleanupMigratedResources` since proper location of cleanup was not decided, but we should have only one sdk.update as you suggested

pkg/resourcecollector/resourcecollector.go
ram-infrac: we use `k8s.Instance()` while collecting resources for eg. `pvcResourceCollector`

pkg/migration/controllers/migration.go
ram-infrac: good catch , will update

pkg/resourcecollector/resourcecollector.go
```diff
@@ -21,30 +21,38 @@ import (
...
 )
 
 const (
 	// Annotation to use when the resource shouldn't be collected
-	skipResourceAnnotation = "stork.libopenstorage.org/skipresource"
-
-	deletedMaxRetries    = 12
-	deletedRetryInterval = 10 * time.Second
+	skipResourceAnnotation   = "stork.libopenstorage.org/skipresource"
+	storkMigrationAnnotation = "stork.libopenstorage.org/storkMigration"
```
disrani-px: Don't need this, not being used anymore

pkg/resourcecollector/resourcecollector.go
```diff
@@ -92,18 +106,22 @@ func resourceToBeCollected(resource metav1.APIResource) bool {
 }
 
 // GetResources gets all the resources in the given list of namespaces which match the labelSelectors
-func (r *ResourceCollector) GetResources(namespaces []string, labelSelectors map[string]string) ([]runtime.Unstructured, error) {
-	err := r.discoveryHelper.Refresh()
+func (r *ResourceCollector) GetResources(namespaces []string, labelSelectors map[string]string, remoteConfig *restclient.Config) ([]runtime.Unstructured, error) {
+	err := r.initClusterConfig(remoteConfig)
```
disrani-px: We can't do this either, the same instance of `ResourceCollector` is being used by different modules. You need to call `k8s.NewInstanceFromRestConfig()` from here and use the sched-ops instance instead of setting it as `r.k8sOps`

pkg/migration/controllers/migration.go
```diff
@@ -293,6 +303,135 @@ func (m *MigrationController) Handle(ctx context.Context, event sdk.Event) error
...
+	}
+	return objects, nil
+}
+
+func objectTobeDeleted(srcObjects, destObjects []runtime.Unstructured) []runtime.Unstructured {
+	var deleteObjects []runtime.Unstructured
+	for _, o := range destObjects {
+		name, namespace, kind, err := getObjectDetails(o)
+		if err != nil {
+			// skip purging if we are not able to get object details
```
disrani-px: Log a message and raise an event if this fails

pkg/migration/controllers/migration.go
```diff
@@ -293,6 +303,135 @@ func (m *MigrationController) Handle(ctx context.Context, event sdk.Event) error
...
+			continue
+		}
+		isPresent := false
+		logrus.Debugf("Checking if destObject(%v:%v:%v) present on source cluster",
```
disrani-px: Too verbose, we start stork with debug logs by default

PR #523, opened 2019-11-22
Add ability to register controllers with label selector

saheienko:
**What type of PR is this?**
feature

Add ability to register controllers that handle resources with custom labels (server-side filtering).

```release-note

```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -50,13 +50,25 @@ func Run() error {
 	return nil
 }
 
-// Register to get callbacks for updates to objects
-// All handlers need to be registered before calling Run()
+// Register registers a controller with empty label selector.
 func Register(
 	gkv *schema.GroupVersionKind,
 	namespace string,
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -50,13 +50,20 @@ func Run() error {
 	return nil
 }
 
-// Register to get callbacks for updates to objects
-// All handlers need to be registered before calling Run()
-func Register(
+// Register registers a controller with empty label selector.
+func Register(gkv *schema.GroupVersionKind, namespace string, resyncPeriod time.Duration, handler sdk.Handler) error {
```
disrani-px: Line it too long with the params, put them on different lines like below

saheienko: Updated.

disrani-px:
@saheienko Is this still required?

saheienko:
This is an obsolete one. Closing it.

PR #535, opened 2019-12-27
Mutating Webhook to update application scheduler to stork

ram-infrac:
**What type of PR is this?**
>feature

This PR adds mutating Webhook admission controller to stork, which allow stork to update application scheduler name to `stork` if application using volume by drivers supported by stork

No

### DIFF
--- a/pkg/webhookadmission/utils.go
+++ b/pkg/webhookadmission/utils.go
@@ -1,0 +1,141 @@
+package webhookadmission
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"time"
+
+	"github.com/portworx/sched-ops/k8s"
+	log "github.com/sirupsen/logrus"
+	hook "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	webhookName        = "stork.webhook.com"
+	storkService       = "stork-service"
+	storkNamespace     = "kube-system"
+	storkServiceAccess = "stork-service.kube-system.svc"
+)
+
+// CreateMutateWebhook create new webhookconfig for stork if not exist already
+func CreateMutateWebhook(caBundle []byte) error {
+	path := "/mutate"
+
+	webhook := hook.Webhook{
+		Name: webhookName,
+		ClientConfig: hook.WebhookClientConfig{
+			Service: &hook.ServiceReference{
+				Name:      storkService,
+				Namespace: storkNamespace,
+				Path:      &path,
+			},
+			CABundle: caBundle,
+		},
+		Rules: []hook.RuleWithOperations{
+			{
+				Operations: []hook.OperationType{hook.Create},
+				Rule: hook.Rule{
+					APIGroups:   []string{"apps", ""},
+					APIVersions: []string{"v1"},
+					Resources:   []string{"deployments", "statefulset", "pods"},
+				},
+			},
+		},
+	}
+	req := &hook.MutatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: storkAdmissionController,
+		},
+		Webhooks: []hook.Webhook{webhook},
+	}
+	_, err := k8s.Instance().CreateMutatingWebhookConfiguration(req)
+	if !k8serr.IsAlreadyExists(err) && err != nil {
+		log.Errorf("unable to create mutate webhook: %v", err)
+		return err
+	}
+	log.Debugf("Created mutating webhook configuration: %v", webhookName)
+	return nil
+}
+
+// GenerateCertificate Self Signed certificate , returns x509 cert
+// and priv key in PEM format
+func GenerateCertificate() ([]byte, []byte, error) {
+	var err error
+	pemCert := &bytes.Buffer{}
+
+	// generate private key
+	priv, err := ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+	if err != nil {
+		log.Errorf("error generating crypt keys: %v", err)
+		return nil, nil, err
+	}
+
+	// create certificate
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			CommonName: storkServiceAccess,
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		log.Fatalf("Failed to create x509 certificate: %s", err)
+		return nil, nil, err
+	}
+
+	err = pem.Encode(pemCert, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/webhookadmission/webhook.go
```diff
@@ -0,0 +1,210 @@
...
+				c.Recorder.Event(&admissionReview, v1.EventTypeWarning, "Could not get pvc details %v", err.Error())
+				isPXResource = false
+				break
+			}
+			if c.Driver.OwnsPVC(pvc) {
+				isPXResource = true
+				break
+			}
+		}
+	case "Deployment":
```
disrani-px: Also need this for pods since operators might not use deployments or statefulsets

pkg/webhookadmission/webhook.go
```diff
@@ -0,0 +1,210 @@
...
+			log.Errorf("Could not unmarshal admission review object: %v", err)
+			c.Recorder.Event(&admissionReview, v1.EventTypeWarning, "could not unmarshal ar object", err.Error())
+		}
+		// check whether pod spec use px volume claims
+		volumes := ss.Spec.Template.Spec.Volumes
+		for _, v := range volumes {
+			pvc, err := k8s.Instance().GetPersistentVolumeClaim(v.PersistentVolumeClaim.ClaimName, ss.Namespace)
+			if err != nil {
+				c.Recorder.Event(&admissionReview, v1.EventTypeWarning, "Could not get pvc details %v", err.Error())
+				isPXResource = false
```
disrani-px: Can be any driver supported by stork, shouldn't refer to PX here

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+			},
+			CABundle: caBundle,
+		},
+		Rules: []hook.RuleWithOperations{
+			{
+				Operations: []hook.OperationType{hook.Create},
+				Rule: hook.Rule{
+					APIGroups:   []string{"apps", ""},
+					APIVersions: []string{"v1"},
+					Resources:   []string{"deployments", "statefulset", "pods"},
```
disrani-px: Does this need to be singular or plural?

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+
+	err = pem.Encode(pemCert, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+	if err != nil {
+		log.Errorf("Unable to encode x509 certificate to PEM format: %v", err)
+		return nil, nil, err
+	}
+
+	b, err := x509.MarshalECPrivateKey(priv)
+	if err != nil {
+		log.Fatalf("Unable to marshal ECDSA private key: %v", err)
```
disrani-px: Log as Errorf

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		log.Fatalf("Failed to create x509 certificate: %s", err)
```
disrani-px: Log at Errorf

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+	hook "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	webhookName        = "stork.webhook.com"
+	storkService       = "stork-service"
+	storkNamespace     = "kube-system"
```
disrani-px: Stork can be started in any namespace, we should read that from an env variable.

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/portworx/sched-ops/k8s"
+	log "github.com/sirupsen/logrus"
+	hook "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	webhookName        = "stork.webhook.com"
```
disrani-px: Let's call this `webhook.stork.libopenstorage.io`

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+	hook "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	webhookName        = "webhook.stork.libopenstorage.io"
+	storkService       = "stork-service"
+	storkNamespace     = "STORK-NAMESPACE"
```
disrani-px: Call this `storkNamespaceEnv`

pkg/webhookadmission/webhook.go
```diff
@@ -0,0 +1,266 @@
...
+
+	var secretData, key, caBundle []byte
+	var err error
+	var ok bool
+	var tlsCert tls.Certificate
+	if c.started {
+		return fmt.Errorf("webhook server has already been started")
+	}
+
+	ns := os.Getenv("STORKNAMESPACE")
```
disrani-px: Use the const `storkNamespaceEnv`

pkg/webhookadmission/utils.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		log.Errorf("error generating crypt keys: %v", err)
+		return nil, nil, err
+	}
+
+	// create certificate
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			CommonName: storkServiceAccess,
```
disrani-px: Can't use `storkServiceAccess` since it has the namespace hardcoded to `kube-system`, need to use the namespace where stork is started.

pkg/webhookadmission/utils.go
ram-infrac: work in both cases, but lets keep it plural for all

pkg/webhookadmission/webhook.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// TODO: This log does not get reflected in stork pods, check other logger
+	arReq := admissionReview.Request
+	switch arReq.Kind.Kind {
+	case "StatefulSet":
+		var ss appv1.StatefulSet
+		if err = json.Unmarshal(arReq.Object.Raw, &ss); err != nil {
+			log.Errorf("Could not unmarshal admission review object: %v", err)
+			c.Recorder.Event(&admissionReview, v1.EventTypeWarning, "could not unmarshal ar object", err.Error())
```
disrani-px: Don't need to do as part of this PR, but events to the admission review won't really be visible anywhere. We should either add it to the deployment/sts/pod or if we don't have that info then add it to the webhook config object

pkg/webhookadmission/webhook.go
```diff
@@ -0,0 +1,268 @@
...
+			return fmt.Errorf("invalid secret certificate")
+		}
+
+		tlsCert, err = GetTLSCertificate(caBundle, secretData)
+		if err != nil {
+			log.Errorf("unable to generate tls certs: %v", err)
+			return err
+		}
+	}
+	c.server = &http.Server{Addr: ":443",
```
disrani-px: In a separate PR, make the port configurable

specs/stork-webhook.yaml
```diff
@@ -0,0 +1,18 @@
...
+      service:
+        name: stork-service
+        namespace: kube-system
+        path: "/mutate"
+      caBundle: ""
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: ["apps", ""]
+        apiVersions: ["v1"]
+        resources: ["deployments","statefulset"]
```
disrani-px: Need `pods` here too

pkg/webhookadmission/webhook.go
ram-infrac: will do

PR #543, opened 2020-01-30
Add support for gcp multi zone disks

disrani-px:
Also, use zone from disk when creating snapshot and disk instead of the zone
where stork is running

**What type of PR is this?**
Bug

No

No

No


drivers/volume/gcp/gcp.go
```diff
@@ -293,17 +377,31 @@ func (g *gcp) GetRestoreStatus(restore *storkapi.ApplicationRestore) ([]*storkap
 		if vInfo.DriverName != driverName {
 			continue
 		}
-		disk, err := g.service.Disks.Get(g.projectID, g.zone, vInfo.RestoreVolume).Do()
-		if err != nil {
-			return nil, err
+		var status string
+		if len(vInfo.Zones) > 1 {
```
ram-infrac: need zone info missing check

disrani-px: Fixed

PR #552, opened 2020-02-11
Update full image name from test script

disrani-px:
**What type of PR is this?**
integration-test

No

No

2.3


disrani-px:
Fixed with #555

PR #580, opened 2020-03-07
Set all configmap after default config map is correctly set

Rohit-PX:

**What type of PR is this?**
integration-test

Set allconfig currently gets set to blank

no

no

2.3

### DIFF
--- a/test/integration_test/applicationbackup_test.go
+++ b/test/integration_test/applicationbackup_test.go
@@ -47,11 +47,6 @@ func TestApplicationBackup(t *testing.T) {
 
 	allConfigMap = configMap.Data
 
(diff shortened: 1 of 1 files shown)


test/integration_test/applicationbackup_test.go
```diff
@@ -60,6 +55,11 @@ func TestApplicationBackup(t *testing.T) {
 	logrus.Infof("Default backup location set to %v", defaultBackupLocation)
 	defaultConfigMap = getBackupConfigMapForType(allConfigMap, defaultBackupLocation)
 
+	// If running on COLO backup to all locations
```
disrani-px: Don't mention COLO, it is for the pxd driver only

PR #582, opened 2020-03-12
Add cache to for schedule policy

disrani-px:
**What type of PR is this?**
Improvement

```release-note
Added a cache for SchedulePolicy object to reduce calls to kube-apiserver
```

2.3


disrani-px:
Code coverage has dropped because the cache is not supported with the fake client.

PR #586, opened 2020-03-17
For azure backups add env variables from px-azure secret

Rohit-PX:

**What type of PR is this?**
integration-test

In case of azure backups, stork needs env vars as used in px-azure secret in PX

no

no

2.3

### DIFF
--- a/test/integration_test/test-deploy.sh
+++ b/test/integration_test/test-deploy.sh
@@ -14,6 +14,7 @@ auth_secret_configmap=""
 volume_driver="pxd"
 short_test=false
 backup_location_path="test-restore-path"
+cloud_secret=""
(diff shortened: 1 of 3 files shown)


test/integration_test/test-deploy.sh
```diff
@@ -137,6 +137,10 @@ else
   echo "No environment variables passed."
 fi
 
+if [ "$volume_driver" == "azure" ] ; then
+    echo "For azure backups add environment variables to stork from existing k8s secret px-azure"
+    kubectl -n  kube-system set env --from=secret/px-azure deploy/stork
```
disrani-px: Should take the secret name as a param

PR #587, opened 2020-03-17
[portworx] Fix in-place restore failure for repl >2 vols

ram-infrac:
**What type of PR is this?**
>bug

Fix in-place restore failure for volume which has repl >3

no

no

yes


disrani-px:
@ram-infrac Can you squash your changes and add [Portworx] to the start of the commit message

PR #604, opened 2020-04-14
Don't start webhook controller if driver is empty

ram-infrac:
**What type of PR is this?**
>bug

Skip starting web hook controller, if storage driver is not provided

no

no

2.4.0

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -171,6 +171,7 @@ func run(c *cli.Context) {
 
 	var d volume.Driver
 	if driverName != "" {
+		log.Infof("Using driver %v", driverName)
 		d, err = volume.Get(driverName)
 		if err != nil {
 			log.Fatalf("Error getting Stork Driver %v: %v", driverName, err)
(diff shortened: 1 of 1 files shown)


cmd/stork/stork.go
```diff
@@ -190,15 +190,14 @@ func run(c *cli.Context) {
 				log.Fatalf("Error starting scheduler extender: %v", err)
 			}
 		}
+		webhook = &webhookadmission.Controller{
+			Driver:   d,
+			Recorder: recorder,
+		}
+		if err := webhook.Start(); err != nil {
+			log.Fatalf("error starting webhook controller: %v", err)
```
Rohit-PX, line 198: can add driver name in the log
```
			log.Fatalf("error starting webhook controller for driver %s: %v", driverName, err)
```

ram-infrac: we are not doing that in `extender` as well, @disrani-px  let me know if you want to add driver name to both servers

disrani-px: Instead of adding driver name to those logs, I think we can just print the driver name being used as a separate log message. ie just `Using driver <name>`

PR #610, opened 2020-04-18
Fix pre exec rule failure in application backup

disrani-px:
Multiple updates were failing because of conflicts. Fetching latest object
before updating in pre exec path
Also if the pre exec rule fails the backup will be marked as Failed instead of
retrying

**What type of PR is this?**
Bug

No

```release-note
Fixed handling of error during Pre-Exec rule in `ApplicationBackup`
```

2.4

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -266,7 +266,8 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		}
 		fallthrough
 	case stork_api.ApplicationBackupStagePreExecRule:
-		terminationChannels, err = a.runPreExecRule(backup)
+		var inProgress bool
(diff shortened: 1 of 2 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -564,21 +579,39 @@ func (a *ApplicationBackupController) runPreExecRule(backup *stork_api.Applicati
...
+		return nil, false, err
+	}
+	err = a.client.Get(context.TODO(), key, backup)
+	if err != nil {
+		for _, channel := range terminationChannels {
+			channel <- true
+		}
+		return nil, false, err
+	}
+	//	if backup.Status.Stage == stork_api.ApplicationBackupStageInitial {
```
ram-infrac, line 613: nit: remove commented code

disrani-px: Done

PR #613, opened 2020-04-22
TLS updates

taaraora:
**What type of PR is this?**
>bug
feature
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

1. TLS for Legacy REST API decided to be disabled.
2. possibility of loading CA cert from an external file was added

```release-note
Portworx driver have got env variable PX_CA_CERT_PATH, which can be used to point to CA cert which will be loaded and used for TLS enabled connections to Portworx. 
```


disrani-px:
Please run the integration tests to make sure there are no regressions. 

@Rohit-PX Can share the instructions with you

drivers/volume/portworx/portworx.go
```diff
@@ -244,21 +250,26 @@ func (p *portworx) Stop() error {
...
+		return []grpc.DialOption{grpc.WithInsecure()}, nil
+	}
+
+	var rootCA []byte
+	var err error
+	if caCertPath != "" {
+		rootCA, err = ioutil.ReadFile(caCertPath)
 		if err != nil {
-			return fmt.Errorf("failed to load CA system certs: %v", err)
+			return nil, fmt.Errorf("unable to read Root CA Cert: %v", err)
```
disrani-px, line 263: Add `caCertPath` to the message

drivers/volume/portworx/portworx.go
```diff
@@ -146,6 +145,9 @@ const (
 	pxEnableTLS    = "PX_ENABLE_TLS"
 	pxSharedSecret = "PX_SHARED_SECRET"
 
+	//path to root certificate for authenticating tls connections to porx
+	pxCaFile = "PX_CA_CERT_PATH"
```
disrani-px, line 149: Specifying the cert as a path will require a user to make multiple changes. They'll need to create a secret and then update volume mounts for the volume to mount the secret. Better to take in env variables for the name of the secret and the key in the secret which contains the cert and read the secret in the driver.

Rohit-PX:
@taaraora 
To run the integration use the following job:
https://jenkins.portworx.co/job/Stork/view/Stork%202.3%20Dashboard/job/stork-dev-test/

Need to change the following:
1. Specify the name of the stork image built from your code changes
2. Build the job

drivers/volume/portworx/portworx.go
taaraora: @disrani-px Thank you, PR was updated.

drivers/volume/portworx/portworx_test.go
```diff
@@ -0,0 +1,311 @@
+// +build unittest
+
+package portworx
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"os"
+	"testing"
+
+	"github.com/libopenstorage/stork/drivers/volume/portworx/mock"
```
disrani-px, line 10:
```suggestion
	"os"
	"testing"

	v1 "k8s.io/api/core/v1"
	"github.com/libopenstorage/stork/drivers/volume/portworx/mock"
```

drivers/volume/portworx/mock/ops.go
```diff
@@ -0,0 +1,465 @@
+// +build unittest
```
disrani-px, line 1: You don't need this, you can use the fake client. Look at the storkctl unit tests for an example

drivers/volume/portworx/portworx.go
```diff
@@ -3028,7 +2946,7 @@ func (p *portworx) createGroupLocalSnapFromPVCs(groupSnap *storkapi.GroupVolumeS
 	if err != nil {
 		return nil, err
 	}
-	resp, err := volDriver.SnapshotGroup("", nil, volNames)
+	resp, err := volDriver.SnapshotGroup("", nil, volNames, false)
```
taaraora, line 2949: @disrani-px I added this `false` to let it compile after pulling dependencies. How can I understand that it doesn't break things?

disrani-px: @talakad Is this the new flag that was added to clean up partially successful group snaps?

talakad: Correct. https://github.com/libopenstorage/openstorage/blob/master/volume/volume.go#L115

taaraora: @talakad what default value should be used for portworx driver?

talakad: This was added specifically for stork, so that stork need not follow up with delete commands for failed snaps in groupsnap request. For stork, this should be set to true.  @disrani-px ,  to comment otherwise. If clients don't mind some snaps failing , then it should be set to false.

disrani-px:
@taaraora Are the integration tests passing?

taaraora:
@disrani-px I still can't make them even start.

taaraora:
@disrani-px integration tests are passing.

PR #619, opened 2020-05-07
change base docker image to redhat-ubi8

ram-infrac:
**What type of PR is this?**
> enhancement

Change docker base image for stork from alpine to UBI-8

no

```release-note

```

yes, 2.4.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,23 @@
-FROM alpine:latest
+FROM registry.access.redhat.com/ubi8-minimal:latest
 MAINTAINER Portworx Inc. <support@portworx.com>
 
-RUN apk add tar python3 ca-certificates && apk upgrade
+ARG VERSION=master
+ARG RELEASE=latest
+LABEL name="Stork" \
+       vendor="Openstorage.org" \
+       version=${VERSION} \
+       release=${RELEASE} \
+       summary="Storage Operator Runtime for Kubernetes" \
+       description="Stork is a Cloud Native storage operator runtime scheduler plugin"
+
+RUN microdnf clean all && microdnf install -y yum
+
+RUN yum install -y wget python3 ca-certificates tar gzip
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,7 +1,14 @@
-FROM alpine:latest
+FROM registry.access.redhat.com/ubi8/ubi
 MAINTAINER Portworx Inc. <support@portworx.com>
 
-RUN apk add tar python3 ca-certificates && apk upgrade
+LABEL name="Portworx operator" \
```
piyush-nimbalkar, line 4: *Stork
update the label values based on stork

Dockerfile
```diff
@@ -1,11 +1,23 @@
-FROM alpine:latest
+FROM registry.access.redhat.com/ubi8-minimal:latest
 MAINTAINER Portworx Inc. <support@portworx.com>
 
-RUN apk add tar python3 ca-certificates && apk upgrade
+ARG VERSION=master
```
disrani-px, line 4: Won't this overwrite whatever is passed in from the Makefile?

ram-infrac: if VERSION is not passed via build-arg use `master`

Dockerfile
```diff
@@ -1,11 +1,23 @@
...
+
+RUN microdnf clean all && microdnf install -y yum
+
+RUN yum install -y wget python3 ca-certificates tar gzip
 
 RUN wget -O /usr/local/bin/aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.10.3/2018-07-26/bin/linux/amd64/aws-iam-authenticator && \
     chmod +x /usr/local/bin/aws-iam-authenticator
-RUN pip3 install --upgrade pip && pip3 install awscli && pip3 install rsa --upgrade
+
+RUN python3 -m pip install awscli && python3 -m pip install rsa
```
disrani-px, line 20: Can you also add the `--upgrade` flag. It'll upgrade the packages to the latest one if already installed.

PR #620, opened 2020-05-08
specs: remove initializers field from specs

chrboe:
**What type of PR is this?**
bug

The metadata.initializers field has been deprecated since Kubernetes
1.13, and has been completely removed in 1.16. So, when using a
Kubernetes version >=1.16, this errors out while creating the
deployment.

Fix this by just removing the key in stork-deployment.yaml and
stork-scheduler.yaml.

no

no

no


disrani-px:
@chrboe Can you rebase your branch so that this can be merged?

PR #629, opened 2020-05-22
fix stork-scheduler RBAC role for Kubernetes 1.18

chrboe:
**What type of PR is this?**
>bug

While trying to deploy stork on a k8s cluster with version 1.18, we
noticed that some permission errors came up when deploying the scheduler.
This fixes the yaml to work for k8s 1.18.

no

```release-note

```

no


chrboe:
Ping – can this be merged?

PR #630, opened 2020-05-27
Create clusterdomainstatus CR at controller init

ram-infrac:
**What type of PR is this?**
>bug

ClusterDomain CR needs to be created be default

no

```
Fixed issue with `ClusterDomainStatus` CR not being created initially
```
2.4.1


Rohit-PX:
Running with the fix here:  https://jenkins.portworx.co/job/Stork/view/Stork%20master%20Dashboard/job/stork-master-px-2.5.2-clusterdomain/12/console

PR #635, opened 2020-06-04
STOR-203 Add IO app to elasticsearch specs

stgleb:


test/integration_test/specs/elasticsearch/elasticsearch.yaml
```diff
@@ -96,4 +109,34 @@ spec:
...
+          image: portworx/torpedo-esload:1.1
+          env:
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          command:
+            - /bin/bash
+            - -c
+            - "./esload.sh --es_address elasticsearch-data-0.elasticsearch-api.$POD_NAMESPACE.svc.cluster.local:9200 --indices 4 --documents 5 --seconds 120 --not-green --clients 1"
```
disrani-px, line 140: The amount of data being written needs to be configurable via env variables passed to the test pod

PR #636, opened 2020-06-04
STOR-219 Parametrize app to run

stgleb:


test/integration_test/applicationbackup_test.go
```diff
@@ -506,7 +540,7 @@ func deletePolicyAndApplicationBackupSchedule(t *testing.T, namespace string, po
 func intervalApplicationBackupScheduleTest(t *testing.T) {
 	backupLocation := "backuplocation"
 
-	ctx := createApp(t, "interval-appbackup-sched-test")
+	ctx := createApp(t, "interval-appbackup-sched-test", "")
```
Rohit-PX, line 543: you can add test key here

test/integration_test/snapshot_test.go
```diff
@@ -363,7 +363,7 @@ func deletePolicyAndSnapshotSchedule(t *testing.T, namespace string, policyName
 }
 
 func intervalSnapshotScheduleTest(t *testing.T) {
-	ctx := createApp(t, "interval-snap-sched-test")
+	ctx := createApp(t, "interval-snap-sched-test", testKey)
```
disrani-px, line 366: Move `testKey` to common_test.go and rename it `'defaultTestKey"

test/integration_test/specs/sysbench-backup/applicationbackup.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: stork.libopenstorage.org/v1alpha1
+kind: ApplicationBackup
+metadata:
+  name: sysbench-backup
```
disrani-px, line 4: Why would you want to backup the sysbench app? You should run sysbench before the backup, do a backup+restore and then verify that the data written by sysbench matches

test/integration_test/applicationbackup_test.go
stgleb: done

test/integration_test/specs/sysbench-backup/applicationbackup.yaml
stgleb: sysbench app actually contains two containers, one is mysql and the second is mysql where sysbench generates load.

test/integration_test/snapshot_test.go
stgleb: Done

test/integration_test/specs/sysbench-backup/applicationbackup.yaml
stgleb: got it, now sysbench is removed. this PR will just add parametrization for running app.

test/integration_test/applicationbackup_test.go
```diff
@@ -360,11 +362,28 @@ func getBackupFromListWithAnnotations(backupList *storkv1.ApplicationBackupList,
 }
 
 func applicationBackupRestoreTest(t *testing.T) {
+	var (
```
Rohit-PX, line 365: We can move this to common.go so that other tests like snapshot can use these variables also, in case we decide to do IO in those tests

stgleb: Well, yes I can. I just have a concern if these test run concurrently for data races.

stgleb:
> Can you also add a link to the stork dev job run?

https://jenkins.portworx.co/job/Stork/job/stork-2-dev-test/74

test/integration_test/applicationbackup_test.go
Rohit-PX: These test run sequentially so we should be ok.

stgleb: done

stgleb:
https://jenkins.portworx.co/job/Stork/job/stork-2-dev-test/

PR #637, opened 2020-06-04
[Portworx] Add options to pass in incremental count frequency

disrani-px:
**What type of PR is this?**
Enhancement

Yes, adds `options` fields in `SchedulePolicy` and `ApplicationBackup` CRDs

```release-note
Users can now specify `portworx.io/cloudsnap-incremental-count` to control the number of incrementals to take for cloudsnaps. This can be given as an `option` in the `SchedulePolicy` or `ApplicationBackup`
```

2.4


pkg/schedule/schedule_test.go
```diff
@@ -423,3 +423,54 @@ func policyRetainTest(t *testing.T) {
 	require.NoError(t, err, "Error getting retain")
 	require.Equal(t, policy.Policy.Monthly.Retain, retain, "Wrong default retain for monthly policy")
 }
+
+func policyOptionsTest(t *testing.T) {
```
ggriffiths, line 427: add this in `TestSchedule`

drivers/volume/portworx/portworx.go
```diff
@@ -854,6 +859,19 @@ func (p *portworx) SnapshotCreate(
 		request.Labels[cloudBackupOwnerLabel] = "stork"
 		p.addCloudsnapInfo(request, snap)
 
+		if value, present := snap.Metadata.Annotations[incrementalCountAnnotation]; present {
+			incrementalCount, err := strconv.ParseUint(value, 10, 32)
+			if err != nil {
+				return nil, nil, fmt.Errorf("invalid cloudsnap-incremental-count specified %v: %v", p, err)
+			}
+			if incrementalCount <= 0 {
```
prashanthpx, line 867: Good to have a check if this is < 0 as it's invalid input.

disrani-px: <0 would mean to keep the default

PR #639, opened 2020-06-05
add flag to enable/disable webhook controller

ram-infrac:
**What type of PR is this?**
>bug

Allow user to enable/disable stork Webhook controller

yes, while deploying stork user can mention webhook=false/true to enable disable Webhooks controller
```
       - command:
         - /stork
         - --driver=pxd
         - --verbose
         - --leader-elect=true
         - --health-monitor-interval=120
         - --webhook=true
```

no

2.4.2

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -138,7 +138,7 @@ func main() {
 			Usage: "Start the controller to monitor PVC creation and deletions (default: true)",
 		},
 		cli.BoolTFlag{
-			Name:  "webhook",
+			Name:  "webhook-controller",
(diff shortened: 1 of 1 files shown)


cmd/stork/stork.go
```diff
@@ -137,6 +137,10 @@ func main() {
 			Name:  "pvc-watcher",
 			Usage: "Start the controller to monitor PVC creation and deletions (default: true)",
 		},
+		cli.BoolTFlag{
+			Name:  "webhook",
```
disrani-px, line 141:
```suggestion
			Name:  "webhook-controller",
```

PR #640, opened 2020-06-05
STOR-205 Add IO load to cassandra

stgleb:


test/integration_test/specs/cassandra/cassandra.yaml
```diff
@@ -98,3 +98,21 @@ spec:
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: cassandra-stress
+          image: chrisbelyea/cassandra-stress:latest
+          command:
+            - "/bin/bash"
+            - "-c"
+            - 'cassandra-stress write n=100000 cl=one -mode native cql3 -node cassandra-0.cassandra.$(POD_NAMESPACE).svc.cluster.local'
```
disrani-px, line 117: The amount of data being written needs to be controlled via env variables passed to the test.
Also, just writing data doesn't really help, need to have a verification that runs after migration/clone/restore to verify that data that was written,

PR #646, opened 2020-06-16
Updates for health monitor

disrani-px:
**What type of PR is this?**
Bug

* [Portworx] Map node status None to Online
* Add events when deleting pods from health monitor

No

```release-note
* [Portworx] Fix issue with mapping node status which could cause pods to be deleted from the HealthMonitor when not required
* Added events to the pod when deleting them from the HealthMonitor
```

2.4

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -384,8 +384,6 @@ func (p *portworx) inspectVolume(volDriver volume.VolumeDriver, volumeID string)
 
 func (p *portworx) mapNodeStatus(status api.Status) storkvolume.NodeStatus {
 	switch status {
-	case api.Status_STATUS_NONE:
-		fallthrough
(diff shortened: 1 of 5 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -396,11 +394,13 @@ func (p *portworx) mapNodeStatus(status api.Status) storkvolume.NodeStatus {
...
 	case api.Status_STATUS_DECOMMISSION:
 		fallthrough
-	case api.Status_STATUS_MAINTENANCE:
-		fallthrough
 	case api.Status_STATUS_NEEDS_REBOOT:
 		return storkvolume.NodeOffline
 
+	case api.Status_STATUS_NONE:
+		fallthrough
+	case api.Status_STATUS_MAINTENANCE:
```
jvinod, line 402: A node in maintenance cannot serve IO. It shouldn't be treated as online.

PR #647, opened 2020-06-16
Cluster pair should work with auth enabled Portworx

taaraora:
**What type of PR is this?**

bug
>feature
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

```release-note

```

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -191,7 +191,6 @@ var restoreStateCallBackoff = wait.Backoff{
 }
 
 type portworx struct {
-	clusterManager  cluster.Cluster
 	store           cache.Store
 	stopChannel     chan struct{}
 	sdkConn         *portworxGrpcConnection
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -243,10 +242,32 @@ func (p *portworx) getClusterDomainClient() (api.OpenStorageClusterDomainsClient
 	return api.NewOpenStorageClusterDomainsClient(conn), nil
 }
 
+func (p *portworx) getClusterManagerClient() (cluster.Cluster, error) {
```
lpabon, line 245: Add comments

drivers/volume/portworx/portworx.go
```diff
@@ -297,6 +310,46 @@ func (p *portworx) initPortworxClients() error {
 	return err
 }
 
+func (p *portworx) tokenGenerator() (string, error) {
```
lpabon, line 313: Add a comment here how it is expected that this code will be executed for those that are new to this model.

drivers/volume/portworx/portworx.go
```diff
@@ -469,7 +528,12 @@ func (p *portworx) GetNodes() ([]*storkvolume.NodeInfo, error) {
 		}
 	}
 
-	cluster, err := p.clusterManager.Enumerate()
+	clusterManager, err:= p.getClusterManagerClient()
```
lpabon, line 531: Why get cluster manager again?

Why was it removed from the object `p`?

taaraora: When cluster manager had static config we could just keep it as part of `p *portworx` struct. Now when we need to populate cluster manager client with token which has relatively short lifetime (1 hour), `getClusterManagerClient()` is used as getter which returns fully configured client with fresh token injected.

drivers/volume/portworx/portworx.go
taaraora: done

drivers/volume/portworx/portworx.go
taaraora: done

PR #648, opened 2020-06-17
STOR-200 Add prepare and verify pods for mysql

stgleb:


io-apps/sysbench/sysbench-prepare.dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM severalnines/sysbench
```
disrani-px, line 1: Can we have just one container with both the scripts and call either verify.sh or prepare.sh from the pod?

disrani-px: We don't even need a separate container IMO. We should just use the sysbench container and put the command in the pod spec. That way we don't need to maintain this extra container

stgleb: Done, I've used only vanila sysbench and mysql client images.

stgleb:
Job https://jenkins.portworx.co/job/Stork/job/stork-2-dev-test/84

PR #653, opened 2020-06-26
Add applicationregistration custom resource

ram-infrac:
**What type of PR is this?**
> feature

Add applicationregistration custom resource to register CRD for migration/backup/restore/clone opertions.

no, will add storkctl change in subsequent PR

yes
```release-note
Register crd resources using application registration CR, which will allow stork to migrate/backup CR associated with it
```

yes, 2.4.2


pkg/apis/stork/v1alpha1/applicationregistration.go
```diff
@@ -0,0 +1,53 @@
...
+
+// ApplicationResource definations to collect resource and fields
+type ApplicationResource struct {
+	// ResourceKind crd kind
+	// ResourceKind string `json:"resourceKind"`
+	// CRDName to register CR on destination clusters
+	metav1.GroupVersionKind `json:",inline"`
+	// default backup/migration collects delete all resources annotation except well known,
+	// keep specific annotation
+	KeepAnnotations []string `json:"keepAnnotations"`
```
disrani-px, line 31: What is the use case for this? By default, we don't delete annotations

ram-infrac: yes we don't, fixed

PR #655, opened 2020-06-29
Added restore volume size to restore CR

prashanthpx:
*What type of PR is this?**
 - Added restore volume size for restore CR
  - For an unsupported version of porx, returning backup size as zero
  
Adds support for adding volume size in restore CR.


disrani-px:
Split the commit into 2, one for the portworx driver change and the other for adding the size in restore.
Also add [Portworx] to the driver commit message

prashanthpx:
> Split the commit into 2, one for the portworx driver change and the other for adding the size in restore.
> Also add [Portworx] to the driver commit message

Split into two commits as suggested.

PR #654, opened 2020-06-26
Support generic CRD migration, backup/restore

ram-infrac:
**What type of PR is this?**
>feature

**What this PR does / why we need it**:
Add support in stork to migration/backup/restore generic CRDs

**Does this PR change a user-facing CRD or CLI?**:
no

**Is a release note needed?**:

```release-note
support in stork to migration/backup/restore generic CRDs. Users has to register their CRD's via ApplicationRegistration CRs. 
```

**Does this change need to be cherry-picked to a release branch?**:
yes, 2.4.2

Note: depends on PR #653

PR #659, opened 2020-06-30
Allow activate/deactivation of migrated crds

ram-infrac:
**What type of PR is this?**
> enhancement

Allow users to activate/deactivate migrated crd resources

no

```release-note

```

Depends on #654


pkg/storkctl/migration.go
```diff
@@ -186,8 +185,7 @@ func newDeactivateMigrationsCommand(cmdFactory Factory, ioStreams genericcliopti
 				updateIBPObjects("IBPCA", ns, false, ioStreams)
```
disrani-px, line 185: Don't need specific handling for IBM types either, right?

PR #665, opened 2020-07-07
Skip backup of PVCs being deleted or in Pending state

disrani-px:
**What type of PR is this?**
Bug

No

```release-note
* PVCs being deleted or in Pending state will be ignored during `ApplicationBackup`
```

2.4


disrani-px:
@siva-portworx Good point. We were skipping unbound PVCs but not PVCs being deleted. Updated the PR

PR #669, opened 2020-07-13
Improve storkctl migration interface

ram-infrac:
**What type of PR is this?**
>enhancement

Adds Enhancement to storkctl migration options
 - add option to allow validating migration spec
- add watch option to storkctl stork resources
- accept storage option for clusterpair

yes
Validate : 
```
# storkctl validate migrations -m ./migration.yaml
Checking Migration Namespaces....Ok
Checking ClusterPair Status....Ok
Checking PreExec Rule....Missing
Checking PostExec Rule....Missing
All Ready!!
```

no

2.4

### DIFF
--- a/pkg/storkctl/applicationbackup.go
+++ b/pkg/storkctl/applicationbackup.go
@@ -81,7 +81,7 @@ func newCreateApplicationBackupCommand(cmdFactory Factory, ioStreams genericclio
 		},
 	}
 	createApplicationBackupCommand.Flags().StringSliceVarP(&namespaceList, "namespaces", "", nil, "Comma separated list of namespaces to backup")
-	createApplicationBackupCommand.Flags().BoolVarP(&waitForCompletion, "wait", "w", false, "Wait for applicationbackup to complete")
+	createApplicationBackupCommand.Flags().BoolVarP(&waitForCompletion, "wait", "", false, "Wait for applicationbackup to complete")
 	createApplicationBackupCommand.Flags().StringVarP(&preExecRule, "preExecRule", "", "", "Rule to run before executing applicationbackup")
 	createApplicationBackupCommand.Flags().StringVarP(&postExecRule, "postExecRule", "", "", "Rule to run after executing applicationbackup")
 	createApplicationBackupCommand.Flags().StringVarP(&backupLocation, "backupLocation", "b", "", "BackupLocation to use for the backup")
@@ -130,6 +130,13 @@ func newGetApplicationBackupCommand(cmdFactory Factory, ioStreams genericcliopti
 
 			if len(applicationBackups.Items) == 0 {
 				handleEmptyList(ioStreams.Out)
+				return
+			}
+			if cmdFactory.IsWatchSet() {
+				if err := printObjectsWithWatch(c, applicationBackups, cmdFactory, applicationBackupColumns, applicationBackupPrinter, ioStreams.Out); err != nil {
+					util.CheckErr(err)
+					return
+				}
 				return
 			}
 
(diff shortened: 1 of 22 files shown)


pkg/storkctl/applicationbackup.go
```diff
@@ -132,6 +132,16 @@ func newGetApplicationBackupCommand(cmdFactory Factory, ioStreams genericcliopti
 				handleEmptyList(ioStreams.Out)
 				return
 			}
+			if cmdFactory.IsWatchSet() {
+				if len(applicationBackups.Items) > 1 {
+					util.CheckErr(fmt.Errorf("watch supported for per resource"))
```
disrani-px, line 137: Why is this only supported for one resource?

pkg/storkctl/clusterpair.go
```diff
@@ -249,5 +276,8 @@ func newGenerateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptio
 		},
 	}
 
+	generateClusterPairCommand.Flags().StringVarP(&storageIP, "storage-ip", "", "", "IP of one of storage node")
```
disrani-px, line 279: These are Portwrox specific, and shouldn't be explicit here. I think we should allow this to be passed in as
```
--options=ip=10.0.0.0,port=9001,token=abcde
```
In the future if drivers require more options we wouldn't need to update storkctl

pkg/storkctl/storkctl.go
```diff
@@ -37,6 +37,7 @@ func NewCommand(cmdFactory Factory, in io.Reader, out io.Writer, errOut io.Write
 		newSuspendCommand(cmdFactory, ioStreams),
 		newResumeCommand(cmdFactory, ioStreams),
 		newVersionCommand(cmdFactory, ioStreams),
+		newValidateCommand(cmdFactory, ioStreams),
```
disrani-px, line 40: Don't need this, right? Since `--validate` will be used instead

pkg/storkctl/watch.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/kubectl/pkg/util/interrupt"
+)
+
+const (
+	// CursorUpLine escape sequence
+	CursorUpLine = "\x1b[1A"
```
disrani-px, line 19: Doesn't need to be exposed from the package.

pkg/storkctl/watch.go
```diff
@@ -0,0 +1,99 @@
...
+	intr := interrupt.New(nil, cancel)
+	return intr.Run(func() error {
+		_, err := watchtools.UntilWithoutRetry(ctx, watchObject, func(e watch.Event) (bool, error) {
+			object, err := convertObjects(e.Object)
+			if err != nil {
+				return false, err
+			}
+			// format screen
+			// TODO: check escape sequence support using ncurses
+			fmt.Print(CursorUpLine)
```
disrani-px, line 43: Is this how kubectl does it too?

pkg/storkctl/watch.go
```diff
@@ -0,0 +1,35 @@
...
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/kubectl/pkg/util/interrupt"
+)
+
+// PrintObjectsWithWatch prints  by listening to even updates
+func PrintObjectsWithWatch(cmd *cobra.Command, object runtime.Object, cmdFactory Factory, columns []string, printerFunc interface{}, out io.Writer) error {
```
disrani-px, line 16:
```suggestion
func printObjectsWithWatch(cmd *cobra.Command, object runtime.Object, cmdFactory Factory, columns []string, printerFunc interface{}, out io.Writer) error {
```

PR #673, opened 2020-07-17
[Portworx]: Fix updating of volume info for backup size on unsupported p…

prashanthpx:
**What type of PR is this?**
Bugfix

On porx version < 2.6, we don't support fetching volume size. Fixed an issue wherein it used to bail out for the first instance of volume when it errors out. Now for unsupported API error, we continue to scan through all available volumes.

No

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2831,16 +2831,19 @@ func (p *portworx) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*stork
 			resp, err := cloudBackupClient.Size(ctx, req)
 			st, _ := status.FromError(err)
 			if err != nil {
+				// On error explicitly make the value to zero
+				vInfo.Size = 0
 				if st.Code() == codes.Unimplemented {
-					// if using old porx version, return 0 size
-					vInfo.Size = 0
-					volumeInfos = append(volumeInfos, vInfo)
-					return volumeInfos, nil
+					// if using old porx version which doesn't support this API, log an
+					// warning and continue to next vol if available
+					log.ApplicationBackupLog(backup).Warnf("Unsupported porx version to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
+				} else {
+					log.ApplicationBackupLog(backup).Errorf("Failed to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
+					return nil, err
 				}
-				logrus.Errorf("Failed to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
-				return nil, err
+			} else {
+				vInfo.Size = resp.GetSize()
 			}
-			vInfo.Size = resp.GetSize()
 		}
 		volumeInfos = append(volumeInfos, vInfo)
 	}


drivers/volume/portworx/portworx.go
```diff
@@ -2834,13 +2834,13 @@ func (p *portworx) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*stork
 				if st.Code() == codes.Unimplemented {
 					// if using old porx version, return 0 size
 					vInfo.Size = 0
-					volumeInfos = append(volumeInfos, vInfo)
-					return volumeInfos, nil
+					logrus.Warnf("Unsupported porx version to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
+				} else {
+					logrus.Errorf("Failed to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
```
siva-portworx, line 2839: If the PORX supports size API but still if we got some error, what should be the behaviour?
Do we need to return error immediately from  GetBackupStatus  or just print the error message and make the size of that volume to be zero?

prashanthpx: My understanding is to continue as there could be a case where each volume can be on a different provider which might not fail.

siva-portworx: 1) If we continue, we might end up in giving wrong size information for those volumes for which size API failed. It might be confusing to the user that backup was success but size is zero.
2) Also, Do we need to make vInfo.Size to zero? Will it not default be zero. If not, we might need to do that in the error case as well.

prashanthpx: Initially, I had it to return an error and then changed it.  I will rever it to return an error that will fail the backup.

drivers/volume/portworx/portworx.go
```diff
@@ -2831,16 +2831,19 @@ func (p *portworx) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*stork
...
+				// On error explicitly make the value to zero
+				vInfo.Size = 0
 				if st.Code() == codes.Unimplemented {
-					// if using old porx version, return 0 size
-					vInfo.Size = 0
-					volumeInfos = append(volumeInfos, vInfo)
-					return volumeInfos, nil
+					// if using old porx version which doesn't support this API, log an
+					// warning and continue to next vol if available
+					logrus.Warnf("Unsupported porx version to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
```
disrani-px, line 2839: Use `log.ApplicationBackupLog(backup)`

drivers/volume/portworx/portworx.go
```diff
@@ -2831,16 +2831,19 @@ func (p *portworx) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*stork
...
+				} else {
+					logrus.Errorf("Failed to fetch backup size for backup ID %v: %v", csStatus.cloudSnapID, err)
```
disrani-px, line 2841: Use `log.ApplicationBackupLog(backup)`

drivers/volume/portworx/portworx.go
prashanthpx: Fixed.

drivers/volume/portworx/portworx.go
prashanthpx: Fixed.

PR #681, opened 2020-07-23
Add validation of backup deletion from cloud object stores

Rohit-PX:

**What type of PR is this?**
integration-test

Add validation to check if application backups have been deleted from the cloud (S3/GCP/Azure) 
after being deleted locally.

no

no

no


test/integration_test/applicationbackup_test.go
```diff
@@ -1266,3 +1306,26 @@ func getSyncedBackupWithAnnotation(appBackup *storkv1.ApplicationBackup, lookUpA
...
+
+func getObjectStoreDriverForBackupLocation(backupLocation *storkv1.BackupLocation) (objectstore.Driver, error) {
+	var objectStoreDriverName string
+	var ok bool
+	backupType, err := getBackupLocationType(backupLocation.Location.Type)
+	if err != nil {
+		return nil, err
+	}
+	backupTypeDriverName := map[string]string{
+		"s3":     "aws",
```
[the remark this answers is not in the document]
Rohit-PX: This code block is not required since, we are using only one objectstore driver.

Rohit-PX:
Testing here:
https://jenkins.portworx.co/job/Stork/view/Stork%20master%20Dashboard/job/stork-master-px-2.5-dev/465/

test/integration_test/common_test.go
```diff
@@ -51,17 +52,18 @@ import (
...
+	remotePairName               = "remoteclusterpair"
+	remoteConfig                 = "remoteconfigmap"
+	specDir                      = "./specs"
+	defaultClusterPairDir        = "cluster-pair"
+	pairFileName                 = "cluster-pair.yaml"
+	remoteFilePath               = "/tmp/kubeconfig"
+	configMapSyncWaitTime        = 3 * time.Second
+	defaultSchedulerName         = "default-scheduler"
+	bucketPrefix                 = "stork-test"
+	torpedoObjectStoreDriverName = "torpedo_objectstore"
```
disrani-px, line 66: Can you call this the default driver instead of torpedo driver?

Rohit-PX:
> Looks like there are other changes not related to the deletion verification in this PR too. Can you put those in a separate commit and mention why they are required, it isn't clean by just looking at the diff

Those changes are required because we are using the `backup.Status.BackupPath` for validation, due to inconsistencies in the specs, the backup path was not consistent in some cases so these changes were introduced.

This way we ensure that we always pass the correct `backup.Status.BackupPath` every time for validation

test/integration_test/applicationbackup_test.go
```diff
@@ -63,6 +64,10 @@ func TestApplicationBackup(t *testing.T) {
 }
 
 func TestScaleApplicationBackup(t *testing.T) {
+
+	if !defaultsBackupSet {
```
Rohit-PX, line 68: This is to ensure that even if `TestScaleApplicationBackup` is run independently we set all the required defaults.
I noticed this during my testing when making these changes.

test/integration_test/applicationbackup_test.go
```diff
@@ -266,11 +289,18 @@ func triggerScaleBackupRestoreTest(
 	}
 	// Cleanup
 	for idx, bkp := range appBkps {
-		err := storkops.Instance().DeleteBackupLocation("backuplocation-"+strconv.Itoa(idx), bkp.Namespace)
-		require.NoError(t, err, "Failed to delete  backup location %s in namespace %s: %v", "backuplocation-"+strconv.Itoa(idx), bkp.Namespace, err)
+		// Get backuplocation so it can be used in the invocation of the validateBackupDeletionFromObjectstore method
```
Rohit-PX, line 292: This is required because `backup.Status.BackuPath` is populated only after backup is completed.

test/integration_test/common_test.go
Rohit-PX: Fixed.

PR #682, opened 2020-07-23
Create new bucket for stork integration tests

Rohit-PX:

**What type of PR is this?**
integration-test

Need to create new buckets in the cloud for every integration test. This is required because if we have a single bucket
for all application backups done as part of the test, on syncing from that bucket we get numerous backups on the synced cluster. We need to avoid the above scenario.

no

no

no


Rohit-PX:
Testing here: https://jenkins.portworx.co/job/Stork/view/Stork%20master%20Dashboard/job/stork-master-px-2.6-dev/233/

PR #693, opened 2020-08-07
Add support to pass in * to select all namespaces to backup and update rule to take container name

disrani-px:
Add support to pass in * to select all namespaces to backup

Update Rule spec to take in container name
This is required since a pod could have multiple containers and we
need to know which one we should run the commands in

Adds container field to Rule spec

```release-note
* Added support to specify * as a wildcard to select all namespaces in ApplicationBackup spec
* Added container name in Rule spec. This is required if you have multiple containers in a pod and want to select a container to run the command in.

```

No

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -51,6 +51,8 @@ const (
 	backupCancelBackoffInitialDelay = 5 * time.Second
 	backupCancelBackoffFactor       = 1
 	backupCancelBackoffSteps        = math.MaxInt32
+
+	allNamespacesSpecifier = "*"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -156,13 +158,34 @@ func (a *ApplicationBackupController) performRuleRecovery() error {
...
 		backup.Status.TriggerTimestamp = backup.CreationTimestamp
+		updated = true
+	}
+	return updated
+}
+
+func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.ApplicationBackup) error {
+	namespaces, err := core.Instance().ListNamespaces(nil)
+	if err != nil {
+		return fmt.Errorf("error updating with all namespaces for wildcard: %v", err)
```
prashanthpx, line 177: nit:  as this is listing, should `updating` be changed to `listing`

pkg/rule/rule.go
```diff
@@ -248,11 +249,13 @@ func ExecuteRule(
...
 		backgroundCommandTermChan := make(chan bool, 1)
 		backgroundPodListChan := make(chan v1.Pod)
-		go cmdTerminationWatcher(backgroundPodListChan, backgroundCommandTermChan, owner, taskID.String())
+		var container string
+		go cmdTerminationWatcher(backgroundPodListChan, &container, backgroundCommandTermChan, owner, taskID.String())
 
 		// backgroundActionPresent is used to track if there is atleast one background action
 		backgroundActionPresent := false
 		for _, item := range rule.Rules {
+			container = item.Container
```
prashanthpx, line 258: To understand, if the user doesn't provide container input in the Rule spec, would it fall back to running on all containers?

pkg/applicationmanager/controllers/applicationbackup.go
disrani-px: The function is trying to update the backup spec with all the namespaces

pkg/rule/rule.go
disrani-px: No, it'll throw an error, which is the current behavior. We can't run the command in all the containers

PR #696, opened 2020-08-10
Added check for Notfound error in GetRestoreStatus of all driver.

siva-portworx:
**What type of PR is this?**
bug (STOR-244)
Added check for Notfound error in GetRestoreStatus of all driver.
        - In all driver module except portworx, add check in get volume
          status during restore to mark the restore of the volume, if we
          NotFound error, while querying volume status.

no
no
2.4

**Running integration test to make sure nothing is breaking.**
https://jenkins.portworx.co/job/Stork/job/stork-2-dev-test/93/consoleFull


drivers/volume/gcp/gcp.go
```diff
@@ -506,6 +515,13 @@ func (g *gcp) GetRestoreStatus(restore *storkapi.ApplicationRestore) ([]*storkap
 		} else {
 			disk, err := g.service.Disks.Get(g.projectID, vInfo.Zones[0], vInfo.RestoreVolume).Do()
 			if err != nil {
+				if googleErr, ok := err.(*googleapi.Error); ok {
```
ram-infrac, line 518: You will need to ignore error in `GetRestoreStatus` call in application restore.go since we are processing volume information afterward

ram-infrac: ok so we are continue working in case of error

drivers/volume/aws/aws.go
```diff
@@ -523,6 +523,14 @@ func (a *aws) GetRestoreStatus(restore *storkapi.ApplicationRestore) ([]*storkap
 		}
 		ebsVolume, err := a.getEBSVolume(vInfo.RestoreVolume, nil)
 		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
+				if awsErr.Code() == "InvalidVolume.NotFound" {
+					vInfo.Status = storkapi.ApplicationRestoreStatusFailed
+					vInfo.Reason = "Restore failed for volume: NotFound"
+					continue
```
disrani-px, line 530: You can't continue, this will not return info for this volume, need to append vInfo to volumeInfos

siva-portworx: Fixed. Appending vInfo to VolumeInfos, before continue.

drivers/volume/gcp/gcp.go
siva-portworx: With addressing @disrani-px comment, now it is handled properly.

PR #697, opened 2020-08-10
CSI backup and restore driver

ggriffiths:
**What type of PR is this?**
feature

Adds CSI driver for stork backup and restore.

**What this PR doesn't do**:
* Delete/cancel backup needs more work to correctly assign the DeletionPolicy and perform a delete.
* Need to somehow add a check that a given CSI driver is capable of doing backup/restore across clusters. Maybe do a pre-flight backup/restore check to validate a CSI Driver is capable of doing the work. This includes checking CSI snapshot capabilities and doing a small test backup/restore

No

```release-note
Added the ability to backup and restore CSI volumes
```

2.6


drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1023 @@
...
+	if err != nil {
+		return err
+	}
+	logrus.Debugf("created vsc for backup %s deletion", snapshotID)
+
+	return nil
+}
+
+func (c *csi) CancelBackup(backup *storkapi.ApplicationBackup) error {
+	return nil
```
disrani-px, line 527: This should delete the VolumeSnapshot objects

drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1023 @@
...
+	storkCSIDriverName = "csi"
+	// snapshotPrefix is appended to CSI backup snapshot
+	snapshotPrefix = "snapshot-"
+	// snapshotClassNamePrefix is the prefix for snapshot classes per CSI driver
+	snapshotClassNamePrefix = "stork-csi-snapshot-class-"
+
+	// snapshotObjectName is the object stored for the volumesnapshot
+	snapshotObjectName = "snapshots.json"
+	// snapshotObjectName is the object stored for the volumesnapshot
+	storageClassesObjectName = "storageclasses.json"
```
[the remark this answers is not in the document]
ggriffiths: The SC is required for restoring. Since we're creating a new PVC and PV, we need to reference the original SC. The other option is to specificy a new SC upon restore. Is that preferred? 

Is there a way to specific a snapshotclass in `ApplicationBackupSpec`, or should I add as an Option?

pkg/resourcecollector/resourcecollector.go
```diff
@@ -444,11 +446,57 @@ func (r *ResourceCollector) prepareResourcesForCollection(
...
+				return isCSIPersistentVolume(&pv)
+			}
+		}
+	}
+
+	return false, nil
+}
+
+func isCSIPersistentVolume(pv *v1.PersistentVolume) (bool, error) {
+	if pv.Spec.CSI != nil {
```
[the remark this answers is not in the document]
ggriffiths: `GetPVDriver` won't work for this case. In this scenario, we're checking if it's okay to restore the PV and PVC. 

We need to backup the PV/PVC yaml to apply labels/etc upon restore. So, this is intended to skip directly restoring the PV/PVC objects. 

During restore, the PV does not exist yet, so csi.OwnsPVC will not work. This is because I need to get the PV corresponding to a given PVC inside of csi.OwnsPVC.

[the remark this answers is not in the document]
ggriffiths: Reworking this to call GetPVDriver for both PV and PVCs. For PVCs, I'll need to find the PV associated with it. 

Had to refactor this around today, as `includeObject` is now called both on backup and restore side.

drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1023 @@
...
+		DeletionPolicy: kSnapshotv1beta1.VolumeSnapshotContentRetain,
+	})
+}
+
+// Construct the full base path for a given backup
+// The format is "namespace/backupName/backupUID" which will be unique for each backup
+func (c *csi) getObjectPath(
+	backup *storkapi.ApplicationBackup,
+) string {
+	return filepath.Join(backup.Namespace, backup.Name, string(backup.UID))
```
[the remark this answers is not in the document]
ggriffiths: exposed getObjectPath in `pkg/applicationmanager/controllers` and used it here instead.

ggriffiths:
Ready for re-review @disrani-px @ram-infrac

pkg/log/log.go
```diff
@@ -205,6 +205,18 @@ func PVCLog(pvc *v1.PersistentVolumeClaim) *logrus.Entry {
 	return logrus.WithFields(logrus.Fields{})
 }
 
+// PVLog formats a log message with pv information
+func PVLog(pv *v1.PersistentVolume) *logrus.Entry {
+	if pv != nil {
```
disrani-px, line 210: Can you add a simple UT in log_test.go

pkg/resourcecollector/resourcecollector.go
```diff
@@ -454,6 +455,45 @@ func (r *ResourceCollector) prepareResourcesForCollection(
...
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find PV %s for PVC %s", pvc.Spec.VolumeName, pvc.Name)
+}
+
+// PrepareResourceForApply prepares the resource for apply including update
+// namespace and any PV name updates. Should be called before DeleteResources
+// and ApplyResource
```
disrani-px, line 496: Comment needs to be updated

drivers/volume/csi/csi.go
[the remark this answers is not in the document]
disrani-px: It can be a key/value under `Option`

drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1214 @@
...
+	return c.OwnsPV(pv)
+}
+
+func (c *csi) HasNativeVolumeDriverSupport(pv *v1.PersistentVolume) bool {
+	return pv.Spec.CSI.Driver == snapv1.PortworxCsiProvisionerName ||
+		pv.Spec.CSI.Driver == snapv1.PortworxCsiDeprecatedProvisionerName ||
+		pv.Spec.CSI.Driver == "pd.csi.storage.gke.io" ||
+		pv.Spec.CSI.Driver == "ebs.csi.aws.com" ||
+		pv.Spec.CSI.Driver == "disk.csi.azure.com" ||
+		pv.Spec.CSI.Driver == "linstor.csi.linbit.com"
```
disrani-px, line 160: We don't support linstor natively for backup, so you can remove this

drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1214 @@
...
+	pv, err := core.Instance().GetPersistentVolume(pvc.Spec.VolumeName)
+	if err != nil {
+		log.PVCLog(pvc).Warnf("error getting pv %v for pvc %v: %v", pvc.Spec.VolumeName, pvc.Name, err)
+		return false
+	}
+	return c.OwnsPV(pv)
+}
+
+func (c *csi) HasNativeVolumeDriverSupport(pv *v1.PersistentVolume) bool {
+	return pv.Spec.CSI.Driver == snapv1.PortworxCsiProvisionerName ||
```
disrani-px, line 155: Instead of having a hardcoded list here, it would be best for each driver to register their csi driver name to the volume driver. Then this can just make a call to get all the supported drivers and look at the list.

Can be done in a separate PR

drivers/volume/csi/csi.go
```diff
@@ -0,0 +1,1214 @@
...
+	// resourcesObjectName is the object stored for the all backup resources
+	resourcesObjectName = "resources.json"
+
+	// optCSIDriverName is an option for storing which CSI Driver a volumesnapshot was created with
+	optCSIDriverName = "csi-driver-name"
+	// optCSISnapshotClassName is an option for providing a snapshot class name
+	optCSISnapshotClassName = "csi-snapshot-class-name"
+
+	// ann is an option for providing a snapshot class name
+	annSnapshotClassStorkOwned = "csi-snapshot-class-stork-owned"
```
disrani-px, line 59: Use `stork.libopenstorage.org/` as prefix for annotations and labels added by stork

ie `stork.libopenstorage.org/csi-snapshot-class-owned: true`

disrani-px (approved):
Few minor comments, looks good otherwise

PR #699, opened 2020-08-13
Passing empty endpoint to AWS objectstore operations

prashanthpx:
AWS SDK fetches correct endpoint based on provided region

**What type of PR is this?**
Enhancement

AWS SDK supports fetching of correct endpoints based on region provided if the user passes an empty endpoint.
This makes the user experience better wherein they need to remember different endpoints for different regions.

No

No

No

### DIFF
--- a/pkg/objectstore/s3/s3.go
+++ b/pkg/objectstore/s3/s3.go
@@ -14,8 +14,15 @@ import (
 )
 
 func getSession(backupLocation *stork_api.BackupLocation) (*session.Session, error) {
+	// AWS SDK fetches the correct endpoint based on region provided if endpoint is passed empty
+	var endpoint string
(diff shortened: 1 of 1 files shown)


pkg/objectstore/s3/s3.go
```diff
@@ -14,8 +15,16 @@ import (
 )
 
 func getSession(backupLocation *stork_api.BackupLocation) (*session.Session, error) {
+	// AWS SDK fetches the correct endpoint based on region provided if endpoint is passed empty
+	// For all AWS endpoint, we can pass in empty
+	var endpoint string
+	if strings.Contains(backupLocation.Location.S3Config.Endpoint, "amazonaws.com") {
```
disrani-px, line 21: Only do this if the endpoint is exactly "s3.amazonaws.com"

prashanthpx: Fixed.

PR #701, opened 2020-08-16
Fix to populate backup and restore size for azure volumes

prashanthpx:
**What type of PR is this?**
Bug fix

Backup/Restore size for azure volumes are not shown as currently, it was populating for in-progress backups. Moved the same to fetch once the backup is completed.

No

No

No


disrani-px:
Add [Azure] prefix to the commit message

PR #704, opened 2020-08-19
Allow backuplocation to be passed in for migration 

ram-infrac:
**What type of PR is this?**
>feature

Allow using backuplocation for migration operation
  - accept backuplocation as storage option for clusterpair
  - create backuplocation cr on destination cluster
  - pass backuplocation to portworx driver clusterpair api
 -  update openstorage vendor
no
no

2.5


drivers/volume/portworx/portworx.go
```diff
@@ -2125,11 +2125,16 @@ func (p *portworx) CreatePair(pair *storkapi.ClusterPair) (string, error) {
 		return "", fmt.Errorf("cannot get cluster manager, err: %s", err.Error())
 	}
 
+	var credID string
+	if bkpl, ok := pair.Spec.Options["backuplocation"]; ok {
```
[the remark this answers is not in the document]
ram-infrac: yes we can use stork_api backuplocationname used for cr registration

pkg/migration/controllers/clusterpair.go
```diff
@@ -147,18 +150,72 @@ func (c *ClusterPairController) handle(ctx context.Context, clusterPair *stork_a
...
+					err.Error())
+				return c.client.Update(context.TODO(), clusterPair)
+			}
+			resp.ResourceVersion = ""
+			if _, err := remoteClient.CreateBackupLocation(resp); err != nil && !errors.IsAlreadyExists(err) {
+				c.recorder.Event(clusterPair,
+					v1.EventTypeWarning,
+					string(clusterPair.Status.SchedulerStatus),
+					err.Error())
+				return c.client.Update(context.TODO(), clusterPair)
```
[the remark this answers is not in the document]
ram-infrac: yes, make sense . move to func. thanks

pkg/migration/controllers/clusterpair.go
```diff
@@ -220,3 +230,45 @@ func (c *ClusterPairController) createCRD() error {
...
+			if err != nil && !errors.IsAlreadyExists(err) {
+				return err
+			}
+		}
+		// create backuplocation on destination cluster
+		resp, err := storkops.Instance().GetBackupLocation(bkpl, clusterPair.GetNamespace())
+		if err != nil {
+			return err
+		}
+		resp.ResourceVersion = ""
```
disrani-px, line 268: Don't you also have to remove the UUID, self link and other members from the metadata?

pkg/migration/controllers/clusterpair.go
```diff
@@ -220,3 +230,45 @@ func (c *ClusterPairController) createCRD() error {
 
 	return apiextensions.Instance().ValidateCRD(resource, validateCRDTimeout, validateCRDInterval)
 }
+
+func createBackupLocationOnRemote(remoteConfig *restclient.Config, clusterPair *stork_api.ClusterPair) error {
```
disrani-px, line 234:
```suggestion
func (c *ClusterPairController) createBackupLocationOnRemote(remoteConfig *restclient.Config, clusterPair *stork_api.ClusterPair) error {
```

drivers/volume/portworx/portworx.go
```diff
@@ -2125,11 +2125,16 @@ func (p *portworx) CreatePair(pair *storkapi.ClusterPair) (string, error) {
 		return "", fmt.Errorf("cannot get cluster manager, err: %s", err.Error())
 	}
 
+	var credID string
+	if bkpl, ok := pair.Spec.Options[storkapi.BackupLocationResourceName]; ok {
```
disrani-px, line 2129: Add a version check for 2.6.0 here and return an error if it is an older version

pkg/migration/controllers/clusterpair.go
ram-infrac: only  setting `resourceVersion` restrict resource creation on remote cluster

PR #706, opened 2020-08-26
Add support to specify resources during application backup

disrani-px:
**What type of PR is this?**
Feature

Yes, adds `IncludeResources` in ApplicationBackup and ApplicationBackupSchedule

```release-note
Added support to specify which resource to backup using ApplicationBackup. By default all resources in the namespace will be backed up
```

2.5


ram-infrac (approved):
lgtm

PR #707, opened 2020-08-30
Submitting the volume StartBackup in a batch count of ten.

siva-portworx:
**What type of PR is this?**
enchancement

Submitting the volume StartBackup in a batch count of ten.
This way, we will update the volume backup status frequently to the CR content.

No

No

2.4?

Verified the fix on Bhavan's setup.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -53,6 +54,9 @@ const (
 	backupCancelBackoffSteps        = math.MaxInt32
 
 	allNamespacesSpecifier = "*"
+	backupVolumeBatchCount = 10
+	maxRetry               = 10
+	retrySleep             = 10 * time.Second
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -470,37 +477,39 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
-					string(stork_api.ApplicationBackupStatusFailed),
-					message)
-				backup.Status.Status = stork_api.ApplicationBackupStatusFailed
-				backup.Status.Stage = stork_api.ApplicationBackupStageFinal
-				backup.Status.Reason = message
+				backup.Status.Volumes = append(backup.Status.Volumes, volumeInfos...)
+				backup.Status.Status = stork_api.ApplicationBackupStatusInProgress
+				backup.Status.Reason = "Volume backups are in progress"
 				backup.Status.LastUpdateTimestamp = metav1.Now()
 				err = a.client.Update(context.TODO(), backup)
```
disrani-px, line 508: Since you are updating it here further updates will fail because of the version conflict. Need to read the new object before updating again

siva-portworx: Update returns the updated version of the CRD. But if the update happens outside or by some other thread, it will fail.
Added fix for the same.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -413,94 +420,97 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+			if driverName != "" {
+				if pvcMappings[driverName] == nil {
+					pvcMappings[driverName] = make([]v1.PersistentVolumeClaim, 0)
 				}
+				pvcMappings[driverName] = append(pvcMappings[driverName], pvc)
 			}
 		}
+	}
+
+	if len(backup.Status.Volumes) != len(pvcMappings) {
```
disrani-px, line 474: This will always be true since `backup.Status.Volumes` is being initialized on line 425

siva-portworx: Fixed the issue.

**Manually Tested below use case:**
1) In the code, made the batch count to one (testing purpose) and added a sleep of 5 mints before updating the CR content after calling driver.StartBackup().
2) After sleep, I was manually updating the CR content directly using "kubectl edit", To make the CR update fail from the stork code.
3) After starting first volume backup, when it was waiting (sleep), I manually updated the CR. This made the CR update fail and returned.
4) When the reconciler was called again, It started from the beginning.
5) This time, I allowed the first volume update to go through and during the wait for second volume backup, I manually updated CR again.
6) The update failed again and this time, the backup.Volume.Status was not nil and len(backup.Status.Volumes) != pvcCount. 
7) So it continued again from the beginning and also the make was called again. 
8) Then allowed the backup to continue and it completed successfully.

**Also ran the TestApplicationBackup focus integration test with the change.**
https://jenkins.portworx.dev/job/Stork/job/stork-2-dev-test/103/

disrani-px (approved):

PR #710, opened 2020-09-02
Prometheus metrics for stork extender and monitor

ram-infrac:
**What type of PR is this?**
>feature

Adds prometheus metrics for stork extender and monitor

no

yes, will be put in final pr part for stork prometheus metrics


pkg/extender/extender.go
```diff
@@ -265,6 +276,91 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
 	}
 }
 
+func (e *Extender) isNodeHyperConverged() error {
```
disrani-px, line 279:
```suggestion
func (e *Extender) startMetrics() error {
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	// HyperConvergedPodsCounter for pods hyper-converged by stork scheduler i.e scheduled on
+	// node where replicas for all pod volumes exists
+	HyperConvergedPodsCounter = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
disrani-px, line 10: These should be in the extender and health monitor packages. Just have helper methods here to register the metrics

specs/prometheus/rule.yaml
```diff
@@ -0,0 +1,24 @@
...
+    role: prometheus-stork-rulefiles
+  name: prometheus-stork-rules-stork.rules.yaml
+  namespace: kube-system
+spec:
+  groups:
+  - name: stork-rules
+    rules:
+    - alert: StorkNonHyperConvergedPod
+      annotations:
+        message: Application pod {{ $labels.exported_pod }} in namespace {{ $labels.exported_namespace
```
disrani-px, line 15: Should those be `pod` and `namespace` instead of `exported_pod` and `exported_namespace`? Because I think that is the label you are setting in the metrics

specs/prometheus/alert_manager/alert_manager.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: stork #This name is important since the Alertmanager pods wont start unless a secret named alertmanager-${ALERTMANAGER_NAME} is created. in this case if would expect alertmanager-portworx secret in the kube-system namespace
```
disrani-px, line 4:
```suggestion
  name: stork #This name is important since the Alertmanager pods won't start unless a secret named alertmanager-${ALERTMANAGER_NAME} is created. in this case, if would expect alertmanager-stork secret in the kube-system namespace
```
That comment is talking about a secret, did you mean service?

pkg/extender/extender.go
```diff
@@ -265,6 +276,91 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
...
+	fn := func(object runtime.Object) error {
```
disrani-px, line 280: Can you add unit tests for this function? You should be able to create pods and query the metrics endpoints for the stats

pkg/extender/extender.go
```diff
@@ -41,6 +45,27 @@ const (
...
+	// where pod volume replicas exists
+	NonHyperConvergePodsCounter = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "non_hyperconverged_pods_total",
+		Help: "The total number of pods that are not hyper-converge by stork scheduler",
+	}, []string{"pod", "namespace"})
+	// SemiHyperConvergePodsCounter for pods which scheduled on node where replicas for all
+	// volumes does not exists
+	SemiHyperConvergePodsCounter = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "semi_hyperconverged_pods_total",
+		Help: "The total number of pods that partially hyper-converge by stork scheduler",
```
disrani-px, line 65:
```suggestion
		Help: "The total number of pods that are partially hyper-converged by stork scheduler",
```

pkg/extender/extender.go
```diff
@@ -63,11 +88,17 @@ func (e *Extender) Start() error {
 	e.server = &http.Server{Addr: ":8099"}
 
 	http.HandleFunc("/", e.serveHTTP)
+	http.Handle("/metrics", promhttp.Handler())
```
disrani-px, line 91: I think this needs to be registered at a higher level, since the metrics will be used by other modules too. If the extender is disabled it'll end up not registering the metrics handler

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/libopenstorage/stork/pkg/extender"
+	"github.com/libopenstorage/stork/pkg/monitor"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Register stork metrics
+func init() {
+	prometheus.MustRegister(extender.HyperConvergedPodsCounter)
```
disrani-px, line 11: These should be in `pkg/extender/extender.go` since it is being populated by that package

pkg/extender/extender_test.go
```diff
@@ -952,3 +955,66 @@ func preferLocalNodeTest(t *testing.T) {
...
+	nonHyperPod := newPod("NonHyperPodTest", []string{"non-metric-vol"})
+	nonHyperPod.Spec.NodeName = "node5"
+	nonHyperPod.Status.Conditions = make([]v1.PodCondition, 1)
+	nonHyperPod.Status.Conditions[0].Type = v1.PodReady
+	nonHyperPod.Status.Conditions[0].Status = v1.ConditionTrue
+	_, err = core.Instance().CreatePod(nonHyperPod)
+	require.NoError(t, err, "failed to create pod")
+
+	time.Sleep(3 * time.Second)
+	require.Equal(t, testutil.ToFloat64(NonHyperConvergePodsCounter), float64(1), "non_hyperconverged_pods_total not matched")
```
disrani-px, line 1019: Can you add test cases to delete the pods and make sure the 3 counters go to 0.

Also, can you verify the labels on the counter

specs/prometheus/alert_manager/alert_manager.yaml
disrani-px: @ram-infrac Does this need to be updated?

specs/prometheus/alert_manager/servcie.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
disrani-px, line 1: File name has a typo

pkg/extender/extender.go
ram-infrac: should we have seperate prom metric server, which can be disable at start?

pkg/extender/extender_test.go
ram-infrac: `testutil` needs to be updated with latest release for label check & `0` in ToFloat64()

disrani-px: As discussed this will be done in a separate PR since it required some vendor updates which have conflicts

disrani-px (approved):

PR #718, opened 2020-09-11
Add security annotations to CRDs for auth-runs

Rohit-PX:

**What type of PR is this?**
integration-test

For auth-enabled runs when not using custom method to create CRDs,  we need to add security annotations.

no

no

no

### DIFF
--- a/test/integration_test/applicationbackup_test.go
+++ b/test/integration_test/applicationbackup_test.go
@@ -30,6 +30,10 @@ const (
 	s3SecretName         = "s3secret"
 	azureSecretName      = "azuresecret"
 	googleSecretName     = "googlesecret"
+	secretNameKey        = "secret_name"
+	secretNamespaceKey   = "secret_namespace"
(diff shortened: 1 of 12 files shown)


test/integration_test/applicationbackup_test.go
```diff
@@ -1243,3 +1330,97 @@ func getSyncedBackupWithAnnotation(appBackup *storkv1.ApplicationBackup, lookUpA
 	}
 	return backupToRestore, nil
 }
+
+func addSecurityAnnotation(spec interface{}) error {
```
disrani-px, line 1334: Move this to common_test.go

disrani-px (approved):

test/integration_test/applicationbackup_test.go
Rohit-PX: Done.

ggriffiths (approved):

PR #729, opened 2020-10-16
[CSI] Add restoreInfo for initial restore and during status check

ggriffiths:
**What type of PR is this?**
bug

Adds additional info to restore info object.

No

No

2.6.0


drivers/volume/csi/csi.go
```diff
@@ -1033,56 +1036,64 @@ func (c *csi) createRestoreSnapshotsAndPVCs(
 		snapshotID := vbInfo.BackupID
 		vsc, err := csiBackupObject.GetVolumeSnapshotContent(snapshotID)
 		if err != nil {
-			return err
+			return volumeRestoreInfos, err
```
disrani-px, line 1039: You can return `nil, err` in case of error

disrani-px (approved):

PR #731, opened 2020-10-26
Allow disabling/enabling cronjob object after migration

ram-infrac:
**What type of PR is this?**
>enhancement

This PR adds support for disable/enable cronjob object once migrated to remote cluster

no

yes

```release-note
Adds support for disabling cronjob object upon migration to remote cluster.  One can now activate/deactivate cronjob object as well using `storkctl activate/deactivate migration <migration_namespace>`
```

yes, 2.6.0 (or any early release before 2.6.0)

Output: 
```
# kubectl get all -ncron
NAME                  SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
cronjob.batch/hello   */1 * * * *   True      0        <none>          75s

# storkctl activate  migration -ncron
Updated suspend option for cronjob cron/hello to false

# kubectl get all -ncron
NAME                  SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
cronjob.batch/hello   */1 * * * *   False     0        <none>          27m
```

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -855,6 +855,11 @@ func (m *MigrationController) prepareResources(
 			if err != nil {
 				return fmt.Errorf("error preparing %v resource %v: %v", o.GetObjectKind().GroupVersionKind().Kind, metadata.GetName(), err)
 			}
+		case "CronJob":
+			err := m.prepareJobResource(migration, o)
+			if err != nil {
+				return fmt.Errorf("error preparing %v resource %v: %v", o.GetObjectKind().GroupVersionKind().Kind, metadata.GetName(), err)
+			}
 		}
 
 		// prepare CR resources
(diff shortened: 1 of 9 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -39,6 +39,10 @@ const (
 	// StorkMigrationReplicasAnnotation is the annotation used to keep track of
 	// the number of replicas for an application when it was migrated
 	StorkMigrationReplicasAnnotation = "stork.libopenstorage.org/migrationReplicas"
+	// StorkMigrationSuspendAnnotation is the annotation used to enable/disable
+	// k8s object which requires field to be set to true/false
+	StorkMigrationSuspendAnnotation = "stork.libopenstorage.org/migrationSuspend"
```
disrani-px, line 44: Why do we need to store that in the annotation? I think we can always suspend it on migration if `StartApplications` is false and trigger it when migrations are activated/deactivated

disrani-px (approved):

PR #733, opened 2020-10-27
Reporting volume size in bytes for EKS

prashanthpx:
**What type of PR is this?**
Bug

Reports EKS volume size in bytes instead of GiB

No

No

For 2.5

### DIFF
--- a/drivers/volume/aws/aws.go
+++ b/drivers/volume/aws/aws.go
@@ -15,6 +15,7 @@ import (
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	snapshotVolume "github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
 	"github.com/kubernetes-sigs/aws-ebs-csi-driver/pkg/cloud"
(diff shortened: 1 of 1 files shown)


drivers/volume/aws/aws.go
```diff
@@ -369,8 +369,9 @@ func (a *aws) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 		case "completed":
 			vInfo.Status = storkapi.ApplicationBackupStatusSuccessful
 			vInfo.Reason = "Backup successful for volume"
-			vInfo.TotalSize = uint64(*snapshot.VolumeSize)
-			vInfo.ActualSize = uint64(*snapshot.VolumeSize)
+			// converting to bytes
+			vInfo.TotalSize = uint64(*snapshot.VolumeSize) * 1024 * 1024 * 1024
```
disrani-px, line 373: Use constants from here: https://github.com/docker/go-units/blob/master/size.go#L24

prashanthpx: Fixed. Use the following libopenstorage/openstorage/pkg/units/units.go

disrani-px (approved):

PR #740, opened 2020-11-11
[CSI] Make VolumeSnapshot objects unique per backup request

ggriffiths:

**What type of PR is this?**
bug

Appends backup UID to the end of each snapshot.

This makes it harder for different backups to conflict with each other.

This was spotted because an orphaned Snapshot was not cleaned up (before #739 merged) 

No

No

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -37,7 +37,7 @@ const (
 	// not to be confused with a CSI Driver Name
 	storkCSIDriverName = "csi"
 	// snapshotPrefix is appended to CSI backup snapshot
-	snapshotPrefix = "snapshot-"
+	snapshotPrefix = "snapshot"
 	// snapshotClassNamePrefix is the prefix for snapshot classes per CSI driver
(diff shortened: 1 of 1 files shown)


drivers/volume/csi/csi.go
```diff
@@ -569,6 +570,7 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 
 		case time.Now().After(snapshot.CreationTimestamp.Add(snapshotTimeout)):
 			vInfo.Status = storkapi.ApplicationBackupStatusFailed
+			backup.Status.Status = storkapi.ApplicationBackupStatusFailed
```
disrani-px, line 573: This shouldn't be marking the entire backup failed, that should be done by the controller. The drivers should only mark the individual volumes as failed. Is the controller not marking it as failed?

ggriffiths: I thought this was needed, but some OCS testing earlier hit a timeout px-backup behaved correctly. Will remove this

disrani-px (approved):

PR #745, opened 2020-11-14
[CSI] Namespace mapping, skip VS obj in includeObject, and UID fix

ggriffiths:

**What type of PR is this?**
bug

* Some CSI providers require specific VSClass parameters. We should not delete the VSClass as the user may have added their driver-specific settings. We will only update the VSClass to have the retain policy, as that is required for backups. We will still create the VSClass if it does not exist.
* Skip vs object in includeObject (to hide from UI and ignore in backups)
* add namespace mapping support
* add CSI restore timeout

No

No

No

### DIFF
--- a/pkg/resourcecollector/resourcecollector.go
+++ b/pkg/resourcecollector/resourcecollector.go
@@ -94,6 +94,9 @@ func (r *ResourceCollector) Init(config *restclient.Config) error {
 
 func resourceToBeCollected(resource metav1.APIResource, grp schema.GroupVersion, crdKinds []metav1.GroupVersionKind, optionalResourceTypes []string) bool {
 	for _, res := range crdKinds {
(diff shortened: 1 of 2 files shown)


ggriffiths:
Tested this fix on the OCS cluster.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -511,6 +511,11 @@ func (r *ResourceCollector) includeObject(
 		return false, err
 	}
 
+	// Skip VolumeSnapshot objects
+	if objectType.GetKind() == "VolumeSnapshot" {
```
disrani-px, line 515: This needs to be in `resourceToBeCollected()`

disrani-px (approved):

PR #747, opened 2020-11-17
[CSI] Restore cleanup without backup obj and various fixes

ggriffiths:

**What type of PR is this?**
bug

* When a restore is issued through PX-Backup, it creates a backup in the namespace and cleans it up when finished. There was some race where the cleanup was failing to find the backup object to cleanup the VS objects.
* This PR makes the cleanup independent of the backup object, by only cleaning up objects associated with the given restore UID.
* Uses PVC size for backup/restore info sizes
* Fix VolumeSnapshot skip check to not look at group

No

No

No

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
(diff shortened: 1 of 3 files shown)


drivers/volume/csi/csi.go
```diff
@@ -1058,6 +1118,14 @@ func (c *csi) restorePVC(
 	return pvc, nil
 }
 
+func (c *csi) getRestoreSnapshotName(pvc *v1.PersistentVolumeClaim, restore *storkapi.ApplicationRestore) string {
+	return fmt.Sprintf("%s-vs-%s-%s", snapshotRestorePrefix, string(restore.UID), pvc.Name)
```
disrani-px, line 1122: Since the name can be close to the max length limit this constructed name can exceed the limit.

Can we use just the last part of the UID of the restore and PVC to fix the size of the names?

ggriffiths: Yeah, I updated this in the PR. I also updated the px-backup deletion objects to use similar naming.

disrani-px (approved):

PR #748, opened 2020-11-20
[CSI] Fix multi-namespace backups

ggriffiths:

**What type of PR is this?**
bug

* Improves GetBackupStatus to handle cleanup and success states better
* Adds deletion retry if deletion timestamp went away
* Cleanup timeout after 5min from point of deletion 

No

No

No

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -65,6 +65,8 @@ const (
 
 	// snapshotTimeout represents the duration to wait before timing out on snapshot completion
 	snapshotTimeout = time.Minute * 5
+	// snapshotTimeout represents the duration to wait before timing out on snapshot completion
+	snapshotCleanupTimeout = time.Minute * 5
 	// restoreTimeout is the duration to wait before timing out the restore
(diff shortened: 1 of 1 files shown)


drivers/volume/csi/csi.go
```diff
@@ -634,14 +683,11 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 		if contentName != "" {
 			vscError = c.getSnapshotContentError(contentName)
 		}
-
 		switch {
 		case volumeSnapshotReady && volumeSnapshotContentReady:
-			vInfo.Status = storkapi.ApplicationBackupStatusInCleanup
-			vInfo.Reason = "Backup uploaded, cleaning up snapshot objects"
+			vInfo.Reason = "Snapshot finished, pending completion of other snapshots"
```
disrani-px, line 688: As discussed offline we'll revisit this in a future release to all individual volumes to be marked as Successful

disrani-px (approved):

PR #758, opened 2020-12-04
[CSI] Remove cleanup checks for GetBackupStatus

ggriffiths:

**What type of PR is this?**
bug

Removes the cleanup checks to improve the reliability of backup status.

No

No

2.6

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -65,8 +65,6 @@ const (
 
 	// snapshotTimeout represents the duration to wait before timing out on snapshot completion
 	snapshotTimeout = time.Minute * 5
-	// snapshotTimeout represents the duration to wait before timing out on snapshot completion
-	snapshotCleanupTimeout = time.Minute * 5
 	// restoreTimeout is the duration to wait before timing out the restore
 	restoreTimeout = time.Minute * 5
 )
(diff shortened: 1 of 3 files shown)


ggriffiths:
Tested on pure cluster

drivers/volume/csi/csi.go
```diff
@@ -549,10 +547,6 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 	vsClassMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshotClass)
 	for _, vInfo := range backup.Status.Volumes {
 		if vInfo.DriverName != storkCSIDriverName {
-			continue
-		}
-		// Once a snapshot has been backed up and cleaned up, skip it
-		if vInfo.Status == storkapi.ApplicationBackupStatusSuccessful {
 			volumeInfos = append(volumeInfos, vInfo)
```
siva-portworx, line 550: If it is not a CSI driver, we do not need to append the volume to the volumeInfos. We can just have continue alone.

siva-portworx (approved):

drivers/volume/csi/csi.go
ggriffiths: fixed this

PR #765, opened 2020-12-15
Fix crd migration issue for api extensionV1 crds

ram-infrac:
**What type of PR is this?**
>bug

Allow CRDs using apiextenstion v1 api to migrate/backup/restore. Fix replica count issue for crds with `int` as suspendValue

no

np

2.6.1

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -12,6 +12,7 @@ import (
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/controllers"
 	"github.com/libopenstorage/stork/pkg/crypto"
+	"github.com/libopenstorage/stork/pkg/k8sutils"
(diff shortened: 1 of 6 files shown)


disrani-px (approved):

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -645,17 +649,66 @@ func (a *ApplicationRestoreController) downloadCRD(
...
-			return err
+		// wait for crd to be ready
+		if err := k8sutils.ValidateCRD(client, crd.GetName()); err != nil {
+			logrus.Warnf("Unable to validate crds %v,%v", crd.GetName(), err)
 		}
+
 	}
+
+	for _, crd := range crdsV1 {
+		if val, ok := regCrd[crd.GetName()]; ok && val {
```
disrani-px, line 681: I don't see the key for this map being set to true for any key. I only see it being set to false on line 669

ram-infrac: thanks for catching this, added at line no 668

disrani-px (approved):

PR #777, opened 2021-01-15
stor:335 Adding batch processing logic in passing namespace list to the GetResources API of resourceCollector pkg

siva-portworx:
**What type of PR is this?**
bug

GetResources takes more time, if we have more number of namespaces in namespace list. This make backup looks to be hang as the CR is not updated. This makes the px-backup to mark the backup as failed with timeout error.

So, submitting it in batches and in between each batch, we will updating the LastUpdateTimestamp in backup CR to show that backup is progressing and avoid timeout failure.

No

No

Yes, to 2.6 branch.

**Testing:
Testcase1: ( To make sure that new logic does not add any extra time.)**
Increased the timeout value to 2 hours in px-backup.
Added 750 namespaces.
Added debug statement to calculate the time taken by GetResources with the new logic and old logic.
**New logic:**
siva-test-1-523b642 time taken for GetResources batching **52m49.060242546s**"
**Old logic:**
time="2021-01-15T13:15:36Z" level=info msg="Time taken by GetResources api **52m49.045374967s**"

**Testcase2:**
Tried with default timeout value of 30 sec with old code and new code.
With old code, it failed with timeout error.
With new code, it is success.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -53,10 +53,11 @@ const (
 	backupCancelBackoffFactor       = 1
 	backupCancelBackoffSteps        = math.MaxInt32
 
-	allNamespacesSpecifier = "*"
-	backupVolumeBatchCount = 10
-	maxRetry               = 10
-	retrySleep             = 10 * time.Second
+	allNamespacesSpecifier    = "*"
+	backupVolumeBatchCount    = 10
+	backupResourcesBatchCount = 15
+	maxRetry                  = 10
+	retrySleep                = 10 * time.Second
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -968,21 +969,67 @@ func (a *ApplicationBackupController) backupResources(
...
+		allObjects = append(allObjects, objects...)
+		// Do a dummy update to the backup CR to update only the last update timestamp
+		namespacedName := types.NamespacedName{}
+		namespacedName.Namespace = backup.Namespace
+		namespacedName.Name = backup.Name
+		for i := 0; i < maxRetry; i++ {
+			err := a.client.Get(context.TODO(), namespacedName, backup)
+			if err != nil {
+				time.Sleep(retrySleep)
+				continue
```
prashanthpx, line 999: If all the attempts of `client.Get()` fails, do we need to still proceed to update() or fail from here?

prashanthpx (approved):
LGTM

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: If the Get fails continuously, we will return error after the for loop.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -968,21 +969,67 @@ func (a *ApplicationBackupController) backupResources(
...
-		true)
-	if err != nil {
-		log.ApplicationBackupLog(backup).Errorf("Error getting resources: %v", err)
-		return err
+	namespacelist := backup.Spec.Namespaces
+	// GetResources takes more time, if we have more number of namespaces
+	// So, submitting it in batches and in between each batch,
+	// updating the LastUpdateTimestamp to show that backup is progressing
+	var allObjects []runtime.Unstructured
+	allObjects = make([]runtime.Unstructured, 0)
```
disrani-px, line 977:
```suggestion
	allObjects := make([]runtime.Unstructured, 0)
```

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -968,21 +969,67 @@ func (a *ApplicationBackupController) backupResources(
...
+			if err != nil {
+				time.Sleep(retrySleep)
+				continue
+			} else {
+				break
+			}
+		}
+		if err != nil {
+			return err
+		}
```
disrani-px, line 1012: This will never be hit

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Fixed

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Fixed. using error variable which was declared outside the for loop, so that check outside the for loop will use correct err variable.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -968,21 +969,63 @@ func (a *ApplicationBackupController) backupResources(
...
+			if err != nil {
+				time.Sleep(retrySleep)
+				continue
+			} else {
+				break
+			}
+		}
+	}
+	var updatedAllObjects []runtime.Unstructured
+	updatedAllObjects = make([]runtime.Unstructured, 0)
```
disrani-px, line 1011:
```suggestion
	updatedAllObjects := make([]runtime.Unstructured, 0)
```

disrani-px (approved):

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Missed it. Fixed now.

PR #788, opened 2021-02-03
Reduce the backupVolumeBatchCount value to 3 from 10, to avoid timeout failure.

siva-portworx:
**What type of PR is this?**
bug

Reduce the backupVolumeBatchCount value to 3 from 10, to avoid timeout failures.
If batch is 10 and if there are more failures in CloudBackupCreate() and retried, the CR is not getting updated with the progress for more than 30 mints. This makes the px-backup to mark the backup with timeout failure.

No

Yes, 2.6 branch.

**Testing:**
Validated that there is not much difference in the time taken to complete the StartBackup for 15 volume, irrespective of whether volumebatch count is 10 or 3.

With volumebatch count of 10:time="2021-02-03T07:11:05Z" level=info msg="sivakumar total time taken for startBackup ------> 33.440006339s"

With volumebatch count of 3:time="2021-02-03T07:37:25Z" level=info msg="sivakumar ### total time taken for startBackup ------> 33.902654002s"

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -5,8 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


prashanthpx (approved):

prashanthpx (approved):
LGTM

PR #798, opened 2021-03-10
Added logic to exclude the controlled resources from GetResources API.

siva-portworx:
**What type of PR is this?**
bug

- Added logic to exclude the controlled resources from GetResources API.
- Define a list for excluding controlled resources being returned as part of 
   response of resourceCollector GetResources API.
- For now added configmap kube-root-ca.crt, which was getting created
   by kube-controller-manager in every namespace.

no

no

yes, to 2.6 branch.

**Testing:**
Tested with k8s 1.20.0 that GetResources API does not return the kube-root-ca.crt configMap.

```
Actual configMap in the namespace test:
[root@central-steel-drifter-0 tmp]#
[root@central-steel-drifter-0 tmp]# kubectl get configmap -n test
NAME               DATA   AGE
kube-root-ca.crt   1      25h
test1              0      3m41s
testcm             0      40h
With fix, we are excluding kube-root-ca.crt  in GetResources API:
[root@central-steel-drifter-0 tmp]# /tmp/get
allobjects: ---> [0xc00064bce8 0xc00064bcf0]
name: test1
name: testcm

With out fix, we were including kube-root-ca.crt :
[root@central-steel-drifter-0 tmp]# /tmp/get
allobjects: ---> [0xc0003cae40 0xc0003cae48 0xc0003cae50]
name: kube-root-ca.crt
name: test1
name: testcm
[root@central-steel-drifter-0 tmp]#
```

### DIFF
--- a/pkg/resourcecollector/resourcecollector.go
+++ b/pkg/resourcecollector/resourcecollector.go
@@ -330,6 +330,8 @@ func (r *ResourceCollector) objectToBeCollected(
 		return r.roleBindingToBeCollected(object)
 	case "Ingress":
 		return r.ingressToBeCollected(object)
+	case "ConfigMap":
+		return r.configmapToBeCollected(object)
 	}
 
 	return true, nil
--- a/pkg/resourcecollector/configmap.go
+++ b/pkg/resourcecollector/configmap.go
@@ -1,0 +1,30 @@
+package resourcecollector
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func (r *ResourceCollector) configmapToBeCollected(
+	object runtime.Unstructured,
+) (bool, error) {
+	metadata, err := meta.Accessor(object)
+	if err != nil {
+		return false, err
+	}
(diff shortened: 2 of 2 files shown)


pkg/resourcecollector/resourcecollector.go
```diff
@@ -273,6 +281,22 @@ func skipOwnerRefCheck(annotations map[string]string) bool {
 	return false
 }
 
+// excludeControlledResources will check whther the given resource is part of
+// excludeControlledResourcesList. If present, it will not be included in the result
+// of the GetResources API.
+func (r *ResourceCollector) excludeControlledResources(name string, kind string) bool {
```
disrani-px, line 287: Would be better to do this like other resource types in `objectToBeCollected()`. Create configmap.go in this package and add  a function called `configMapToBeCollected(...)`

siva-portworx: Sure. will fix it.

siva-portworx: Fixed the comment.

ram-infrac (approved):

prashanthpx (approved):

PR #813, opened 2021-03-25
Clusterpair failures

Rohit-PX:
**What type of PR is this?**
integration-test

Integration test for clusterpair failures

no

no

2.6


test/integration_test/migration_test.go
```diff
@@ -1,4 +1,4 @@
-// +build integrationtest
+// +buil integrationtest
```
adityadani, line 1: looks like a typo ? - the build is failing as well

adityadani (approved):
lgtm after travis is fixed

test/integration_test/migration_test.go
Rohit-PX: yes, this was a typo. Fixed.

PR #816, opened 2021-03-30
rename stork prometheus metrics with standard prefix

ram-infrac:
**What type of PR is this?**
>cleanup

Standardize stork prometheus metrics with set prefix `stork`

no

yes
-->
```release-note
All stork prometheus metrics will now have `stork_` as prefix, eg
stork_application_backup_status, stork_migration_status, stork_hyperconverged_pods_total etc
All existing metrics will now move over to `stork_` prefix convention 
```

yes, 2.6.3


adityadani (approved):
The release note should mention that all the existing metrics will move over to this new model which means that reporting will stop for the old metric names

PR #817, opened 2021-04-01
Add pluralizer rules for prometheus crd

ram-infrac:
**What type of PR is this?**
>bug

Adds pluraliser rule for prometheus CR detections during backup/restore and migration. 

no
no

2.6.3

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -1291,7 +1291,11 @@ func (m *MigrationController) applyResources(
 			return err
 		}
 	}
-
+	// TODO: we should use k8s code generator logic to pluralize
+	// crd resources instead of depending on inflect lib
+	ruleset := inflect.NewDefaultRuleset()
+	ruleset.AddPlural("quota", "quotas")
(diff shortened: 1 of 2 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1291,7 +1291,11 @@ func (m *MigrationController) applyResources(
 			return err
 		}
 	}
-
+	// TODO: we should use k8s code generator logic to pluralize
+	// crd resources instead of depending on inflect lib
+	ruleset := inflect.NewDefaultRuleset()
+	ruleset.AddPlural("quota", "quotas")
+	ruleset.AddPlural("prometheus", "prometheuses")
```
piyush-nimbalkar, line 1298: Shouldn't line 1305 be using this ruleset instead of `inflect.Pluralize`

ram-infrac: yes, good catch. thanks

piyush-nimbalkar (approved):

adityadani (approved):

PR #827, opened 2021-04-20
Get Stork to use new secure port instead of 9001

puremudassir:
**What type of PR is this?**
>bug
>feature
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

PWX-18788

no

```release-note
no, no change from the user perspective
```

No, but is dependent on other changes... i.e. this must be merged after the following
https://github.com/libopenstorage/openstorage/pull/1855
https://github.com/libopenstorage/operator/pull/284
https://github.com/portworx/px-installer/pull/1006


ggriffiths (approved):
Looks good. Is there a plan in the operator to configure stork with 9001 for older PX versions?  

I.e. stork master on PX 2.7.0 will not work with this change by default. A quick check in the operator could fix this.

puremudassir:
> Looks good. Is there a plan in the operator to configure stork with 9001 for older PX versions?
> 
> I.e. stork master on PX 2.7.0 will not work with this change by default. A quick check in the operator could fix this.

@adityadani ... Is there a scenario where a newer version of stork will be deployed with an older version of px ?

adityadani:
@puremudassir That is a very common scenario where stork gets updated to the latest version but dependent drivers like PX are still on older versions.

puremudassir:
> @puremudassir That is a very common scenario where stork gets updated to the latest version but dependent drivers like PX are still on older versions.

@adityadani , @ggriffiths What other inter-dependencies do we need to be aware of for operator installs?
Newer version of operator can install either an older version of px (TLS secured port not available) or the newer version of px (TLS secured port will be available). Operator needs to ensure that the right service ports are exposed. Stork needs to be able to work in both scenarios. Same applies for autopilot (we have a different Jira for that).

Will we have an older version of Operator, older version of px, but newer version of stork?

I'm assuming that for Daemonset installs, we will need to change central to intelligently create `portworx-service` and `portworx-api` Services with the right ports (i.e. TLS secured port available or not depending on px version). Will create a separate JIRA for that.

What else?

PR #835, opened 2021-05-25
Migrate updated PVC specs each time

ram-infrac:
**What type of PR is this?**
>bug

Fixes issue where  pvc specs are  updated only on  first migration and consecutive migration schedules skips pvc spec update

no

yes
-->
```
Target PVC's does not reflect source size when updated
```

yes, 2.6

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -57,6 +57,11 @@ const (
 	// StorkMigrationCRDDeactivateAnnotation is the annotation used to keep track of
 	// the value to be set for deactivating crds
 	StorkMigrationCRDDeactivateAnnotation = "stork.libopenstorage.org/migrationCRDDeactivate"
+	// storageClassAnnotation for pvc sc
+	storageClassAnnotation = "volume.beta.kubernetes.io/storage-class"
+	// PVReclaimAnnotation for pvc's reclaim policy
+	PVReclaimAnnotation = "stork.libopenstorage.org/reclaimPolicy"
+
 	// Max number of times to retry applying resources on the desination
 	maxApplyRetries      = 10
 	cattleAnnotations    = "cattle.io"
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1398,6 +1441,90 @@ func (m *MigrationController) applyResources(
...
+			log.MigrationLog(migration).Errorf("Error deleting %v %v during migrate: %v", pvc.Kind, pvc.GetName(), err)
+		} else {
+			for i := 0; i < deletedMaxRetries; i++ {
+				obj, err := adminClient.CoreV1().PersistentVolumeClaims(pvc.GetNamespace()).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+				if err != nil && errors.IsNotFound(err) {
+					break
+				}
+				createTime := obj.GetCreationTimestamp()
+				if deleteStart.Before(&createTime) {
+					logrus.Warnf("Object[%v] got re-created after deletion. So, Ignore wait. deleteStart time:[%v], create time:[%v]",
```
adityadani, line 1468: nit:
```suggestion
					logrus.Warnf("Object[%v] got re-created after deletion. Not retrying deletion. deleteStart time:[%v], create time:[%v]",
```

pkg/migration/controllers/migration.go
```diff
@@ -1398,6 +1441,90 @@ func (m *MigrationController) applyResources(
...
+						obj.GetName(), deleteStart, createTime)
+					break
```
adityadani, line 1470: why not mark the migration failed here with this error message? 
Anyway the migration is going to fail once we try to create the PVC on line1476. At least we will update with the real reason why it failed

adityadani (approved):
looks good

pkg/migration/controllers/migration.go
ram-infrac: Migration will be marked as PartialSuccess since its failed as application stage.  We will be adding error we receive from create api in status field for failed pvc
`Error applying resource: <err from create pvc api`

PR #838, opened 2021-05-29
Yum has vulnerabilty issues, no need to install and use that.  Just u…

jrivera-px:
…se microdnf.

Signed-off-by: Jose Rivera <jose@portworx.com>

**What type of PR is this?**
cleanup

Fixes security vulnerability in some of the packages.   By not installing 'yum' the vulnerabilities in the dependent  packages disappear since the packages are not installed.

no

v2.6.3


harsh-px (approved):
@jrivera-px did you verify if the `python3 ca-certificates tar gzip` are still present after using microdnf?

adityadani (approved):

ram-infrac (approved):

PR #840, opened 2021-06-03
[STOR-383] add metrics for volume snapshot schedule

JimOuPure:
adding counter metrics for error/ready/pending volume snapshot.

### DIFF
--- a/pkg/metrics/volumesnapshotschedule.go
+++ b/pkg/metrics/volumesnapshotschedule.go
@@ -1,0 +1,44 @@
+package metrics
+
+import (
+	"fmt"
+
+	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var (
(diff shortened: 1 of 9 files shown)


JimOuPure:
I am updating some files under vendor in this PR just for review purpose, will move the change to corresponding repo before merging them.

pkg/metrics/volumesnapshotschedule.go
```diff
@@ -0,0 +1,107 @@
...
+	labels[metricName] = volumeSnapshotSchedule.Name
+	labels[metricNamespace] = volumeSnapshotSchedule.Namespace
+	labels[metricPolicy] = volumeSnapshotSchedule.Spec.SchedulePolicyName
+	suspend := volumeSnapshotSchedule.Spec.Suspend
+
+	if volumeSnapshotSchedule.DeletionTimestamp != nil {
+		volumeSnapshotScheduleStatusCounter.Delete(labels)
+		return nil
+	}
+	if *suspend {
```
adityadani, line 50: should we switch this behavior 
- if suspended then set to 1
- else set to 0

This matches with snapshot ready - 0

JimOuPure: sounds good, had it switched.

ram-infrac (approved):

adityadani (approved):

PR #844, opened 2021-06-08
Allow option to prevent resource modification during backup/migration

ram-infrac:
**What type of PR is this?**
>enhancement

Add CRD option to `ApplicationBackup` and `Migration` for skipping resource field update during backup/migration operation. 

yes, User can now specify `SkipServiceUpdate` in migration/backup CR specs so that stork will not modify cluster/loadbalancer ip field during resource collections
eg.: 
```
kind: ApplicationBackup
metadata:
  name: backup-mysql
  namespace: mysql
spec:
  backupLocation: mysql
  namespaces:
  - mysql
  reclaimPolicy: Delete
  skipServiceUpdate: true
 ```

yes
```release-note
Users can now specify `skipServiceUpdate` in backup/migration spec to avoid updating Loadbalancers IP/clusterIP field. Alternatively they can also use `stork.libopenstorage.org/skip-modify-resource` for single service resource
```

yes, 2.6.4

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -757,6 +757,14 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration, v
 	}
 	resKinds := make(map[string]string)
 	var updateObjects []runtime.Unstructured
+
+	// Don't modify resources if mentioned explicitly in specs
+	if *migration.Spec.SkipServiceUpdate {
+		if m.resourceCollector.Opts == nil {
+			m.resourceCollector.Opts = make(map[string]string)
+		}
+		m.resourceCollector.Opts[resourcecollector.ServiceKind] = "true"
+	}
 	allObjects, err := m.resourceCollector.GetResources(
 		migration.Spec.Namespaces,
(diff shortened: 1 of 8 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1013,6 +1021,15 @@ func (m *MigrationController) checkAndUpdateDefaultSA(
 			destSA.ImagePullSecrets = append(destSA.ImagePullSecrets, s)
 		}
 	}
+	// merge annotation for SA
+	if destSA.Annotations != nil {
```
adityadani, line 1025: This should be switched. You should be checking for sourceSA.Annotation != nil and create a destSA.Annotations map.

adityadani (approved):
looks good after comment is fixed.

pkg/migration/controllers/migration.go
ram-infrac: This is migration path - 
- we try to create service account on destination cluster
if already exists- 
  - fetch destination service account
  - update destination annotation with source cluster annotation (merge in case of same key found)

PR #846, opened 2021-06-11
Add a new annotation to skip PVCs from stork's scheduler scoring algorithm

adityadani:

**What type of PR is this?**

feature

- Use the stork.libopenstorage.org/skip-scheduler-scoring: true annotation on PVCs
  which should not be considered while scoring nodes in scheduler's prioritize request.
- Currently this change is being done only for the Portworx storage driver.

No

Yes

```release-note
Use annotation `stork.libopenstorage.org/skipSchedulerScoring` on the PVCs whose replicas should not be considered while scoring nodes in stork scheduler.
```

Yes - 2.6


pkg/extender/extender_test.go
```diff
@@ -543,6 +548,58 @@ func multipleVolumeTest(t *testing.T) {
...
+	provNodes = []int{1, 2}
+	if err := driver.ProvisionVolume("excluded-volume", provNodes, 1, map[string]string{skipScoringLabel: "true"}); err != nil {
+		t.Fatalf("Error provisioning volume: %v", err)
+	}
+
+	filterResponse, err := sendFilterRequest(pod, nodes)
+	if err != nil {
+		t.Fatalf("Error sending filter request: %v", err)
+	}
+	verifyFilterResponse(t, nodes, []int{0, 1, 2, 3, 4}, filterResponse)
```
ram-infrac, line 586: filter response should not include 2 ?

ram-infrac (approved):

pkg/extender/extender_test.go
adityadani: There is no change to the filtering logic. We are only skipping the volume in the prioritize request. Also even there the node will not be skipped but it will get a low score

PR #847, opened 2021-06-11
Do not fail backups when storage provider returns ErrStorageProviderBusy

adityadani:
**What type of PR is this?**
bug

Do not fail migration/backups when the storage provider returns ErrStorageProviderBusy

- Handle AWS and Portworx providers.
- If the storage provider returns a known "BUSY" error the driver wraps it
  and returns to the controller.
- If the controller sees this error, instead of marking the backup as failed it will retry
  in the next reconcile loop.

No

Yes
```release-note
Do not fail application backups when the storage provider returns a busy error code, instead backoff and retry the operation after sometime.
```

Yes - 2.6


ram-infrac (approved):

prashanthpx (approved):

drivers/volume/portworx/portworx.go
```diff
@@ -2855,7 +2854,13 @@ func (p *portworx) StartBackup(backup *storkapi.ApplicationBackup,
...
 			return true, nil
 		})
-		if err != nil || cloudBackupCreateErr != nil {
+		if err == nil {
+			// Only add volumeInfos if this was a successful backup
+			volumeInfos = append(volumeInfos, volumeInfo)
+		} else if err != nil || cloudBackupCreateErr != nil {
+			if _, ok := cloudBackupCreateErr.(*ost_errors.ErrCloudBackupServerBusy); ok {
+				return volumeInfos, &storkvolume.ErrStorageProviderBusy{Reason: cloudBackupCreateErr.Error()}
+			}
```
siva-portworx, line 2874: We might need to append the volumeInfo to the list, when we hit ErrExists error.

drivers/volume/portworx/portworx.go
```diff
@@ -2866,13 +2865,19 @@ func (p *portworx) StartBackup(backup *storkapi.ApplicationBackup,
 			}
 			return true, nil
 		})
+
 		if err != nil || cloudBackupCreateErr != nil {
+			if isCloudBackupServerBusyError(cloudBackupCreateErr) {
```
ram-infrac, line 2870: should we retry in case portworx driver returning server busy err

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -328,51 +333,111 @@ func (a *ApplicationRestoreController) getNamespacedObjectsToDelete(restore *sto
...
 		}
-		backupVolumeInfoMappings := make(map[string][]*storkapi.ApplicationBackupVolumeInfo)
-		objectMap := storkapi.CreateObjectsMap(restore.Spec.IncludeResources)
-		info := storkapi.ObjectInfo{
-			GroupVersionKind: metav1.GroupVersionKind{
-				Group:   "core",
-				Version: "v1",
-				Kind:    "PersistentVolumeClaim",
-			},
+		logrus.Infof("Updating restore  %s/%s in stage/stagus: %s/%s to volume stage", restore.Namespace, restore.Name, restore.Status.Stage, restore.Status.Status)
```
ram-infrac, line 351: Can you handle check for stale update meaning if cr in application stage don't update back it to volume stage

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -501,6 +509,12 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 			if pvc.Status.Phase != v1.ClaimBound || pvc.DeletionTimestamp != nil {
 				continue
 			}
+			// Don't backup PVCs which are already added to Status and for
+			// which backup was triggered
+			if _, isVolBackupDone := backupStatusVolMap[pvc.Namespace+"-"+pvc.Name]; isVolBackupDone {
```
ram-infrac, line 514: backupStatusVolMap is [string]string also I don't see `backupStatusVolMap` getting updated anywhere ?

drivers/volume/portworx/portworx.go
adityadani: So we already retrying with an  exponential backoff above. So it is better to try again in the next reconcile cycle

pkg/applicationmanager/controllers/applicationbackup.go
adityadani: oops.. remnant of some alternate approach I was trying..will remove this snippet

pkg/applicationmanager/controllers/applicationrestore.go
adityadani: Yes will do.

ram-infrac: you will need to check `ApplicationStage` too

ram-infrac (approved):

ram-infrac (approved):

PR #848, opened 2021-06-13
Backing up selected resource and all resoucres of given namespace

prashanthpx:
**What type of PR is this?**
-Enhancement

- User can choose to backup all resources in one namespace and selected resources
  in another namespace.

No

No

Yes, 2.6.4

**Manul test**
Here we are backing up all resources from `default` ns but only `configmap` from test2 ns

_Resources being backed up_

```
[root@prashanth-muck-bug-0 specs]# k get secret -ntest2
NAME                  TYPE                                  DATA   AGE
default-token-n2gh6   kubernetes.io/service-account-token   3      173m
secret1               Opaque                                1      171m
```

```
[root@prashanth-muck-bug-0 specs]# k get cm -ntest2
NAME               DATA   AGE
configmap1         0      173m
kube-root-ca.crt   1      173m
```

```
[root@prashanth-muck-bug-0 specs]# k get deployment
NAME    READY   UP-TO-DATE   AVAILABLE   AGE
mysql   1/1     1            1           18d
```

```
[root@prashanth-muck-bug-0 specs]# k get pvc
NAME         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
mysql-data   Bound    pvc-443df082-6b70-4a31-8d4a-b936228d5e4e   2Gi        RWO            px-mysql-sc    18d
```

**CRD used to backup**
```
kind: ApplicationBackup
metadata:
  name: backup-resr-4
  # make sure to create in kube-system if more than one ns
  # to be backed up
  namespace: kube-system
spec:
  backupLocation: mysql
  namespaces:
  - test2
  - default
  includeResources:
  - name: configmap1
    group: core
    kind: ConfigMap
    namespace: test2
    version: v1
  #- Secret
  #- LimitRange
  reclaimPolicy: Delete
  selectors:
  preExecRule:
  postExecRule:
```

**Content of resources.json file**
```
[
 {
  "apiVersion": "v1",
  "kind": "ConfigMap",
  "metadata": {
   "name": "configmap1",
   "namespace": "test2"
  }
 },
 {
  "apiVersion": "v1",
  "kind": "PersistentVolumeClaim",
  "metadata": {
   "annotations": {
    "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"PersistentVolumeClaim\",\"metadata\":{\"annotations\":{\"volume.beta.kubernetes.io/storage-class\":\"px-mysql-sc\"},\"labels\":{\"app\":\"mysql\"},\"name\":\"mysql-data\",\"namespace\":\"default\"},\"spec\":{\"accessModes\":[\"ReadWriteOnce\"],\"resources\":{\"requests\":{\"storage\":\"2Gi\"}}}}\n",
    "pv.kubernetes.io/bind-completed": "yes",
    "pv.kubernetes.io/bound-by-controller": "yes",
    "volume.beta.kubernetes.io/storage-class": "px-mysql-sc",
    "volume.beta.kubernetes.io/storage-provisioner": "kubernetes.io/portworx-volume"
   },
   "labels": {
    "app": "mysql"
   },
   "name": "mysql-data",
   "namespace": "default"
  },
  "spec": {
   "accessModes": [
    "ReadWriteOnce"
   ],
   "resources": {
    "requests": {
     "storage": "2Gi"
    }
   },
   "volumeMode": "Filesystem",
   "volumeName": "pvc-443df082-6b70-4a31-8d4a-b936228d5e4e"
  },
  "status": {
   "accessModes": [
    "ReadWriteOnce"
   ],
   "capacity": {
    "storage": "2Gi"
   },
   "phase": "Bound"
  }
 },
 {
  "apiVersion": "v1",
  "kind": "PersistentVolume",
  "metadata": {
   "annotations": {
    "kubernetes.io/createdby": "portworx-volume-dynamic-provisioner",
    "pv.kubernetes.io/bound-by-controller": "yes",
    "pv.kubernetes.io/provisioned-by": "kubernetes.io/portworx-volume"
   },
   "name": "pvc-443df082-6b70-4a31-8d4a-b936228d5e4e"
  },
  "spec": {
   "accessModes": [
    "ReadWriteOnce"
   ],
   "capacity": {
    "storage": "2Gi"
   },
   "claimRef": null,
   "persistentVolumeReclaimPolicy": "Delete",
   "portworxVolume": {
    "volumeID": "505825786065281652"
   },
   "storageClassName": "",
   "volumeMode": "Filesystem"
  },
  "status": {
   "phase": "Bound"
  }
 },
 {
  "apiVersion": "v1",
  "kind": "Service",
  "metadata": {
   "annotations": {
    "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Service\",\"metadata\":{\"annotations\":{},\"labels\":{\"app\":\"wordpress\"},\"name\":\"wordpress\",\"namespace\":\"default\"},\"spec\":{\"ports\":[{\"name\":\"web\",\"port\":80}],\"selector\":{\"app\":\"wordpress\",\"tier\":\"frontend\"},\"sessionAffinity\":\"ClientIP\",\"type\":\"NodePort\"}}\n"
   },
   "labels": {
    "app": "wordpress"
   },
   "name": "wordpress",
   "namespace": "default"
  },
  "spec": {
   "clusterIP": "",
   "clusterIPs": null,
   "externalTrafficPolicy": "Cluster",
   "loadBalancerIP": "",
   "ports": [
    {
     "name": "web",
     "nodePort": 30794,
     "port": 80,
     "protocol": "TCP",
     "targetPort": 80
    }
   ],
   "selector": {
    "app": "wordpress",
    "tier": "frontend"
   },
   "sessionAffinity": "ClientIP",
   "sessionAffinityConfig": {
    "clientIP": {
     "timeoutSeconds": 10800
    }
   },
   "type": "NodePort"
  },
  "status": {
   "loadBalancer": {}
  }
 },
 {
  "apiVersion": "apps/v1",
  "kind": "Deployment",
  "metadata": {
   "annotations": {
    "deployment.kubernetes.io/revision": "1",
    "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"apps/v1\",\"kind\":\"Deployment\",\"metadata\":{\"annotations\":{},\"labels\":{\"app\":\"mysql\"},\"name\":\"mysql\",\"namespace\":\"default\"},\"spec\":{\"replicas\":1,\"selector\":{\"matchLabels\":{\"app\":\"mysql\",\"version\":\"1\"}},\"strategy\":{\"rollingUpdate\":{\"maxSurge\":1,\"maxUnavailable\":1},\"type\":\"RollingUpdate\"},\"template\":{\"metadata\":{\"labels\":{\"app\":\"mysql\",\"version\":\"1\"}},\"spec\":{\"containers\":[{\"env\":[{\"name\":\"MYSQL_ROOT_PASSWORD\",\"value\":\"password\"}],\"image\":\"mysql:5.6\",\"livenessProbe\":{\"exec\":{\"command\":[\"sh\",\"-c\",\"mysqladmin -u root -p$MYSQL_ROOT_PASSWORD ping\"]},\"initialDelaySeconds\":70,\"periodSeconds\":10,\"timeoutSeconds\":5},\"name\":\"mysql\",\"ports\":[{\"containerPort\":3306}],\"readinessProbe\":{\"exec\":{\"command\":[\"sh\",\"-c\",\"mysql -u root -p$MYSQL_ROOT_PASSWORD -e \\\"select 1\\\"\"]},\"initialDelaySeconds\":60,\"periodSeconds\":10,\"timeoutSeconds\":5},\"volumeMounts\":[{\"mountPath\":\"/var/lib/mysql\",\"name\":\"mysql-persistent-storage\"}]}],\"volumes\":[{\"name\":\"mysql-persistent-storage\",\"persistentVolumeClaim\":{\"claimName\":\"mysql-data\"}}]}}}}\n"
   },
   "labels": {
    "app": "mysql"
   },
   "name": "mysql",
   "namespace": "default"
  },
  "spec": {
   "progressDeadlineSeconds": 600,
   "replicas": 1,
   "revisionHistoryLimit": 10,
   "selector": {
    "matchLabels": {
     "app": "mysql",
     "version": "1"
    }
   },
   "strategy": {
    "rollingUpdate": {
     "maxSurge": 1,
     "maxUnavailable": 1
    },
    "type": "RollingUpdate"
   },
   "template": {
    "metadata": {
     "creationTimestamp": null,
     "labels": {
      "app": "mysql",
      "version": "1"
     }
    },
    "spec": {
     "containers": [
      {
       "env": [
        {
         "name": "MYSQL_ROOT_PASSWORD",
         "value": "password"
        }
       ],
       "image": "mysql:5.6",
       "imagePullPolicy": "IfNotPresent",
       "livenessProbe": {
        "exec": {
         "command": [
          "sh",
          "-c",
          "mysqladmin -u root -p$MYSQL_ROOT_PASSWORD ping"
         ]
        },
        "failureThreshold": 3,
        "initialDelaySeconds": 70,
        "periodSeconds": 10,
        "successThreshold": 1,
        "timeoutSeconds": 5
       },
       "name": "mysql",
       "ports": [
        {
         "containerPort": 3306,
         "protocol": "TCP"
        }
       ],
       "readinessProbe": {
        "exec": {
         "command": [
          "sh",
          "-c",
          "mysql -u root -p$MYSQL_ROOT_PASSWORD -e \"select 1\""
         ]
        },
        "failureThreshold": 3,
        "initialDelaySeconds": 60,
        "periodSeconds": 10,
        "successThreshold": 1,
        "timeoutSeconds": 5
       },
       "resources": {},
       "terminationMessagePath": "/dev/termination-log",
       "terminationMessagePolicy": "File",
       "volumeMounts": [
        {
         "mountPath": "/var/lib/mysql",
         "name": "mysql-persistent-storage"
        }
       ]
      }
     ],
     "dnsPolicy": "ClusterFirst",
     "restartPolicy": "Always",
     "schedulerName": "default-scheduler",
     "securityContext": {},
     "terminationGracePeriodSeconds": 30,
     "volumes": [
      {
       "name": "mysql-persistent-storage",
       "persistentVolumeClaim": {
        "claimName": "mysql-data"
       }
      }
     ]
    }
   }
  },
  "status": {
   "availableReplicas": 1,
   "conditions": [
    {
     "lastTransitionTime": "2021-05-26T14:21:41Z",
     "lastUpdateTime": "2021-05-26T14:21:41Z",
     "message": "Deployment has minimum availability.",
     "reason": "MinimumReplicasAvailable",
     "status": "True",
     "type": "Available"
    },
    {
     "lastTransitionTime": "2021-05-26T14:21:41Z",
     "lastUpdateTime": "2021-05-26T14:23:37Z",
     "message": "ReplicaSet \"mysql-7597d5b8c7\" has successfully progressed.",
     "reason": "NewReplicaSetAvailable",
     "status": "True",
     "type": "Progressing"
    }
   ],
   "observedGeneration": 1,
   "readyReplicas": 1,
   "replicas": 1,
   "updatedReplicas": 1
  }
 }
]
```

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -469,225 +469,237 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 	// Start backup of the volumes if we don't have any status stored
 	pvcMappings := make(map[string][]v1.PersistentVolumeClaim)
 	backup.Status.Stage = stork_api.ApplicationBackupStageVolumes
-
-	objectMap := stork_api.CreateObjectsMap(backup.Spec.IncludeResources)
-	info := stork_api.ObjectInfo{
-		GroupVersionKind: metav1.GroupVersionKind{
-			Group:   "core",
-			Version: "v1",
-			Kind:    "PersistentVolumeClaim",
-		},
-	}
-
-	var pvcCount int
-	for _, namespace := range backup.Spec.Namespaces {
-		pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, backup.Spec.Selectors)
-		if err != nil {
-			return fmt.Errorf("error getting list of volumes to migrate: %v", err)
-		}
-
-		for _, pvc := range pvcList.Items {
-			// If a list of resources was specified during backup check if
-			// this PVC was included
-			info.Name = pvc.Name
-			info.Namespace = pvc.Namespace
-			if len(objectMap) != 0 {
-				if val, present := objectMap[info]; !present || !val {
-					continue
-				}
-			}
-
-			// Don't backup pending or deleting PVCs
-			if pvc.Status.Phase != v1.ClaimBound || pvc.DeletionTimestamp != nil {
-				continue
-			}
-			driverName, err := volume.GetPVCDriver(core.Instance(), &pvc)
-			if err != nil {
-				// Skip unsupported PVCs
-				if _, ok := err.(*errors.ErrNotSupported); ok {
-					continue
-				}
-				return err
-			}
-			if driverName != "" {
-				if pvcMappings[driverName] == nil {
-					pvcMappings[driverName] = make([]v1.PersistentVolumeClaim, 0)
-				}
-				pvcCount++
-				pvcMappings[driverName] = append(pvcMappings[driverName], pvc)
-			}
-		}
-	}
-	if backup.Status.Volumes == nil ||
-		(backup.Status.Volumes != nil && len(backup.Status.Volumes) != pvcCount) {
-		backup.Status.Volumes = make([]*stork_api.ApplicationBackupVolumeInfo, 0)
-	}
-
 	namespacedName := types.NamespacedName{}
-	namespacedName.Namespace = backup.Namespace
-	namespacedName.Name = backup.Name
-	if len(backup.Status.Volumes) != pvcCount {
+	if IsVolsToBeBackedUp(backup) {
+		isResourceTypePVC := IsResourceTypePVC(backup)
+		objectMap := stork_api.CreateObjectsMap(backup.Spec.IncludeResources)
+		info := stork_api.ObjectInfo{
+			GroupVersionKind: metav1.GroupVersionKind{
+				Group:   "core",
+				Version: "v1",
+				Kind:    "PersistentVolumeClaim",
+			},
+		}
 
-		for driverName, pvcs := range pvcMappings {
-			driver, err := volume.Get(driverName)
+		var pvcCount int
+		for _, namespace := range backup.Spec.Namespaces {
+			pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, backup.Spec.Selectors)
 			if err != nil {
-				return err
+				return fmt.Errorf("error getting list of volumes to migrate: %v", err)
 			}
-			batchCount := defaultBackupVolumeBatchCount
-			if len(os.Getenv(backupVolumeBatchCountEnvVar)) != 0 {
-				batchCount, err = strconv.Atoi(os.Getenv(backupVolumeBatchCountEnvVar))
+
+			for _, pvc := range pvcList.Items {
+				// If a list of resources was specified during backup check if
+				// this PVC was included
+				info.Name = pvc.Name
+				info.Namespace = pvc.Namespace
+				if len(objectMap) != 0 {
+					if resourcecollector.IsNsPresentInIncludeResource(objectMap, namespace) {
+						if val, present := objectMap[info]; !present || !val {
+							continue
+						}
+					} else {
+						// We could have case where includeResource has data, current ns is not part of includeResource
+						// and the user has given ResourceType != PVC. In this case we don't
+						// want to collect vol data from this ns
+						if !isResourceTypePVC {
+							break
+						}
+					}
+				}
+
+				// Don't backup pending or deleting PVCs
+				if pvc.Status.Phase != v1.ClaimBound || pvc.DeletionTimestamp != nil {
+					continue
+				}
+				driverName, err := volume.GetPVCDriver(core.Instance(), &pvc)
 				if err != nil {
-					batchCount = defaultBackupVolumeBatchCount
+					// Skip unsupported PVCs
+					if _, ok := err.(*errors.ErrNotSupported); ok {
+						continue
+					}
+					return err
+				}
+				if driverName != "" {
+					if pvcMappings[driverName] == nil {
+						pvcMappings[driverName] = make([]v1.PersistentVolumeClaim, 0)
+					}
+					pvcCount++
+					pvcMappings[driverName] = append(pvcMappings[driverName], pvc)
 				}
 			}
-			for i := 0; i < len(pvcs); i += batchCount {
-				batch := pvcs[i:min(i+batchCount, len(pvcs))]
-				volumeInfos, err := driver.StartBackup(backup, batch)
+		}
+
+		if backup.Status.Volumes == nil ||
+			(backup.Status.Volumes != nil && len(backup.Status.Volumes) != pvcCount) {
+			backup.Status.Volumes = make([]*stork_api.ApplicationBackupVolumeInfo, 0)
+		}
+
+		namespacedName.Namespace = backup.Namespace
+		namespacedName.Name = backup.Name
+		if len(backup.Status.Volumes) != pvcCount {
+
+			for driverName, pvcs := range pvcMappings {
+				driver, err := volume.Get(driverName)
 				if err != nil {
-					// TODO: If starting backup for a drive fails mark the entire backup
-					// as Cancelling, cancel any other started backups and then mark
-					// it as failed
-					message := fmt.Sprintf("Error starting ApplicationBackup for volumes: %v", err)
+					return err
+				}
+				batchCount := defaultBackupVolumeBatchCount
+				if len(os.Getenv(backupVolumeBatchCountEnvVar)) != 0 {
+					batchCount, err = strconv.Atoi(os.Getenv(backupVolumeBatchCountEnvVar))
+					if err != nil {
+						batchCount = defaultBackupVolumeBatchCount
+					}
+				}
+				for i := 0; i < len(pvcs); i += batchCount {
+					batch := pvcs[i:min(i+batchCount, len(pvcs))]
+					volumeInfos, err := driver.StartBackup(backup, batch)
+					if err != nil {
+						// TODO: If starting backup for a drive fails mark the entire backup
+						// as Cancelling, cancel any other started backups and then mark
+						// it as failed
+						message := fmt.Sprintf("Error starting ApplicationBackup for volumes: %v", err)
+						log.ApplicationBackupLog(backup).Errorf(message)
+						a.recorder.Event(backup,
+							v1.EventTypeWarning,
+							string(stork_api.ApplicationBackupStatusFailed),
+							message)
+						_, err = a.updateBackupCRWithRetry(
+							namespacedName,
+							stork_api.ApplicationBackupStatusFailed,
+							stork_api.ApplicationBackupStageFinal,
+							message,
+							nil,
+						)
+						if err != nil {
+							return err
+						}
+						return nil
+					}
+					backup, err = a.updateBackupCRWithRetry(
+						namespacedName,
+						stork_api.ApplicationBackupStatusInProgress,
+						backup.Status.Stage,
+						"Volume backups are in progress",
+						volumeInfos,
+					)
+					if err != nil {
+						return err
+					}
+				}
+			}
+
+			// Terminate any background rules that were started
+			for _, channel := range terminationChannels {
+				channel <- true
+			}
+			terminationChannels = nil
+
+			// Run any post exec rules once backup is triggered
+			if backup.Spec.PostExecRule != "" {
+				err = a.runPostExecRule(backup)
+				if err != nil {
+					message := fmt.Sprintf("Error running PostExecRule: %v", err)
 					log.ApplicationBackupLog(backup).Errorf(message)
 					a.recorder.Event(backup,
 						v1.EventTypeWarning,
 						string(stork_api.ApplicationBackupStatusFailed),
 						message)
-					_, err = a.updateBackupCRWithRetry(
-						namespacedName,
-						stork_api.ApplicationBackupStatusFailed,
-						stork_api.ApplicationBackupStageFinal,
-						message,
-						nil,
-					)
+
+					backup.Status.Stage = stork_api.ApplicationBackupStageFinal
+					backup.Status.FinishTimestamp = metav1.Now()
+					backup.Status.LastUpdateTimestamp = metav1.Now()
+					backup.Status.Status = stork_api.ApplicationBackupStatusFailed
+					backup.Status.Reason = message
+					err = a.client.Update(context.TODO(), backup)
 					if err != nil {
 						return err
 					}
-					return nil
+					return fmt.Errorf("%v", message)
 				}
-				backup, err = a.updateBackupCRWithRetry(
-					namespacedName,
-					stork_api.ApplicationBackupStatusInProgress,
-					backup.Status.Stage,
-					"Volume backups are in progress",
-					volumeInfos,
-				)
+			}
+		}
+
+		inProgress := false
+		// Skip checking status if no volumes are being backed up
+		if len(backup.Status.Volumes) != 0 {
+			drivers := a.getDriversForBackup(backup)
+
+			volumeInfos := make([]*stork_api.ApplicationBackupVolumeInfo, 0)
+			for driverName := range drivers {
+
+				driver, err := volume.Get(driverName)
 				if err != nil {
 					return err
 				}
-			}
-		}
 
-		// Terminate any background rules that were started
-		for _, channel := range terminationChannels {
-			channel <- true
-		}
-		terminationChannels = nil
-
-		// Run any post exec rules once backup is triggered
-		if backup.Spec.PostExecRule != "" {
-			err = a.runPostExecRule(backup)
-			if err != nil {
-				message := fmt.Sprintf("Error running PostExecRule: %v", err)
-				log.ApplicationBackupLog(backup).Errorf(message)
-				a.recorder.Event(backup,
-					v1.EventTypeWarning,
-					string(stork_api.ApplicationBackupStatusFailed),
-					message)
-
-				backup.Status.Stage = stork_api.ApplicationBackupStageFinal
-				backup.Status.FinishTimestamp = metav1.Now()
-				backup.Status.LastUpdateTimestamp = metav1.Now()
-				backup.Status.Status = stork_api.ApplicationBackupStatusFailed
-				backup.Status.Reason = message
-				err = a.client.Update(context.TODO(), backup)
+				status, err := driver.GetBackupStatus(backup)
 				if err != nil {
-					return err
+					return fmt.Errorf("error getting backup status for driver %v: %v", driverName, err)
 				}
-				return fmt.Errorf("%v", message)
+				volumeInfos = append(volumeInfos, status...)
 			}
-		}
-	}
+			backup.Status.Volumes = volumeInfos
 
-	inProgress := false
-	// Skip checking status if no volumes are being backed up
-	if len(backup.Status.Volumes) != 0 {
-		drivers := a.getDriversForBackup(backup)
-
-		volumeInfos := make([]*stork_api.ApplicationBackupVolumeInfo, 0)
-		for driverName := range drivers {
-
-			driver, err := volume.Get(driverName)
-			if err != nil {
-				return err
-			}
-
-			status, err := driver.GetBackupStatus(backup)
-			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/resourcecollector/resourcecollector.go
```diff
@@ -346,8 +346,20 @@ func (r *ResourceCollector) GetResources(
 					if !ok {
 						return nil, fmt.Errorf("error casting object: %v", o)
 					}
-
-					collect, err := r.objectToBeCollected(includeObjects, labelSelectors, resourceMap, runtimeObject, crbs, ns, allDrivers)
+					var collect bool
```
ram-infrac, line 349: we can process this earlier loop rather than for per object

pkg/resourcecollector/resourcecollector.go
```diff
@@ -380,6 +392,17 @@ func (r *ResourceCollector) GetResources(
 	return allObjects, nil
 }
 
+// IsNsPresnetInIncludeResource checks if a given ns is present in the IncludeResource object
+func IsNsPresnetInIncludeResource(includeObjects map[stork_api.ObjectInfo]bool, namespace string) bool {
```
ram-infrac, line 396: nit: IsNsPresentInIncludeResource

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -991,6 +991,12 @@ func (a *ApplicationBackupController) backupResources(
 	var err error
 	var resourceTypes []metav1.APIResource
 	// Listing all resource types
+	if len(backup.Spec.ResourceTypes) != 0 && len(backup.Spec.IncludeResources) != 0 {
+		// Currently this is not supported, error out
+		errMsg := "cannot have ResourceType and IncludeResources together"
+		log.ApplicationBackupLog(backup).Errorf(errMsg)
+		return fmt.Errorf(errMsg)
+	}
```
siva-portworx, line 999: Do we need to have this check in backupVolume as well, where we access the includeResource for extracting PVCs

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -991,6 +991,12 @@ func (a *ApplicationBackupController) backupResources(
 	var err error
 	var resourceTypes []metav1.APIResource
 	// Listing all resource types
+	if len(backup.Spec.ResourceTypes) != 0 && len(backup.Spec.IncludeResources) != 0 {
```
siva-portworx, line 994: Also how does the case where the resourceType mentioned as PVCs is handled currently? Looks we are not going though ResourceType in the backupVolume().

pkg/applicationmanager/controllers/applicationbackup.go
prashanthpx: Added in Handle

pkg/applicationmanager/controllers/applicationbackup.go
prashanthpx: So basically we are looking at the use case if a user doesn't specify PVC in ResourceType do we still need to backupvolumes?  Need to discuss this use case. @adityadani / @ram-infrac ?

siva-portworx: Also the case, where do we need to backup all the volume backup in a namespace, if namespace is not present in the includeresources? Same criteria that we apply for other resources in the backupResources.

adityadani: Ideally I want to follow the same protocol for both resources and volumes, which means we need to change the `backupVolumes()` implementation, but we will break backward compatibility. How about we do this?

1. If a PVC resource is provided in `IncludeResources` then only include those specific PVCs in that namespace while backing up and ignore other PVCs from that namespace.
2. If PVC is provided as a type in the list of `ResourceTypes` then across all the namespaces backup only the volumes.
3. If `ResourceTypes` list is **NOT** empty, but it does **NOT** contain PVCs, then lets **NOT** backup PVCs -> This is not the current behavior but I think this will be the right direction since it is less confusing when someone is using this CRD. This will break backward compatibility but PX-Backup does not use this specific behavior, so the impact will be less.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -238,7 +238,13 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 			err.Error())
 		return nil
 	}
-
+	// Listing all resource types
+	if len(backup.Spec.ResourceTypes) != 0 && len(backup.Spec.IncludeResources) != 0 {
+		// Currently this is not supported, error out
+		errMsg := "cannot have ResourceType and IncludeResources together"
```
adityadani, line 244: nit: `cannot set both ResourceType and IncludeResources in ApplicationBackup request`

adityadani:
Also this PR should be release noted, so please update your comment

pkg/applicationmanager/controllers/applicationbackup.go
prashanthpx: Yes, this makes sense and document the changes.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -469,225 +469,237 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+				if len(objectMap) != 0 {
+					if resourcecollector.IsNsPresentInIncludeResource(objectMap, namespace) {
+						if val, present := objectMap[info]; !present || !val {
+							continue
+						}
+					} else {
+						// We could have case where includeResource has data, current ns is not part of includeResource
+						// and the user has given ResourceType != PVC. In this case we don't
+						// want to collect vol data from this ns
+						if !isResourceTypePVC {
```
siva-portworx, line 505: This check could be outside the PVC for loop as it is generic. It need not be run for each PVC items.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -346,8 +351,15 @@ func (r *ResourceCollector) GetResources(
...
-					collect, err := r.objectToBeCollected(includeObjects, labelSelectors, resourceMap, runtimeObject, crbs, ns, allDrivers)
+					var collect bool
+					var err error
+					// If a namespace is present in both namespace list and IncludeResource Object,
+					// IncludeResource takes priority and only those resources are backed up.
+					// If a ns is only present in namespace list, all resources in that ns
+					// is backed up.
+					// With this now a user can choose to backup all resources in a ns and some
+					// selected resources from different ns
+					collect, err = r.objectToBeCollected(objectToInclude, labelSelectors, resourceMap, runtimeObject, crbs, ns, allDrivers)
```
siva-portworx, line 362: nit: Cann't we not skip calling objectToBeCollected directly here, if the objcetToInclude is empty?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1252,3 +1283,39 @@ func (a *ApplicationBackupController) createCRD() error {
 
 	return apiextensions.Instance().ValidateCRD(resource, validateCRDTimeout, validateCRDInterval)
 }
+
+// IsVolsToBeBackedUp for a given backupsepc do we need to have volumes backed up
```
siva-portworx, line 1287: typo: backupsepc

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1010,9 +1021,26 @@ func (a *ApplicationBackupController) backupResources(
 	allObjects := make([]runtime.Unstructured, 0)
 	for i := 0; i < len(namespacelist); i += backupResourcesBatchCount {
 		batch := namespacelist[i:min(i+backupResourcesBatchCount, len(namespacelist))]
-		if len(backup.Spec.ResourceTypes) == 0 {
+		var incResNsBatch []string
+		var resourceTypeNsBatch []string
+		for _, ns := range batch {
+			// As we support both includeResource and ResourceType to be mentioned
+			// match out ns for which we want to take incudeResource path and
```
siva-portworx, line 1028: typo: incudeResource

siva-portworx (approved):

pkg/resourcecollector/resourcecollector.go
adityadani: If objectToInclude is empty, that could also mean that it was not part of IncludeResources but should be considered as a part of a type provided in the list of ResourceTypes.

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: What I meant is that, we can move the check of moving to next NS outside. This need not to be done for PVC and   
then move to the next NS. Let say the line number 506 continue hits for every PVC that condition can be moved out.
```
if len(objectMap) != 0 {					
        if !resourcecollector.IsNsPresentInIncludeResource(objectMap, namespace) {
                   
               if !isResourceTypePVC {
               		continue
               }
        }
 }

File: pkg/applicationmanager/controllers/applicationbackup.go [hunk 10]
```diff
@@ -469,225 +469,237 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
-		},
-	}
-
-	var pvcCount int
-	for _, namespace := range backup.Spec.Namespaces {
-		pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, backup.Spec.Selectors)
-		if err != nil {
-			return fmt.Errorf("error getting list of volumes to migrate: %v", err)
+	namespacedName := types.NamespacedName{}
+	if IsVolsToBeBackedUp(backup) {
```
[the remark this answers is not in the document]
siva-portworx: This he has done because, he want to re-use the code that will move the status to ApplicationBackupStageApplications and call the backupResources(). We can not return nil from here.

pkg/resourcecollector/resourcecollector.go
siva-portworx: My understanding is that, we will not call GetResources() to collect resource for ResourcType.  We will be calling GetResourcesForType for resourceType case.
But any how, we need to have the other four check that we do in the objectToBeCollected. We might need to call objectToBeCollected, even if the objectToInclude is empty.
```
metadata, err := meta.Accessor(object)	
    if err != nil {		return false, err	}
	if SkipResource(metadata.GetAnnotations())
          {		return false, err	}
	// Skip if we've already processed this object	
	if _, ok := resourceMap[metadata.GetUID()]; ok {		return false, nil	}
	objectType, err := meta.TypeAccessor(object)	if err != nil {		return false, err	}
	```

File: pkg/applicationmanager/controllers/applicationbackup.go [hunk 11]
```diff
@@ -1010,9 +1021,26 @@ func (a *ApplicationBackupController) backupResources(
...
+			// match out ns for which we want to take incudeResource path and
+			// for which we want to take ResourceType path
+			if len(backup.Spec.ResourceTypes) != 0 {
+				if !resourcecollector.IsNsPresentInIncludeResource(objectMap, ns) {
+					resourceTypeNsBatch = append(resourceTypeNsBatch, ns)
+				} else {
+					incResNsBatch = append(incResNsBatch, ns)
+				}
+			} else {
+				incResNsBatch = batch
```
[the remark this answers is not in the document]
siva-portworx: Is the case1 in the document not does that?

pkg/applicationmanager/controllers/applicationbackup.go
prashanthpx: Lets take the following CRD 
```apiVersion: stork.libopenstorage.org/v1alpha1
kind: ApplicationBackup
metadata:
  name: case-9
  namespace: kube-system
spec:
  backupLocation: mysql
  namespaces:
  - ns1
  - ns2
  - ns3
  resourceTypes:
  - ConfigMap
  includeResources:
  - name: ns2-pvc-2
    group: core
    kind: PersistentVolumeClaim
    namespace: ns2
    version: v1
  reclaimPolicy: Delete
  selectors:
  preExecRule:
  postExecRule:
```

 In the above CRD, we would like to backup the following
 - Only PVC from ns2
 - For ns1 and ns3, we want to skip vol backups and only backup resouces.

This check if for every ns to know what they need to be backed up.
If a ns is not part of the includeResource list but includeResource has refererence to non-pvc resource, then we need to be skipping those resoruce.
Moving this to before ns check would not solve this issue and we would end up backing vols data for those ns even though ResourceType != PVC.

This check has to be executed only when the ns is not part of includeResource list and doesn't have any PVC  hence added here.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1252,3 +1283,39 @@ func (a *ApplicationBackupController) createCRD() error {
...
+		}
+		return false
+	}
+
+	return true
+}
+
+// IsResourceTypePVC check if given ResourceType is PVC
+func IsResourceTypePVC(backup *stork_api.ApplicationBackup) bool {
+	for _, resType := range backup.Spec.ResourceTypes {
```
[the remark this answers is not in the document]
prashanthpx: My guess is you are thinking from the use case where the user just gives ns and doesn't mention anything in ResourceType and includeResource where the user wants to take backup of everything in the ns.  In this case, `IsVolsToBeBackedUp()` returns true and none of the checks at line406 is true and hence all will be backed up.

This need not return `True` coz in `backupVolumes()` where this is used,  its' signifiance is only when there is something in includeResource. This function is just to say if ResoruceType has PVC and in the case of empty it doesn't have anything.

pkg/applicationmanager/controllers/applicationbackup.go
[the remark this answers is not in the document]
prashanthpx: Yup returning nil will not backupresource for which we want to skip backingup vols.

pkg/applicationmanager/controllers/applicationbackup.go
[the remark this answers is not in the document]
prashanthpx: Yup, case-1 in the document does the following 
For reference pasting here
```
[root@prashanth-muck-bug-0 stor-409]# cat case-1.yaml
# Takes backup of all resources in given ns
kind: ApplicationBackup
metadata:
name: case-1
namespace: kube-system
spec:
backupLocation: mysql
namespaces:
- ns1
- ns2
- ns3
reclaimPolicy: Delete
selectors:
preExecRule:
postExecRule:
```

pkg/resourcecollector/resourcecollector.go
prashanthpx: As discussed there is no change needed here as these checks are already done inside `objectToBeCollected()` API.

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: As discussed, the check "if !isResourceTypePVC " in 505 line, will done for all the PVC in the namespace even if the namespaces is not part of includeResources and global ResourceType does not have PVC in it. 
When we decide that we need to skip to next namespace in line 505, either we can have break instead of continue, that way we can skip pvc for loop Or have check outside the PVC for loop.

prashanthpx: Addressed and commited.

prashanthpx:
@adityadani Can you pls review as addressed few of review comments?

adityadani (approved):
looks good

PR #850, opened 2021-06-21
Prevent invalid CR updates for application backup

ram-infrac:
**What type of PR is this?**
>bug

Fix issue where 
- stork does dirty update of application backup CRs
- collect service account for user with no namespace access 

no
yes
```release-note
- Fix invalid CR updates of application backup
- Dont display/collect service account for namespace where user does not have access
```

yes, 2.6

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -420,7 +420,7 @@ func min(x, y int) int {
 	return y
 }
 
-func (a *ApplicationBackupController) updateBackupCRWithRetry(
+func (a *ApplicationBackupController) updateBackupCRInVolumeStage(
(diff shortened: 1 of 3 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -435,7 +435,11 @@ func (a *ApplicationBackupController) updateBackupCRWithRetry(
 			time.Sleep(retrySleep)
 			continue
 		}
-		if backup.Status.Stage == stork_api.ApplicationBackupStageFinal {
+		// since updateBackupCRWithRetry called during volume stage , make sure
```
adityadani, line 438: nit: can you rename this function to `updateBackupCRInVolumeStage`

adityadani (approved):

PR #857, opened 2021-07-07
Fix the issues caused by git merge conflicts.

adityadani:
**What type of PR is this?**
bug

Backup volume count was reset to 0 due to an incorrect change that got pushed in after resolving a 3 way merge conflict.

No

No

Yes - 2.6


ram-infrac (approved):
lgtm

PR #858, opened 2021-07-08
Update service account resource instead of delete/create

ram-infrac:
**What type of PR is this?**
>bug

Stork now update SA instead of recreating it now, we don't need to handle default SA seperately

no

yes, 2.6


Rohit-PX (approved):
Looks good to me.

adityadani (approved):

PR #860, opened 2021-07-11
Removed incorrect return statement in the backupVolumes function

siva-portworx:
**What type of PR is this?**
bug
Removed incorrect return statement in the backupVolumes function

No

No

2.6 

**Testing:**
Tested the failing usecase listed in the pb-1732 bug.

Screenshot 2021-07-11 at 12 21 21 PM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -498,7 +498,7 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 		for _, namespace := range backup.Spec.Namespaces {
 			pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, backup.Spec.Selectors)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -518,7 +518,6 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 						if !isResourceTypePVC {
 							break
 						}
-						return err
```
ram-infrac, line 521: let correct error msg at https://github.com/libopenstorage/stork/blob/master/pkg/applicationmanager/controllers/applicationbackup.go#L501 as well

ram-infrac (approved):

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Fixed.

prashanthpx (approved):

PR #864, opened 2021-07-15
Increase timeout for cassandara apps

Rohit-PX:
**What type of PR is this?**
integration-test

A lot of integration tests are failing because cassandra apps are taking longer to come up after migration (possibly 

No

No
```

Master

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	migrationRetryInterval = 10 * time.Second
-	migrationRetryTimeout  = 5 * time.Minute
+	migrationRetryTimeout  = 10 * time.Minute
 )
 
 func testMigration(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by adityadani — APPROVED** [review]

File: test/integration_test/migration_test.go [hunk 1]
```diff
@@ -27,7 +27,7 @@ import (
 
 const (
 	migrationRetryInterval = 10 * time.Second
-	migrationRetryTimeout  = 5 * time.Minute
+	migrationRetryTimeout  = 10 * time.Minute
```
ram-infrac, line 30: we also wanted to increase groupcloudsnapshot timeouts?

ram-infrac (approved):

test/integration_test/migration_test.go
Rohit-PX: Fixed.

PR #866, opened 2021-07-20
stor-435: Migration failing, pvc already exists

ram-infrac:
**What type of PR is this?**
>bug

Migration failing intermittently when run with `startApplication` set to true

no

no

2.6.4

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -1344,8 +1344,151 @@ func (m *MigrationController) applyResources(
 			return err
 		}
 	}
-	var pvObjects, pvcObjects []runtime.Unstructured
+	var pvObjects, pvcObjects, updatedObjects []runtime.Unstructured
+	// collect pv,pvc separately
 	for _, o := range objects {
+		objectType, err := meta.TypeAccessor(o)
+		if err != nil {
+			return err
+		}
+		switch objectType.GetKind() {
+		case "PersistentVolume":
+			pvObjects = append(pvObjects, o)
+		case "PersistentVolumeClaim":
+			pvcObjects = append(pvcObjects, o)
+		default:
+			updatedObjects = append(updatedObjects, o)
+		}
+	}
+	// create/update pv object with updated policy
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1344,116 +1344,60 @@ func (m *MigrationController) applyResources(
...
+				stork_api.MigrationStatusFailed,
+				fmt.Sprintf("Error unmarshalling pv resource: %v", err))
 			continue
 		}
-		for {
-			_, err = dynamicClient.Create(context.TODO(), unstructured, metav1.CreateOptions{})
+		log.MigrationLog(migration).Infof("Applying %v %v", pv.Kind, pv.GetName())
+		_, err = adminClient.CoreV1().PersistentVolumes().Create(context.TODO(), &pv, metav1.CreateOptions{})
+		if err != nil {
 			if err != nil && (errors.IsAlreadyExists(err) || strings.Contains(err.Error(), portallocator.ErrAllocated.Error())) {
```
adityadani, line 1378: You don't need the portallocator error check here anymore.

pkg/migration/controllers/migration.go
```diff
@@ -1536,6 +1487,100 @@ func (m *MigrationController) applyResources(
...
+			if err != nil && (errors.IsAlreadyExists(err) || strings.Contains(err.Error(), portallocator.ErrAllocated.Error())) {
+				switch objectType.GetKind() {
+				case "ServiceAccount":
+					err = m.checkAndUpdateDefaultSA(migration, o)
+				default:
+					// Delete the resource if it already exists on the destination
+					// cluster and try creating again
+					deleteStart := metav1.Now()
+					err = dynamicClient.Delete(context.TODO(), metadata.GetName(), metav1.DeleteOptions{})
+					if err != nil {
```
adityadani, line 1540: Shouldn't there be a IsNotFound check here so that it falls into the else condition ?

adityadani (approved):
looks good - had a few comments

PR #867, opened 2021-07-21
stor-439: Backup respective pv object for pvc resourcetypes

ram-infrac:
**What type of PR is this?**
>bug

Collect respective PV objects when PVC are passed as `resourceTypes` in backup CR

no

```release-note
For application backup CR no need to pass PV object for custom resource type backup,  user has to specify only `PersistentVolumeClaim` and its respective PV will be backed up
```

eg. 
```
  Options:
  Post Exec Rule:
  Pre Exec Rule:
  Reclaim Policy:  Retain
  Resource Types:
    ConfigMap
    PersistentVolumeClaim
  Selectors:            <nil>
  Skip Service Update:  false
```

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -1088,7 +1088,6 @@ func (a *ApplicationBackupController) backupResources(
 				incResNsBatch = batch
 			}
 		}
-
 		if len(incResNsBatch) != 0 {
 			objects, err := a.resourceCollector.GetResources(
 				incResNsBatch,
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1141,24 +1141,10 @@ func (a *ApplicationBackupController) backupResources(
 			}
 		}
 	}
-	updatedAllObjects := make([]runtime.Unstructured, 0)
```
ram-infrac, line 1144: Dont need this anymore , since `GetResources` and `GetResourceTypes` delete all metadata of objects UID will always be empty

prashanthpx (approved):
looks good

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1106,7 +1105,8 @@ func (a *ApplicationBackupController) backupResources(
 		if len(resourceTypeNsBatch) != 0 {
 			for _, backupResourceType := range backup.Spec.ResourceTypes {
 				for _, resource := range resourceTypes {
-					if resource.Kind == backupResourceType {
+					if resource.Kind == backupResourceType || (backupResourceType == "PersistentVolumeClaim" && resource.Kind == "PersistentVolume") {
+						log.ApplicationBackupLog(backup).Debugf("GetResourcesType for : %v", resource.Kind)
```
adityadani, line 1109: This log line will be very chatty. Lets make it `Tracef`

adityadani (approved):

PR #874, opened 2021-08-03
Stor 389: migration recreate svc only if its been updated on source cluster

ram-infrac:
**What type of PR is this?**
> enhancement

Recreate service only if its been changed on primary/source cluster

no

yes
```release-note
Stork will not recreate service on destination/DR cluster if it's not changed on primary cluster during migration
```

yes, 2.6.4-*

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -16,6 +16,7 @@ import (
 	"github.com/libopenstorage/stork/pkg/log"
 	"github.com/libopenstorage/stork/pkg/resourcecollector"
 	"github.com/libopenstorage/stork/pkg/rule"
+	"github.com/mitchellh/hashstructure"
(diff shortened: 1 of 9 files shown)


prashanthpx (approved):
LGTM

pkg/migration/controllers/migration.go
```diff
@@ -982,6 +983,59 @@ func (m *MigrationController) getRemoteAdminConfig(migration *stork_api.Migratio
...
+			if err != nil {
+				return false, err
+			}
+			if old == objHash {
+				log.MigrationLog(migration).Infof("Skipping service update, no changes found since last migration %d/%d", old, objHash)
+				return true, nil
+			}
+		}
+	}
+	if _, ok := svc.Annotations[resourcecollector.SkipModifyResources]; ok {
```
adityadani, line 1017: shouldn't this check be before the curr.Annotations check ?

PR #878, opened 2021-08-16
Do not force delete pods in VolumeSnapshotRestore.

adityadani:

**What type of PR is this?**
>bug

- Force deleting the pods immediately deletes the pod objects from k8s etcd,
which gives a false indication to stork that the pods have been deleted. The
subsequent restore call fails since the underlying PVC is not yet detached by k8s.

No

Yes
```release-note
Issue: In-place VolumeSnapshotRestore issued a "force" delete of the pods using the PVC that needs to be restored. The "force" delete of the pods gave a false indication that the PVC is not being used by any apps. The subsequent Restore command used to fail since the volume was still in use.
User Impact: In-place VolumeSnapshotRestore would fail intermittently.
Resolution: Perform a regular delete of the pods and wait for pod deletion while performing in-place VolumeSnapshotRestore.
```

Yes - 2.6

### DIFF
--- a/pkg/snapshot/controllers/snapshotrestore.go
+++ b/pkg/snapshot/controllers/snapshotrestore.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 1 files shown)


ram-infrac (approved):
nit in release note:
we are still doing force delete , but waiting on pods to get cleaned up by k8s

PR #883, opened 2021-08-25
Added migration failover and failback tests.

adityadani:
**What type of PR is this?**
>integration-test

- The existing migration tests setup one way cluster pairs between two clusters
  however the migrations also happen only in one direction. Once the migration
  is complete the resources are deleted and a reverse migration is triggered.
- This does not execute the scenario of failover and failback of the same
  application and volume between the two clusters.
- Currently the test is only added for  portworx driver. Using portworx encryted volumes with mysql statefulsets.

No

No

Yes - 2.6


test/integration_test/migration_failover_failback_test.go
```diff
@@ -0,0 +1,174 @@
...
+	err = schedulerDriver.Destroy(preMigrationCtx, nil)
+	require.NoError(t, err, "Error destroying ctx: %+v", preMigrationCtx)
+	err = schedulerDriver.WaitForDestroy(preMigrationCtx, defaultWaitTimeout)
+	require.NoError(t, err, "Error waiting for destroy of ctx: %+v", preMigrationCtx)
+
+	// ensure app starts on cluster 1
+	err = setSourceKubeConfig()
+	require.NoError(t, err, "Error resetting remote config")
+
+	err = schedulerDriver.WaitForRunning(postMigrationCtx, defaultWaitTimeout, defaultWaitInterval)
```
ram-infrac, line 170: you will need to start app on source cluster , since `appstart` in migration is set to false

ram-infrac (approved):

PR #890, opened 2021-09-08
stor-453: Allow support for kdmp controlled generic backup 

ram-infrac:
**What type of PR is this?**
>feature


This PR adds support for KDMP controlled generic backups - 
- Added KDMP volume driver support for backups
- Handle generic backup types
- Start KDMP controller for handling DataExport and VolumeBackup CRs 

no 

yes

2.7


Tested with 3 px pvcs - 
bkp.yaml :- 
```
kind: ApplicationBackup
metadata:
  name: mysql-test-bkp-1
spec:
  backupLocation: kopia-bl
  backupType: generic
  # namespaces to be backed up
  namespaces:
  - tt
  # What to do with the data in the objectstore when the backup object in k8s is deleted. Valid options are Delete and Retain
  reclaimPolicy: Delete
  # List of label selectors to choose specific objects and volumes to backup
  selectors:
  # Rule to run before triggering the backup
  preExecRule:
  # Rule to run after the backup has been triggered
  postExecRule:
```

```
Every 2.0s: storkctl get applicationbackups -ntt                                                                                                                                                                                                        Thu Sep  9 11:23:56 2021

NAME               STAGE   STATUS       VOLUMES   RESOURCES   CREATED               ELAPSED
mysql-test-bkp-1   Final   Successful   3/3       11          09 Sep 21 11:22 UTC   1m3s
```

![Screenshot 2021-09-09 at 4 55 09 PM](https://user-images.githubusercontent.com/35250819/132677521-8b9fbdfa-8f91-43b2-9f92-f0f23a5af3ab.png)

**Note** : bucket name is `kdmp-test`


drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+	kdmpShedOps "github.com/portworx/sched-ops/k8s/kdmp"
+	"github.com/portworx/sched-ops/k8s/storage"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8shelper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+)
+
+const (
+	// driverName is the name of the aws driver implementation
```
prashanthpx, line 27: nit: Should this be kdmp instead of aws?

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (k *kdmp) OwnsPVC(coreOps core.Ops, pvc *v1.PersistentVolumeClaim) bool {
+	// Try to get info from the PV since storage class could be deleted
+	pv, err := coreOps.GetPersistentVolume(pvc.Spec.VolumeName)
+	if err != nil {
+		log.PVCLog(pvc).Warnf("error getting pv %v for pvc %v: %v", pvc.Spec.VolumeName, pvc.Name, err)
+		return false
+	}
+	// for px proxy volume use kdmp by default
+	storageClassName := k8shelper.GetPersistentVolumeClaimClass(pvc)
```
prashanthpx, line 64: Would there be any case wherein `storageClassName` is empty and if so, do we need to handle this?

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	return k.OwnsPV(pv)
+}
+
+func (k *kdmp) OwnsPV(pv *v1.PersistentVolume) bool {
+
+	// for csi volumes (with or without snapshot) use kdmp driver by default
+	// except for px csi volumes
+	if pv.Spec.CSI != nil {
+		// We px csi volume fall back to default csi volume driver
```
prashanthpx, line 82: nit: Remove the word `We`

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+		return true
+	}
+
+	return false
+}
+
+func (k *kdmp) StartBackup(backup *storkapi.ApplicationBackup,
+	pvcs []v1.PersistentVolumeClaim,
+) ([]*storkapi.ApplicationBackupVolumeInfo, error) {
+	log.ApplicationBackupLog(backup).Debugf("started kdmp backup: %v", backup.Name)
```
prashanthpx, line 97: Would be good to call generic instead of kdmp in the message

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+		volumeInfo.DriverName = driverName
+		volume, err := core.Instance().GetVolumeForPersistentVolumeClaim(&pvc)
+		if err != nil {
+			return nil, fmt.Errorf("error getting volume for PVC: %v", err)
+		}
+		volumeInfo.Volume = volume
+		// create kdmp cr
+
+		dataExport := &kdmpapi.DataExport{}
+		dataExport.Name = "stork-" + pvc.Name + "-" + pvc.Namespace
```
prashanthpx, line 116: Was thinking instead can we the backup CR name here instead of stork that way we can actually map the backup name (given in UI) directly to the DataExport CR. Even in the pod creation, we use the DE CR name that way we have end-to-end mapping.

If we have one more backup triggered on the same PVC when the first one is in progress, it gets difficult to know which backup points to which CR.

Also good to have pvc.Namespace + pvc.Name so that we have the same uniformity referencing to the backup path (generic-backup/<ns>-<pvc name>

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+		dataExport.Namespace = pvc.Namespace
+		// TODO: this needs to be generic, maybe read it from configmap
```
prashanthpx, line 118: I didn't understand the command. What are we planning to store in configMap?

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		dataExport.Spec.Source = kdmpapi.DataExportObjectReference{
+			Kind:       pvc.Kind,
+			Name:       pvc.Name,
+			Namespace:  pvc.Namespace,
+			APIVersion: pvc.APIVersion,
+		}
+		logrus.Debugf("Source for KDMP: %v", dataExport.Spec.Source)
+		logrus.Debugf("Destination for KDMP: %v", dataExport.Spec.Destination)
+		// TODO: do we need retries here
```
prashanthpx, line 134: With the change to job framework model in kdmp limiting the number of jobs, do we still need this? Jobs will not even start if the threshold has not been reached.

Nit: typo `excedded`

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+		// in case no of CRs excedded limit handled by kdmp controller
+		_, err = kdmpShedOps.Instance().CreateDataExport(dataExport)
+		if err != nil {
+			logrus.Errorf("failed to create job: %v", err)
```
prashanthpx, line 138: nit: Failed to create DataExport CR and not job. Even in other instance we are referencing to job, can be changed to DE

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,300 @@
...
+				vInfo.Status = storkapi.ApplicationBackupStatusFailed
+				vInfo.Reason = fmt.Sprintf("Backup failed for volume: %v", dataExport.Status.Reason)
+			} else {
+				vInfo.Status = storkapi.ApplicationBackupStatusSuccessful
+				vInfo.Reason = "Backup successful for volume"
+				// delete kdmp crs
+				if err := kdmpShedOps.Instance().DeleteDataExport("stork-"+vInfo.PersistentVolumeClaim+"-"+vInfo.Namespace, backup.Namespace); err != nil {
+					logrus.Warnf("failed to delete data export CR: %v", err)
+				}
+				// TODO: KDMP will take car of removing this
```
prashanthpx, line 186: nit: typo `car`

prashanthpx (approved):
Looks good to me after fixing the comments.

drivers/volume/kdmp/kdmp.go
ram-infrac: yes it can be , in that case it will be handled by looking at pv object

drivers/volume/kdmp/kdmp.go
ram-infrac: good idea, changed to `backup.name-namespace-pvcname`

drivers/volume/kdmp/kdmp.go
ram-infrac: talking about case where we change underlying KDMP driver by reading it from configmap

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,297 @@
...
+			volumeBackup, err := kdmpShedOps.Instance().GetVolumeBackup(vInfo.Volume, backup.Namespace)
+			if err != nil {
+				logrus.Errorf("failed to get volumebackup CR: %s, err:%v", vInfo.Volume, err)
+				return volumeInfos, err
+			}
+			vInfo.BackupID = volumeBackup.Status.SnapshotID
+			if isBackupActive(dataExport.Status) {
+				vInfo.Status = storkapi.ApplicationBackupStatusInProgress
+				vInfo.Reason = fmt.Sprintf("Volume backup in progress. BytesDone: %v",
+					volumeBackup.Status.ProgressPercentage)
```
adityadani, line 170: progress percentage is not BytesDone. Since we don't have the actual progress percentage nor the bytes that have been transferred, lets not add it to the Reason.

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,297 @@
...
+				// TODO: KDMP does not return size right now, we will need to fill up
+				// size for volume once support is available
+			} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+				vInfo.Status = storkapi.ApplicationBackupStatusFailed
+				vInfo.Reason = fmt.Sprintf("Backup failed for volume: %v", dataExport.Status.Reason)
+			} else {
```
adityadani, line 176: I would make this as `else if` as well, so that if in future we add any new state (which is not active) it would not fallback into this else.

adityadani (approved):

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,297 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8shelper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+)
+
+const (
+	// driverName is the name of the kdmp driver implementation
+	driverName    = "kdmp"
+	proxyEndpoint = "proxy_endpoint"
+	proxyPath     = "proxy_nfs_exportpath"
```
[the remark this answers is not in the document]
ram-infrac: `proxyEndpoint` is needed to figure out PX proxy volumes

drivers/volume/kdmp/kdmp.go
```diff
@@ -0,0 +1,297 @@
...
+			logrus.Warnf("Error getting storageclass %v for pvc %v: %v", storageClassName, pvc.Name, err)
+		}
+		if _, ok := storageClass.Parameters[proxyEndpoint]; ok {
+			return true
+		}
+	}
+	return k.OwnsPV(pv)
+}
+
+func (k *kdmp) OwnsPV(pv *v1.PersistentVolume) bool {
```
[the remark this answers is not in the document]
ram-infrac: correct, we will need to change restore flow as well, planning to add it as part of generic restore changes

PR #891, opened 2021-09-09
Add a Snapshotter interface

adityadani:
**What type of PR is this?*
>cleanup

- The snapshotter interface provides APIs for snapshots and restores.
- Extracted out the snapshotting functionality from the CSI driver and added it
  to the snapshotter interface

The goal is to separate out the snapshot functionality from the actual drivers so
that different components in stork can use it.

No

No

No


pkg/snapshotter/snapshotter_csi.go
```diff
@@ -0,0 +1,464 @@
...
+	if contentName != "" {
+		vscError = c.getSnapshotContentError(contentName)
+	}
+	switch {
+	case volumeSnapshotReady && volumeSnapshotContentReady:
+		snapshotInfo.Status = StatusReady
+		snapshotInfo.Reason = "Snapshot successful for volume"
+		snapshotInfo.Size = uint64(size)
+
+	case time.Now().After(snapshot.CreationTimestamp.Add(snapshotTimeout)):
```
ram-infrac, line 220: should we make `snapshotTimeout`  configurable by reading it as env variable, by defaults its 5 mins

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -0,0 +1,464 @@
...
+
+func (c *csiDriver) snapshotContentReady(vscontent *kSnapshotv1beta1.VolumeSnapshotContent) bool {
+	return vscontent.Status != nil && vscontent.Status.ReadyToUse != nil && *vscontent.Status.ReadyToUse
+}
+
+func formatReasonErrorMessage(reason, vsError, vscError string) string {
+	var snapshotError string
+
+	switch {
+	case vsError != "" && vscError != "" && vsError != vscError:
```
ram-infrac, line 403: nice

ram-infrac (approved):
lgtm

PR #895, opened 2021-09-19
[Portworx Driver] Add support for PX-Security enabled clusters.

adityadani:
- Add support for backing and restoring PVCs which are provisioned
  on a PX-Security enabled cluster.

Backup:
  - CSI Volumes: Use the token secret name/namespace specified
    as a part of NodePublish secrets
  - In-tree Volumes: Use the token secret name/namespace provided as annotations
  - Save the secret name and namespaces as a part of backup.VolumeInfo.Options
    so that they can be used while restoring.
  - If a templatized namespace is provided, then store ${pvc.namespace} as a part
    of backup.VolumeInfo.Options so that on restore the token secret in the namespace where
    restore is being done is used.

Restore:
  - Use the token secret name and namespace provided as a part of backup options.
  - If a templatized namespace is found use the secret in the namespace where the
    PVC is being restored.

**What type of PR is this?**
>bug
>feature
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

```release-note

```


ram-infrac (approved):
lgtm

drivers/volume/portworx/portworx.go
```diff
@@ -804,28 +828,104 @@ func (p *portworx) secretRefFromAnnotations(annotations map[string]string) *v1.S
...
+		// If the secret ref namespace is same as the PVC namespace check if
+		// templatized namespace was provided
+		if pv.Spec.CSI.NodePublishSecretRef.Namespace == pvc.Namespace {
+			sc, err := core.Instance().GetStorageClassForPVC(pvc)
+			if err != nil {
+				logrus.Warnf("Failed to fetch storage class for PVC %v/%v: %v", pvc.Name, pvc.Namespace, err)
+				// Best effort - return what we have
+				return pv.Spec.CSI.NodePublishSecretRef.Name, pv.Spec.CSI.NodePublishSecretRef.Namespace
+			}
+			if val, ok := sc.Parameters["csi.storage.k8s.io/node-publish-secret-namespace"]; ok && val == templatizedNamespace {
```
ggriffiths, line 881: Not sure if we want to add this, but CSI also supports the `pvc.Name` template:
https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#createdelete-volume-secret

Users of CSI might expect `{{pvc.Name`}}` as a possibility. If this isn't the case, we can just document accordingly.

ggriffiths (approved):
Looks good. Have one question/doc item

drivers/volume/portworx/portworx.go
ggriffiths: Sounds good

PR #901, opened 2021-09-22
csidriverfix: handle completed backups

ram-infrac:
**What type of PR is this?**
>bug

Currently CSI driver uploads all of snapshot CR to Objectstore and delete once done. This causes issue when other non px backup is still in progress & all CSI volume backup is completed, CSI driver will still look at snapshot which is deleted previously and cause application backup to fail. As separate clean up task we should introduce API for driver to clean-up post backup operation, since handling everything in `GetBackupStatus` is prone to error

no

no

2.7

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -604,6 +604,23 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 	vsMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshot)
 	vsContentMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshotContent)
 	vsClassMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshotClass)
+
+	// check if all csi vol backup is successful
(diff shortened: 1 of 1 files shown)


drivers/volume/csi/csi.go
```diff
@@ -604,6 +604,22 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
...
+
+	// check if all csi vol backup is successful
+	isCompleted := true
+	currVolInfo := make([]*storkapi.ApplicationBackupVolumeInfo, 0)
+	for _, vInfo := range backup.Status.Volumes {
+		if vInfo.DriverName != storkCSIDriverName {
+			continue
+		}
+		if vInfo.Status != storkapi.ApplicationBackupStatusSuccessful {
+			isCompleted = false
```
adityadani, line 616: you can break the for loop here.

adityadani (approved):

drivers/volume/csi/csi.go
ram-infrac: correct, thanks

PR #904, opened 2021-09-24
pb-1908: Added backup and restore CR name and pvc name in the dataexport CR.

siva-portworx:
**What type of PR is this?**
bug.

pb-1908

no

no

2.7?

Note: There is a dependent PR in kdmp where these labels will be used. 
https://github.com/portworx/kdmp/pull/62

### DIFF
--- a/drivers/volume/kdmp/kdmp.go
+++ b/drivers/volume/kdmp/kdmp.go
@@ -51,6 +51,9 @@ const (
 	secretNamespace         = "kube-system"
 	pxbackupAnnotation      = "portworx.io/created-by"
 	pxbackupAnnotationValue = "px-backup"
+	backupCRNameKey         = "kdmp.portworx.com/backup-cr-name"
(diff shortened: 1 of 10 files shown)


drivers/volume/kdmp/kdmp.go
```diff
@@ -471,6 +489,10 @@ func (k *kdmp) StartRestore(
 
 		// create kdmp cr
 		dataExport := &kdmpapi.DataExport{}
+		labels := make(map[string]string)
+		labels[backupCRNameKey] = restore.Name
```
ram-infrac, line 493: need to do same for restore as well ?

ram-infrac (approved):

drivers/volume/kdmp/kdmp.go
siva-portworx: Fixed it.

PR #911, opened 2021-09-28
Add CleanupBackup/Restore object volume driver api

ram-infrac:
**What type of PR is this?**
>bug
>cleanup

no

not yet decided

**Note** : 
- tested single ns backup and restore by deleting resources
- multiple-namespace backup and restore to different namespaces

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -566,23 +566,6 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 	vsMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshot)
 	vsContentMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshotContent)
 	vsClassMap := make(map[string]*kSnapshotv1beta1.VolumeSnapshotClass)
-
-	// check if all csi vol backup is successful
-	isCompleted := true
-	currVolInfo := make([]*storkapi.ApplicationBackupVolumeInfo, 0)
-	for _, vInfo := range backup.Status.Volumes {
-		if vInfo.DriverName != storkCSIDriverName {
-			continue
-		}
-		if vInfo.Status != storkapi.ApplicationBackupStatusSuccessful {
-			isCompleted = false
-			break
-		}
-		currVolInfo = append(currVolInfo, vInfo)
-	}
-	if isCompleted {
-		return currVolInfo, nil
-	}
 	for _, vInfo := range backup.Status.Volumes {
 		if vInfo.DriverName != storkCSIDriverName {
 			continue
@@ -670,13 +653,6 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 			return nil, err
 		}
 		log.ApplicationBackupLog(backup).Debugf("finished and uploaded %v snapshots and %v snapshotcontents", len(vsMap), len(vsContentMap))
-
-		// cleanup after a successful object upload
-		err = c.cleanupSnapshots(vsMap, vsContentMap, true)
-		if err != nil {
-			return nil, fmt.Errorf("failed to cleanup snapshots: %v", err)
-		}
-		log.ApplicationBackupLog(backup).Debugf("started clean up of %v snapshots and %v snapshotcontents", len(vsMap), len(vsContentMap))
 	}
 
 	return volumeInfos, nil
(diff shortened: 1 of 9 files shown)


drivers/volume/volume.go
```diff
@@ -152,6 +154,8 @@ type BackupRestorePluginInterface interface {
 	GetRestoreStatus(*storkapi.ApplicationRestore) ([]*storkapi.ApplicationRestoreVolumeInfo, error)
 	// Cancel the restore of volumes specified in the status
 	CancelRestore(*storkapi.ApplicationRestore) error
+	// CleanupBackupResources the backup of resource specified backup
```
siva-portworx, line 157: Typo: comment should be corrected for restore.

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -1280,3 +1301,18 @@ func (a *ApplicationRestoreController) createCRD() error {
...
+func (a *ApplicationRestoreController) cleanupResources(restore *storkapi.ApplicationRestore) error {
+	drivers := a.getDriversForRestore(restore)
+	for driverName := range drivers {
+
+		driver, err := volume.Get(driverName)
+		if err != nil {
+			return err
+		}
+		if err := driver.CleanupRestoreResources(restore); err != nil {
+			logrus.Errorf("unable to cleanup post backup resources, err: %v", err)
```
siva-portworx, line 1314: typo: backup to restore

siva-portworx (approved):
LGTM

drivers/volume/csi/csi.go
```diff
@@ -670,13 +653,6 @@ func (c *csi) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.A
 			return nil, err
 		}
 		log.ApplicationBackupLog(backup).Debugf("finished and uploaded %v snapshots and %v snapshotcontents", len(vsMap), len(vsContentMap))
-
```
adityadani, line 673: With this new change we are only cleaning up resources if a backup fails. But in case of CSI we clean up event after a successful backup. Same for snapshotsForRestore

adityadani (changes requested):

drivers/volume/kdmp/kdmp.go
```diff
@@ -492,15 +491,15 @@ func (k *kdmp) StartRestore(
...
 			Name:       restore.Spec.BackupLocation,
-			Namespace:  restore.Namespace,
+			Namespace:  restore.Namespace, // since this can be kube-system in case of multple namespace restore
 			APIVersion: StorkAPIVersion,
 		}
-		volBackup.Spec.Repository = fmt.Sprintf("%s/%s-%s/", prefixRepo, bkpvInfo.Namespace, bkpvInfo.PersistentVolumeClaim)
+		volBackup.Spec.Repository = fmt.Sprintf("%s/%s-%s/", prefixRepo, volumeInfo.SourceNamespace, bkpvInfo.PersistentVolumeClaim)
 		volBackup.Status.SnapshotID = bkpvInfo.BackupID
 		if _, err := kdmpShedOps.Instance().CreateVolumeBackup(volBackup); err != nil {
 			logrus.Errorf("unable to create backup cr: %v", err)
```
prashanthpx, line 505: nit:  `backup cr` -> `volumeBackupCR`

drivers/volume/kdmp/kdmp.go
```diff
@@ -523,8 +522,8 @@ func (k *kdmp) StartRestore(
 		dataExport.Labels = labels
```
prashanthpx, line 522: Not related to this line but for this, can we trim of `-` is its the last char got after truncating? This observation from the issue we saw while testing when names were lengthier and restore was failing.
```
labels[restoreCRNameKey] = truncate.Truncate(restore.Name, labelNamelimit, "", truncate.PositionEnd)
```

drivers/volume/kdmp/kdmp.go
```diff
@@ -550,11 +549,20 @@ func (k *kdmp) StartRestore(
...
+
+		}
+		restoreNamespace := val
+		crName := getGenericCRName(restore.Name, restoreNamespace, vInfo.PersistentVolumeClaim)
+		if err := kdmpShedOps.Instance().DeleteDataExport(crName, restoreNamespace); err != nil {
 			logrus.Errorf("failed to delete data export CR: %v", err)
 			return err
 		}
+		if err := kdmpShedOps.Instance().DeleteVolumeBackup(prefixRestore+"-"+crName, restoreNamespace); err != nil {
+			logrus.Tracef("failed to delete volume backup CR:%s/%s, err: %v", restoreNamespace, crName, err)
```
prashanthpx, line 564: Don't want to return failure from here?

prashanthpx (approved):
lgtm

drivers/volume/kdmp/kdmp.go
ram-infrac: yes, since this is will be final stage where backup/restore is already completed. If we return err, we get immediate reconciler event

drivers/volume/kdmp/kdmp.go
ram-infrac: got it, added it as separate commit

PR #913, opened 2021-10-01
Portworx: Parse the storage class mapping while restoring a volume.

adityadani:
**What type of PR is this?**
>feature

This change allows backing up a Portworx volume of a particular storage class
and properties, and restore that volume with a different storage class.
For ex. Backup a volume with HALevel=2 and restore it with HALevel=1

- Parse the storage class mapping from restore spec and find if a mapping
  for the current PVC's storage class is found.
- If found, fetch the actual contents of storage class and parse them into
  a RestoreVolumeSpec.
- Invoke the CloudBackupRestore API with this RestoreVolumeSpec.

No

```release-note
Add support for backing up a Portworx volume with a particular storage class
and properties, and restore it with a different storage class.
```

No


ram-infrac (approved):
lgtm

PR #914, opened 2021-10-02
pb-1941,pb-1944: modified the generic backup resource naming convention.

siva-portworx:
**What type of PR is this?**
bug

```
    pb-1941,pb-1944: modified the generic backup resource name to be match
    new naming format as given below

        - name format <backup>-<shortUID of backup CR>-<shortID of
              PVC>-<NS>
        - Calling cleanup function in Final case is creating issue as
          px-backup will delete the CR as soon the stork updates the CR
          status to success or failure and stage as Final. Because of that
          application backup CR is getting deleted before dataexport CR.
        - Now calling cleanup before moving the status of final stage
          and success.
```

no

no

No

**Note:
This is regarding the pb-1944:**
1) Currently we are calling the cleanupResources in Final stage. But when we set the stage Final and status to Success, the px-backup detects that deletes the applicationbackup CR. 
2) Because of this the cleanupResources as Final stage is not executed and the dataexport are not deleted.
3) For now calling the cleanupResources api before setting the stage to Final and status as success.
4) One issue is that in the case of failure status and Final stage, cleanupResources will not be called. I hope it is ok as we were planning to retain the CR. If not we need to fix it.

**Testing results are attached to kdmp PR, which is related to this change:**
https://github.com/portworx/kdmp/pull/71


drivers/volume/kdmp/kdmp.go
```diff
@@ -296,13 +320,26 @@ func deleteKdmpSnapshot(backup *storkapi.ApplicationBackup) (bool, error) {
...
+				labels[pvcNameKey] = getValidLabel(vInfo.PersistentVolumeClaim)
+				labels[pvcUIDKey] = getValidLabel(vInfo.PersistentVolumeClaimUID)
+				// If backup from px-backup, update the backup object details in the label
+				if val, ok := backup.Annotations[pxbackupAnnotationCreateByKey]; ok {
+					if val == pxbackupAnnotationCreateByValue {
+						labels[backupObjectNameKey] = getValidLabel(backup.Annotations[pxbackupObjectNameKey])
+						labels[backupObjectUIDKey] = getValidLabel(backup.Annotations[pxbackupObjectUIDKey])
+					}
+				}
+				err := dataexport.CreateCredentialsSecret(secretName, backup.Spec.BackupLocation, backup.Namespace, backup.Namespace, labels)
```
prashanthpx, line 342: Should the secret name also follow the new naming convention coz here it just uses `backup.Name` which is the CR name?

drivers/volume/kdmp/kdmp.go
```diff
@@ -511,23 +549,23 @@ func (k *kdmp) StartRestore(
...
-			crLabelname := truncate.Truncate(restore.Name, labelNamelimit, "", truncate.PositionEnd)
-			labels[restoreCRNameKey] = strings.Trim(crLabelname, "-")
-		}
-		if len(bkpvInfo.PersistentVolumeClaim) <= labelNamelimit {
-			labels[pvcNameKey] = bkpvInfo.PersistentVolumeClaim
-		} else {
-			crLabelName := truncate.Truncate(bkpvInfo.PersistentVolumeClaim, labelNamelimit, "", truncate.PositionEnd)
-			labels[pvcNameKey] = strings.Trim(crLabelName, "-")
+		labels[applicationRestoreCRNameKey] = getValidLabel(restore.Name)
+		labels[applicationRestoreCRUIDKey] = getValidLabel(string(restore.UID))
```
prashanthpx, line 555: Can we add similar kinds of labels for `VolumeBackupCR` would be helpful when debugging for filteration?

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -1171,15 +1172,19 @@ func (a *ApplicationRestoreController) restoreResources(
...
+	err = a.cleanupResources(restore)
+	if err != nil {
+		return err
+	}
 	restore.Status.Stage = storkapi.ApplicationRestoreStageFinal
 	restore.Status.FinishTimestamp = metav1.Now()
 	restore.Status.Status = storkapi.ApplicationRestoreStatusSuccessful
 	restore.Status.Reason = "Volumes and resources were restored up successfully"
 	for _, resource := range restore.Status.Resources {
 		if resource.Status != storkapi.ApplicationRestoreStatusSuccessful {
```
prashanthpx, line 1185: If the overall status is Failure, wouldn't the reason say success with this message?

prashanthpx (approved):
lgtm after addressing review comments.

drivers/volume/kdmp/kdmp.go
siva-portworx: Fixed

drivers/volume/kdmp/kdmp.go
siva-portworx: Fixed

drivers/volume/kdmp/kdmp.go
```diff
@@ -709,3 +747,19 @@ func init() {
 		logrus.Panicf("Error registering kdmp volume driver: %v", err)
 	}
 }
+
+// getValidLabel - will validate the label to make sure the length is less 63 and contains valid label format.
+// If the length is greater then 63, it will truncate to 63 character.
+func getValidLabel(labelVal string) string {
+	if len(labelVal) > validation.LabelValueMaxLength {
+		labelVal = truncate.Truncate(labelVal, validation.LabelValueMaxLength, "", truncate.PositionEnd)
+		// make sure the truncated value does not end with the hypen.
```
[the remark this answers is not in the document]
siva-portworx: Fixed

drivers/volume/kdmp/kdmp.go
```diff
@@ -155,24 +179,24 @@ func (k *kdmp) StartBackup(backup *storkapi.ApplicationBackup,
...
-		} else {
-			crLabelName := truncate.Truncate(pvc.Name, labelNamelimit, "", truncate.PositionEnd)
-			labels[pvcNameKey] = strings.Trim(crLabelName, "-")
+		labels[applicationBackupCRNameKey] = getValidLabel(backup.Name)
+		labels[applicationBackupCRUIDKey] = getValidLabel(string(backup.UID))
+		labels[pvcNameKey] = getValidLabel(pvc.Name)
+		labels[pvcUIDKey] = getValidLabel(string(pvc.UID))
+		// If backup from px-backup, update the backup object details in the label
+		if val, ok := backup.Annotations[pxbackupAnnotationCreateByKey]; ok {
+			if val == pxbackupAnnotationCreateByValue {
```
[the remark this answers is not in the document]
siva-portworx: As discussed, this the labels for the dataExport CR. So can not be added in px-backup code.

PR #915, opened 2021-10-04
Portworx: Check for auth params in in-tree storage class definition as well

adityadani:
**What type of PR is this?**
>feature

Portworx supports providing auth tokens as a part of storage classes defined for using the in-tree plugin.
Currently stork was only looking at Storage Classes defined for Portworx's CSI plugin. 

No

No - parent issue does have a release note

Yes - 2.7


ram-infrac (approved):
lgtm

PR #917, opened 2021-10-05
STOR-521: Fixed generic backup failure

prashanthpx:
**What type of PR is this?**
Bug

- Currently, the VolumeBackup CR is created as part of the job pod.
  There could be a scenario where this CR creation is taking time.
- The moment DE CR is submitted for Backup, as part of DataExportStageTransferInProgress stage VB CR is queried.
  In this scenario, CR is still not available and backup is marked as a failure.
- The fix is to mark the backup as failed only when we are in DataExportStageFinal stage

No

No

Yes 2.8


drivers/volume/kdmp/kdmp.go
```diff
@@ -255,6 +248,16 @@ func (k *kdmp) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.
 				vInfo.Reason = "Volume backup in progress"
 				// TODO: KDMP does not return size right now, we will need to fill up
 				// size for volume once support is available
+			} else if dataExport.Status.Status == kdmpapi.DataExportStatusFailed {
+				// As DE CR is in DataExportStageTransferInProgress we check VB CR status
+				// and as this is created by job, there could be a race where this CR
+				// is not yet available as the jon pod is still creating it
```
siva-portworx, line 254: type jon

siva-portworx (approved):

PR #918, opened 2021-10-05
stor-520 : Nil handle while doing application restore of namespace 

ram-infrac:
**What type of PR is this?**
>bug

Stork-520 : Handle empty label while doing application-restore
Correct err msg for crd activate/deactivate

no

no

2.7.0


prashanthpx (approved):
lgmt

PR #935, opened 2021-10-21
Reading backup type from config map

prashanthpx:
**What type of PR is this?**
Big

- User can set all backups to be generic by adding BACKUP_TYPE=generic in kdmp-config map
- This way all backups would be forced to be generic

No

No

No


drivers/drivers.go
```diff
@@ -1,5 +1,12 @@
 package drivers
 
+const (
+	// KdmpConfigmapName kdmp config map name
+	KdmpConfigmapName = "kdmp-config"
+	// KdmpConfigmapNamespace kdmp config map ns
+	KdmpConfigmapNamespace = "kube-system"
```
siva-portworx, line 7: It is better to defaultAdminNamespace value as it is used during configmap creation.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -516,7 +518,11 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 			if err != nil {
 				return fmt.Errorf("error getting list of volumes to backup: %v", err)
 			}
-
+			kdmpData, err := core.Instance().GetConfigMap(drivers.KdmpConfigmapName, drivers.KdmpConfigmapNamespace)
+			if err != nil {
+				return fmt.Errorf("error readig kdmp config map: %v", err)
```
siva-portworx, line 523: I think it is better to log the message and continue with the default option for the backup rt?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -542,7 +548,8 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 					continue
 				}
 				driverName := stork_api.GenericDriver
-				if backup.Spec.BackupType == stork_api.ApplicationBackupGeneric {
+				if backup.Spec.BackupType == stork_api.ApplicationBackupGeneric ||
+					driverType == stork_api.ApplicationBackupGeneric {
```
siva-portworx, line 552: There is also one place, where we are checking for the genericBackup. 
https://github.com/libopenstorage/stork/blob/master/pkg/applicationmanager/controllers/applicationbackup.go#L590

siva-portworx (approved):

PR #949, opened 2021-11-01
   pb-1988: Added alreadyExists check for creation of volumeSnapshot and     PVC creation

siva-portworx:
**What type of PR is this?**
bug

```
    pb-1988: Added alreadyExists check for creation of volumeSnapshot and
    PVC creation

        - While deleting the snapshot CRs, setting the proper retain
          value. For latest snapshot, retain will be set true, so that local
              snapshot will be retained. For other snapshots, it will be set
              false, so that both the CR and local snapshot files are also
              will be deleted

```

no

no

2.8

**Related kdmp PR:**
https://github.com/portworx/kdmp/pull/96

### DIFF
--- a/pkg/snapshotter/snapshotter_csi.go
+++ b/pkg/snapshotter/snapshotter_csi.go
@@ -121,6 +121,9 @@ func (c *csiDriver) CreateSnapshot(opts ...Option) (string, string, string, erro
 		vs,
 		metav1.CreateOptions{},
 	); err != nil {
+		if k8s_errors.IsAlreadyExists(err) {
+			return o.Name, o.PVCNamespace, pv.Spec.CSI.Driver, nil
+		}
 		return "", "", "", err
 	}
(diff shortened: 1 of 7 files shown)


pkg/snapshotter/snapshotter_csi.go
```diff
@@ -785,7 +791,7 @@ func (c *csiDriver) getCSISnapshotsCRListForDelete(backupLocation *storkapi.Back
 
 	sort.Strings(timestamps)
 	if len(timestamps) > localCSIRetention {
```
diptiranjanpx, line 793: Let's remove this check from here.

diptiranjanpx (approved):

pkg/snapshotter/snapshotter_csi.go
siva-portworx: Fixed

PR #961, opened 2021-11-11
STOR-528: CSI VolumeSnapshotClass fixes

adityadani:
**What type of PR is this?**
>improvement

- If the SnapshotClassName is set as "default" or "Default"
let Kubernetes handle the snapshot class selection. It wil use
the default snapshot class set by the admin for that CSI provider.

- Do not create stork-* volume snapshot class on startup if a snapshot
class for the CSI driver already exists.

No

Yes
```release-note
STORK will not create a VolumeSnapshot class if there already exists one for the CSI provider.
```

Yes - 2.8

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -137,14 +137,9 @@ func (c *csi) Init(_ interface{}) error {
 	}
 	c.snapshotClient = cs
 
-	if c.isCSIInstalled() {
-		logrus.Infof("Creating default CSI SnapshotClasses")
-		err = c.createDefaultSnapshotClasses()
-		if err != nil {
-			return err
-		}
-	} else {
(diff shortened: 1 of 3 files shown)


drivers/volume/csi/csi.go
```diff
@@ -186,6 +182,16 @@ func (c *csi) createDefaultSnapshotClasses() error {
 			continue
 		}
 
+		foundSnapClass := false
+		for _, existingSnapClass := range existingSnapshotClasses.Items {
+			if driver.Name == existingSnapClass.Driver {
+				logrus.Infof("CSI VolumeSnapshotClass exists for driver %v. Skipping creation of snapshotclass", driver.Name)
+				foundSnapClass = true
```
diptiranjanpx, line 189: can put a break here.

diptiranjanpx (approved):

drivers/volume/csi/csi.go
adityadani: done

PR #986, opened 2021-12-04
pb-2098: Added google file and azure file check to default to KDMP driver.

siva-portworx:
**What type of PR is this?**
bug
```
pb-2098: Added google file and azure file check to default to KDMP
    driver.

        - Removed IsDriverWithoutSnapshotSupport and
          isCSISnapshotClassRequired.
        - Added a IsCSIDriverWithoutSnapshotSupport common API in volume pkg
          and adding it in kdmp and csi driver.
        - Also added csiDriverWithOutSnapshotKey configmap field to get user
          input for driver that does not support snapshot.
```
no

No

no

**Testing:**
Verified the configmap field csiDriverWithOutSnapshotKey by setting it to efs by removed the efs case for testing purpose.
The backup was successful with the kdmp driver.
```
root@ip-70-0-97-59:~# kubectl  describe cm kdmp-config -n kube-system
Name:         kdmp-config
Namespace:    kube-system
Labels:       <none>
Annotations:  <none>

Data
====
KDMP_KOPIAEXECUTOR_REQUEST_CPU:
----
0.1
KDMP_KOPIAEXECUTOR_REQUEST_MEMORY:
----
700Mi
KDMP_MAINTENANCE_JOB_LIMIT:
----
5
KDMP_BACKUP_JOB_LIMIT:
----
5
KDMP_DELETE_JOB_LIMIT:
----
5
KDMP_KOPIAEXECUTOR_IMAGE:
----
portworx/kopiaexecutor:1.0.0
KDMP_KOPIAEXECUTOR_LIMIT_MEMORY:
----
1Gi
CSI_DRIVER_WITHOUT_SNAPSHOT:
----
efs.csi.aws.com
KDMP_KOPIAEXECUTOR_IMAGE_SECRET:
----

KDMP_KOPIAEXECUTOR_LIMIT_CPU:
----
0.2
KDMP_RESTORE_JOB_LIMIT:
----
5
Events:  <none>
root@ip-70-0-97-59:~#
```

### DIFF
--- a/drivers/volume/volume.go
+++ b/drivers/volume/volume.go
@@ -7,6 +7,7 @@ import (
 
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
 	snapshotVolume "github.com/kubernetes-incubator/external-storage/snapshot/pkg/volume"
+	"github.com/libopenstorage/stork/drivers"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/errors"
 	"github.com/portworx/sched-ops/k8s/core"
@@ -33,7 +34,15 @@ const (
 	// LinstorDriverName is the name of the Linstor driver implementation
 	LinstorDriverName = "linstor"
 	// KDMPDriverName is the name of the kdmp driver implementation
-	KDMPDriverName = "kdmp"
+	KDMPDriverName              = "kdmp"
+	pureCSIProvisioner          = "pure-csi"
+	vSphereCSIProvisioner       = "csi.vsphere.vmware.com"
+	efsCSIProvisioner           = "efs.csi.aws.com"
+	azureFileCSIProvisioner     = "file.csi.azure.com"
+	googleFileCSIProvisioner    = "com.google.csi.filestore"
+	pureBackendParam            = "backend"
+	pureFileParam               = "file"
+	csiDriverWithOutSnapshotKey = "CSI_DRIVER_WITHOUT_SNAPSHOT"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


drivers/volume/volume.go
```diff
@@ -689,3 +698,49 @@ func GetPVCFromObjects(objects []runtime.Unstructured, volumeBackupInfo *storkap
...
+		// pure FB csi driver does not support snapshot
+		if driverName == pureCSIProvisioner {
+			if pv.Spec.CSI.VolumeAttributes[pureBackendParam] == pureFileParam {
+				return true
+			}
+		}
+		// vsphere, efs, azure file and google file does not support snapshot.
+		// So defaulting to kdmp by not setting volumesnapshot class.
+		switch driverName {
+		case vSphereCSIProvisioner:
```
diptiranjanpx, line 720: can we make this as a constant list and if driverName is one among those , then will return true.

drivers/volume/volume.go
```diff
@@ -689,3 +698,49 @@ func GetPVCFromObjects(objects []runtime.Unstructured, volumeBackupInfo *storkap
...
+			}
+			if len(kdmpData.Data[csiDriverWithOutSnapshotKey]) != 0 {
+				nameList := strings.Split(kdmpData.Data[csiDriverWithOutSnapshotKey], ",")
+				for _, name := range nameList {
+					if driverName == name {
+						return true
+					}
+				}
+			}
+			return false
```
diptiranjanpx, line 742: this return is not required.

drivers/volume/volume.go
siva-portworx: Fixed.

drivers/volume/volume.go
siva-portworx: Fixed

diptiranjanpx (approved):

PR #996, opened 2021-12-13
Set the new storage class on the PVC during ApplicationRestore

adityadani:
**What type of PR is this?**
>bug

During an ApplicationRestore, if the storage class mapping was provided, stork would not set the new
storage class on the destination side where the PVC was being restored

- If the storage class mapping is set on ApplicationRestore on the PVC's
source storage class is found in this mapping, then on the destination cluster where
the PVC is being recreated set the new storage class obtained from this mapping.

No

```release-note
Improvement: STORK now supports choosing a different storage class for a PVC which is being restored using the ApplicationRestore CR. This is currently supported only for Portworx and KDMP driver
User Impact: Users can backup a Portworx volume that was created using a specific storage class and volume parameters and restore it to a Portworx volume with a different storage and respective volume parameters. For ex. A repl=3 Portworx volume can be backed up and restored in to a repl=2 Portworx volume
```

Yes - 2.8

### DIFF
--- a/pkg/applicationmanager/controllers/applicationclone.go
+++ b/pkg/applicationmanager/controllers/applicationclone.go
@@ -541,6 +541,7 @@ func (a *ApplicationCloneController) prepareResources(
 			objects,
 			nil,
 			namespaceMapping,
+			nil, // no support for storage class mapping
 			pvNameMappings,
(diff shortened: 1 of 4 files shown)


prashanthpx (approved):
lgtm

siva-portworx (approved):

PR #1001, opened 2021-12-16
PB-1079: Don't list the related clusterrole and clusterrolebinding of a serivce account  if user does not have permission for that namespace.

diptiranjanpx:
signed-off-by: Diptiranjan

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

User should not be able to view resources of the namespace it does not have access to.

```release-note
Issue:
User Impact:
Resolution

```

**Test**

user2 has only access to kube-system namespace but not central namespace.
Able to view the clusterrole and clusterrolebinding.

With existing code:
all-resources-view-with-restricted-user-existing-code

With fix:
all-resources-view-with-restricted-user-with-fix


pkg/resourcecollector/clusterrole.go
```diff
@@ -19,6 +19,14 @@ func (r *ResourceCollector) subjectInNamespace(subject *rbacv1.Subject, namespac
 			return false, nil
 		}
 		if subject.Namespace == namespace {
+			// check if user has an access to sa in namespace
```
ram-infrac, line 22: this will reintroduced STOR-546 again, @siva-portworx  we discussed this offline that we wont be needing this right?

ram-infrac (changes requested):

pkg/resourcecollector/clusterrole.go
ram-infrac: kindly test out use case in stor-546 before merging this

ram-infrac (commented):
kindly test out use case in stor-546 before merging this

ram-infrac (approved):

pkg/resourcecollector/clusterrole.go
diptiranjanpx: you are right, in restore cases, if we bring this check, then it will be like serviceaccount will not be there and this check will fail with NotFound and then the restored clusterrolebinding will not be updated with the subject.

With the current fix, 
 
1.in resource collection kind of use cases we will  check for the k8s object for SA (so that this bug and also the right resources get shown/backed up to the user as per their RBAC)

2. In restore path, as the SA might not have been created before clusterrolebinding , so we can only check for the namespace check and not really look for the SA k8s obejcts.

diptiranjanpx:
> kindly test out use case in stor-546 before merging this

Tested and have updated in PB-1079.

siva-portworx (approved):

siva-portworx:
LGTM. @ram-infrac Can please review the new change. Dipti has validated the stor-546 as well.

PR #1005, opened 2021-12-24
[PB-2143]: Making pvc size same as csi snapshot size to avoid the clone failure.

diptiranjanpx:
signed-off-by: Diptiranjan

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

In case of filestores, the snapshot size usually is the whole filestore size.
pvc creation from snapshot fails if the pvc size is less than the snapshot size.
Hence making the restored pvc size same as that of the snapshot size if available in volumesnapshot.

```release-note
Issue:
User Impact:
Resolution

```

**Test**
Tested following scenarios
1. backup on both disk and file provisioner of gke csi.
2. Restoring from both latest and previous ones of both disk and file provisioner of gke csi.


siva-portworx (approved):
LGTM

PR #1006, opened 2021-12-29
pb-2118: Added changes to support cross-region backup in native GKE driver.

siva-portworx:
**What type of PR is this?**
bug
```
pb-2118: Added changes to support cross-region backup in native GKE
    driver.

        - Update the logic take the destination zone and region in the
          StartRestor function.
        - Updating the PVC resource with nodeselected annotation form
          the destination cluster based on the region/zone selected for
          restore.
```

no

```release-note
Issue: Cross-region backup and restore is supported in the GKE cloud setup.
User Impact: Now user take backup in one region and restore it in different region.
Resolution: Added support for cross-region backup / restore in GKE cloud setup.

```

yes, 2.8

Testing:
Restored pods details:
```
root@ip-70-0-97-59:~# kubectl  get pods -n fio -o wide
NAME    READY   STATUS    RESTARTS   AGE   IP            NODE                                          NOMINATED NODE   READINESS GATES
fio-0   2/2     Running   0          24h   10.104.0.36   gke-siva-2-1-1-b-default-pool-4d9fbaa5-xnm6   <none>           <none>
fio-1   2/2     Running   0          24h   10.104.2.46   gke-siva-2-1-1-b-default-pool-2c1929c0-dbdj   <none>           <none>
fio-2   2/2     Running   0          24h   10.104.1.48   gke-siva-2-1-1-b-default-pool-dca9f120-98d9   <none>           <none>
root@ip-70-0-97-59:~#
```
Screenshot 2021-12-29 at 9 40 47 AM


drivers/volume/volume.go
```diff
@@ -751,3 +751,81 @@ func IsCSIDriverWithoutSnapshotSupport(pv *v1.PersistentVolume) bool {
...
 }
+
+// GetNodeRegion - get the node region
+func GetNodeRegion() (string, error) {
+	nodes, err := core.Instance().GetNodes()
+	if err != nil {
+		logrus.Errorf("getNodeZones: getting node details failed: %v", err)
+		return "", fmt.Errorf("failed in getting the nodes: %v", err)
+	}
+	for _, node := range nodes.Items {
```
prashanthpx, line 762: Wouldn't this code give the first node always (assuming the node list is populated the same way across every call) or a random node name and its corresponding region as we are returning from the first iteration of the loop itself?

drivers/volume/volume.go
```diff
@@ -751,3 +751,81 @@ func IsCSIDriverWithoutSnapshotSupport(pv *v1.PersistentVolume) bool {
...
+		return node.Labels[v1.LabelTopologyRegion], nil
+	}
+	return "", fmt.Errorf("failed in getting the nodes: %v", err)
```
prashanthpx, line 765: When would this error msg be triggered as LabelTopologyRegion is always present for a node right?

prashanthpx (approved):
lgtm

drivers/volume/volume.go
```diff
@@ -751,3 +751,81 @@ func IsCSIDriverWithoutSnapshotSupport(pv *v1.PersistentVolume) bool {
...
+func GetNodeZones() ([]string, error) {
+	nodes, err := core.Instance().GetNodes()
+	if err != nil {
+		logrus.Errorf("getNodeZones: getting node details failed: %v", err)
+		return nil, fmt.Errorf("failed in getting the nodes: %v", err)
+	}
+	var nodeZoneList []string
+	for _, node := range nodes.Items {
+		nodeZone := strings.Split(node.Labels[v1.LabelTopologyZone], "-")
+		nodeZoneList = append(nodeZoneList, nodeZone[2])
```
adityadani, line 778: This should be a map instead of a list. Two nodes can be a part of the same zone.

drivers/volume/volume.go
```diff
@@ -751,3 +751,81 @@ func IsCSIDriverWithoutSnapshotSupport(pv *v1.PersistentVolume) bool {
...
+}
+
+// GetVolumeBackupZones - Get zone details from the volumeBackupInfo
+func GetVolumeBackupZones(
+	volumeBackupInfos []*storkapi.ApplicationBackupVolumeInfo,
+) []string {
+	var backupZoneList []string
+	for _, volume := range volumeBackupInfos {
+		backupZone := strings.Split(volume.Zones[0], "-")
+		backupZoneList = append(backupZoneList, backupZone[2])
```
adityadani, line 790: same here: shouldn't this be a map instead  so we that have unique list of zones?

drivers/volume/volume.go
adityadani: I see that you have a Unique() helper function. It would be better to filter out the list here itself in this function. so that callers of this function dont need to worry about calling Unique()

pkg/resourcecollector/persistentvolumeclaim.go
```diff
@@ -71,6 +75,22 @@ func (r *ResourceCollector) preparePVCResourceForApply(
...
+	nodes, err := core.Instance().GetNodes()
+	if err != nil {
+		return false, fmt.Errorf("failed in getting the nodes: %v", err)
+	}
+	for _, vol := range vInfos {
+		if vol.PersistentVolumeClaim == pvc.Name {
+			for _, node := range nodes.Items {
+				nodeZone := node.Labels[v1.LabelTopologyZone]
+				if nodeZone == vol.Zones[0] {
+					pvc.Annotations[pvutil.AnnSelectedNode] = node.Name
```
adityadani, line 87: add a break here?

drivers/volume/volume.go
siva-portworx: To avoid the compilation error, I added it.

drivers/volume/volume.go
siva-portworx: Fixed. Adding the logic to filter it out in GetNodeZones itself.

adityadani (approved):

PR #1010, opened 2022-01-04
pb-2151: Added check to make sure zone array is not empty in preparePVCResourceForApply

siva-portworx:
**What type of PR is this?**
bug
pb-2151: Added check to make sure zone array is not empty in preparePVCResourceForApply.
For kdmp, we will not have zones set. So added check for the zone array before accessing it.

No

No

2.8

### DIFF
--- a/pkg/resourcecollector/persistentvolumeclaim.go
+++ b/pkg/resourcecollector/persistentvolumeclaim.go
@@ -83,9 +83,14 @@ func (r *ResourceCollector) preparePVCResourceForApply(
 		if vol.PersistentVolumeClaim == pvc.Name {
 			for _, node := range nodes.Items {
 				nodeZone := node.Labels[v1.LabelTopologyZone]
-				if nodeZone == vol.Zones[0] {
-					pvc.Annotations[pvutil.AnnSelectedNode] = node.Name
-					break
+				if len(vol.Zones) != 0 {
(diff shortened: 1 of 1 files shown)


pkg/resourcecollector/persistentvolumeclaim.go
```diff
@@ -83,9 +83,11 @@ func (r *ResourceCollector) preparePVCResourceForApply(
 		if vol.PersistentVolumeClaim == pvc.Name {
 			for _, node := range nodes.Items {
 				nodeZone := node.Labels[v1.LabelTopologyZone]
-				if nodeZone == vol.Zones[0] {
-					pvc.Annotations[pvutil.AnnSelectedNode] = node.Name
-					break
+				if len(vol.Zones) != 0 {
+					if nodeZone == vol.Zones[0] {
+						pvc.Annotations[pvutil.AnnSelectedNode] = node.Name
```
adityadani, line 88: nit: create the pvc.Annotations map if it is nil
you can have a pvc with no annotations

adityadani (approved):

pkg/resourcecollector/persistentvolumeclaim.go
siva-portworx: Fixed

PR #1011, opened 2022-01-06
Multi zone support generic backup support for aws/gke

prashanthpx:
**What type of PR is this?**
Bug

- As part of kdmp restore, it is made sure that kdmp job pod
  and the restore PV's are created in the same zone

No

```release-note
Issue:  Backup of sts app taken having multiple PVC's across zones fails to restore on the same region or different region
User Impact:   When a restore is performed, app pods fail to come up a few of the PV's could be in a different zone for each replica set of the app. Now user can do a successful restore of the same.
Resolution:  Added support to make sure all PV's are in the same node as the pod is running

```

Yes, 2.8

**Manual Test**
- Test details added in the bug with screen shots

### DIFF
--- a/drivers/volume/kdmp/kdmp.go
+++ b/drivers/volume/kdmp/kdmp.go
@@ -97,6 +97,8 @@ const (
 	storageClassKey           = "volume.beta.kubernetes.io/storage-class"
 	storageProvisioner        = "volume.beta.kubernetes.io/storage-provisioner"
 	storageNodeAnnotation     = "volume.kubernetes.io/selected-node"
+	gkeNodeLabelKey           = "cloud.google.com/gke-os-distribution"
+	awsNodeLabelKey           = "alpha.eksctl.io/cluster-name"
 )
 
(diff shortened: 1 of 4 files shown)


drivers/volume/kdmp/kdmp.go
```diff
@@ -180,6 +188,94 @@ func isCSISnapshotClassRequired(pvc *v1.PersistentVolumeClaim) bool {
...
 
+func getRegionalZones(zones string) []string {
+	return strings.Split(zones, zoneSeperator)
+}
+
+func isRegional(zone string) bool {
+	return strings.Contains(zone, zoneSeperator)
+}
+
+func getZones(pv *v1.PersistentVolume) ([]string, error) {
```
siva-portworx, line 199: Suggestion: Will it be good, if we move this getZones to be a common code between the actual cloud driver and kdmp driver. Some thing like moving it in to volume.go. That if any fixes needed can be done in one place.

siva-portworx (approved):

PR #1018, opened 2022-01-13
Skipping zone checks for EFS provisioner for  generic backup/restore paths

prashanthpx:
Skipping zone checks for EFS provisioner for  generic backup/restore paths

No

Not needed as EFS support is coming in as part of this release

Yes, 2.8

Test results in the bug

### DIFF
--- a/drivers/volume/kdmp/kdmp.go
+++ b/drivers/volume/kdmp/kdmp.go
@@ -97,8 +97,9 @@ const (
 	storageClassKey           = "volume.beta.kubernetes.io/storage-class"
 	storageProvisioner        = "volume.beta.kubernetes.io/storage-provisioner"
 	storageNodeAnnotation     = "volume.kubernetes.io/selected-node"
-	gkeNodeLabelKey           = "cloud.google.com/gke-os-distribution"
+	gkeNodeLabelKey           = "topology.gke.io/zone"
 	awsNodeLabelKey           = "alpha.eksctl.io/cluster-name"
+	ocpAWSNodeLabelKey        = "topology.ebs.csi.aws.com/zone"
 )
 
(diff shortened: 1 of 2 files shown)


drivers/volume/kdmp/kdmp.go
```diff
@@ -205,22 +205,25 @@ func getZones(pv *v1.PersistentVolume) ([]string, error) {
 	}
 
 	if awsCluster {
-		ebsName := storkvolume.GetEBSVolumeID(pv)
```
adityadani, line 208: The way we are determining if this is a AWS or a GCE cluster in getCloudProvider is incorrect. Right now we are making hard checks if it is an EKS cluster or a GKE cluster, but you can also deploy an OCP cluster in AWS. Instead we should be looking at the topology labels by EBS
```
topology.ebs.csi.aws.com/zone=us-east-1d
```

There should be something similar for GCE as well.

adityadani (approved):
The changes look OK - but I have a comment on a different piece of code related to this change.

drivers/volume/kdmp/kdmp.go
prashanthpx: Made changes as follows
1. For AWS cluster will check `alpha.eksctl.io/cluster-name` or `topology.ebs.csi.aws.com/zone` is set.  For native EKS, topology is  `topology.kubernetes.io/region`, not sure if any other providers would have the same. Checking for `alpha.eksctl.io/cluster-name` seems better check.

2. For GKE will check `topology.gke.io/zone`.

But one open item is if we have OCP on GKE, not sure what the labels would be.  For now, pushing the changes, once we have OCP deployed on GKE can add that label as separate PR.

PR #1019, opened 2022-01-14
[PB-2148]: Implementing job framework to make pvc bound in case of waitforfirstconsumer case for generic backups.

diptiranjanpx:
signed-off-by: Diptiranjan

**What type of PR is this?**
>bug
feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

If volumebindingmode is WaitForFirstConsumer, in case of csi/generic volume drivers we create the pvc from the local snapshot in case of restore. Now a temp job has been added just to mount the pvc and do nothing more so that we have a real PV associated with the pvc once the volume restore is declared as successful.

```release-note
Issue: Generic backup and restore was failing if volume binding mode was WaitForFirstConsumer and if the volume supports csi snapshots.
User Impact: Generic backup and restore functionality were not working
Resolution: Fixed with the new volume binding job framework

```

**Test**
Covered following scenarios

1. Generic Backup and restore of csi volumes with snapshotting with volumebinding mode as WaitForFirstConsumer.
2.  Generic Backup and restore of csi volumes with snapshotting with volumebinding mode as immediate.
3. Native backup and restore of gke volumes with snapshotting with volumebinding mode as WaitForFirstConsumer.

### DIFF
--- a/pkg/snapshotter/snapshotter_csi.go
+++ b/pkg/snapshotter/snapshotter_csi.go
@@ -12,6 +12,7 @@ import (
 
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1beta1"
 	kSnapshotClient "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
+	"github.com/libopenstorage/stork/drivers"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/crypto"
 	"github.com/libopenstorage/stork/pkg/objectstore"
@@ -52,11 +53,9 @@ const (
 	snapDeleteAnnotation = "snapshotScheduledForDeletion"
 	// snapRestoreAnnotation needs to be set if volume snapshot is scheduled for restore
 	snapRestoreAnnotation = "snapshotScheduledForRestore"
-	// kdmp config map
-	kdmpConfig = "kdmp-config"
 	// pvcNameLenLimitForJob is the max length of PVC name that the bound job
 	// will incorporate in their names
-	pvcNameLenLimitForJob = 237
+	pvcNameLenLimitForJob = 48
 	// shortRetryTimeout gets used for retry timeout
 	shortRetryTimeout = 30 * time.Second
 	// shortRetryTimeout gets used for retry timeout interval
@@ -1091,6 +1090,9 @@ func (c *csiDriver) createJob(pvc *v1.PersistentVolumeClaim, namespace string) (
 	if err == nil {
 		return job, nil
 	}
+	if !k8s_errors.IsNotFound(err) {
+		return nil, err
+	}
 
 	// Setup service account
 	if err := utils.SetupServiceAccount(jobName, namespace, roleFor()); err != nil {
@@ -1137,21 +1139,6 @@ func (c *csiDriver) deletejob(pvc *v1.PersistentVolumeClaim, namespace string) e
 		if err := utils.CleanServiceAccount(jobName, namespace); err != nil {
 			return nil, true, fmt.Errorf("deletion of service account %s/%s failed: %v", namespace, jobName, err)
 		}
-
-		if _, err := batch.Instance().GetJob(jobName, namespace); err != nil && !k8s_errors.IsNotFound(err) {
-			if err == nil {
-				errMsg := fmt.Sprintf("job %s/%s still exists after deletion: %v", namespace, jobName, err)
-				return nil, true, fmt.Errorf(errMsg)
-			}
-		}
-
-		if _, err := core.Instance().GetServiceAccount(jobName, namespace); err != nil && !k8s_errors.IsNotFound(err) {
-			if err == nil {
-				errMsg := fmt.Sprintf("service account %s/%s still exists after deletion: %v", namespace, jobName, err)
-				return nil, true, fmt.Errorf(errMsg)
-			}
-		}
-
 		pods, err := core.Instance().GetPodsUsingPVC(pvc.Name, namespace)
 		if err == nil && len(pods) > 0 {
 			logrus.Debugf("pvc %s/%s is still getting used by job pod %s", namespace, pvc.Name, pods[0].Name)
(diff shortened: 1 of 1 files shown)


pkg/snapshotter/snapshotter_csi.go
```diff
@@ -46,6 +52,15 @@ const (
 	snapDeleteAnnotation = "snapshotScheduledForDeletion"
 	// snapRestoreAnnotation needs to be set if volume snapshot is scheduled for restore
 	snapRestoreAnnotation = "snapshotScheduledForRestore"
+	// kdmp config map
+	kdmpConfig = "kdmp-config"
```
prashanthpx, line 56: Please use `KdmpConfigmapName = "kdmp-config"` from drivers.go here

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -382,11 +397,11 @@ func (c *csiDriver) RestoreStatus(pvcName, namespace string) (RestoreInfo, error
 	}
 
 	switch {
-	case c.pvcWaitingForFirstConsumer(pvc):
-		restoreInfo.VolumeName = pvc.Spec.VolumeName
-		restoreInfo.Size = size
-		restoreInfo.Status = StatusReady
 	case c.pvcBindFinished(pvc):
+		err := c.deletejob(pvc, namespace)
```
prashanthpx, line 401: In the corresponding kdmp PR we have `cleanupJobBoundResources()` which is called part of DE CR cleanup.  What is the use for calling it here?
Can we add some comments also on the same?

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -1052,3 +1083,174 @@ func (c *csiDriver) CleanUpRestoredResources(backupLocation *storkapi.BackupLoca
...
+	jobName := toBoundJobPVCName(pvc.Name, string(pvc.GetUID()))
+	t := func() (interface{}, bool, error) {
+		if err := batch.Instance().DeleteJobWithForce(jobName, namespace); err != nil && !k8s_errors.IsNotFound(err) {
+			return nil, true, fmt.Errorf("deletion of job %s/%s failed: %v", namespace, jobName, err)
+		}
+		if err := utils.CleanServiceAccount(jobName, namespace); err != nil {
+			return nil, true, fmt.Errorf("deletion of service account %s/%s failed: %v", namespace, jobName, err)
+		}
+
+		if _, err := batch.Instance().GetJob(jobName, namespace); err != nil && !k8s_errors.IsNotFound(err) {
```
prashanthpx, line 1141: Any reason we want to check if those resources are present?

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -1052,3 +1083,174 @@ func (c *csiDriver) CleanUpRestoredResources(backupLocation *storkapi.BackupLoca
 
 	return nil
 }
+
+func (c *csiDriver) createJob(pvc *v1.PersistentVolumeClaim, namespace string) (*batchv1.Job, error) {
+	jobName := toBoundJobPVCName(pvc.Name, string(pvc.GetUID()))
+	// if already the job is running or in completed state , no need to rerun the job
+	job, err := batch.Instance().GetJob(jobName, namespace)
+	if err == nil {
```
prashanthpx, line 1091: If there is an error we still proceed with another part of the code which might be a problem.
Need to handle error something like this 

`if err != nil && !apierrors.IsNotFound(err)`

Also, we can't just rely on `err == nil` to know a job is completed 
Let's take a case where:
1. First-time `CreateJob()` is called, API gives no error but eventually job fails for some reason 
2. Next time the reconciler comes up and again enters `createJob()` in which case `GetJob()` doesn't give an error and if we return from here then we would never start the job again.

Jobs status needs to be fetched as follows
```
var jobStatus batchv1.JobConditionType
	if len(job.Status.Conditions) != 0 {
		jobStatus = job.Status.Conditions[0].Type

	}
```
You can refer to `JobStatus()` code in KDMP.

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -1052,3 +1083,174 @@ func (c *csiDriver) CleanUpRestoredResources(backupLocation *storkapi.BackupLoca
...
+			Template: v1.PodTemplateSpec{
+				Spec: v1.PodSpec{
+					RestartPolicy:      v1.RestartPolicyOnFailure,
+					ImagePullSecrets:   utils.ToImagePullSecret(utils.KopiaExecutorImageSecret(kdmpConfig, "kube-system")),
+					ServiceAccountName: jobName,
+					Containers: []v1.Container{
+						{
+							Name:            "kopiaexecutor",
+							Image:           utils.KopiaExecutorImage(kdmpConfig, "kube-system"),
+							ImagePullPolicy: v1.PullAlways,
```
prashanthpx, line 1199: Make this as `PullIfNotPresent` and also `KdmpConfigmapNamespace` instead of "kube-system"

pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: This is to handle the success case where the volume gets bound.
The cleanup in reconciler needs to delete the job if volume does not become bound.

pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: done

pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: ok initially added a safe check. will remove it.

pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: No, we are not going to check for the job status here.
As mentioned earlier , the job to fail the pod needs to get bound and thats essentially we want.

But also ideally createjob should have proceeded if there is error and the error should be also NotFound type.
Addressed in the next commit where if the error is of NotFound kind , createJob will return error.

pkg/snapshotter/snapshotter_csi.go
diptiranjanpx: done

prashanthpx (approved):

PR #1024, opened 2022-01-19
STOR-573 apply updatedObjects in parallel during migration

lpitstickpstg:
**What type of PR is this?**
improvement

Use go routines to parallelize applying resources during migration for improved performance

New int arg for stork "migration-max-threads" (default 4)

```release-note
Issue: 
    New "migration-max-threads" arg for stork (default 4) allowing parallel application of  kubernetes resources during migration.
User Impact:
    Improved migration performance for large number of kubernetes resources.


```

no

**RESULTS**
1k config maps on 4 core vms
  1 thread: 15m
10 thread: 15m

So far no difference, could be the type of resources I used, they are very small, or the VM cluster environment. Will continue experimenting. Aditya suggested we may be bottlenecked on the k8 client side, I'll dig into where we're spending time.

**RESULTS 2**
Tested with 100 config maps, 100 secrets, 500 services and scrambled the order we apply resources to allow mixing of resource types being applied at the same time. Used repeated migration of the same resources (didn't delete everything in-between)
1 thread:  8.5m
4 threads: 7m

I did test applying resources with a script applying secrets at the same time, and it was inconsistently ~15% faster than serially. Kubernetes just doesn't handle many updates effectively in parallel as far as I can tell.

Conclusion: We can get some improvement with this parallelization, but it's not 4x with 4 threads like we hoped.

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1697,109 +1700,149 @@ func (m *MigrationController) applyResources(
...
+		for _, o := range objects {
+			objectChan <- o
+		}
+		close(objectChan)
+	}()
+
+	for result := 0; result < numObjects; result++ {
+		workerErr := <-errorChan
+		if workerErr != nil {
+			return workerErr
```
adityadani, line 1845: We need to let the other goroutines know that we are not going to be processing any more updates when one of them returns an error. In this way you can deterministically cleanup those goroutines.

lpitstickpstg: Changed to use unbuffered channels, so when we find an error or finish results we close objectChan and the workers will exit immediately after their current piece of work. This should address both comments.

adityadani (approved):

PR #1026, opened 2022-01-21
STOR-577: Portworx Driver: Do an explicit not found check in DeletePair

adityadani:
**What type of PR is this?**
>bug

- Change the error condition to do a string check on "not found" in DeletePair()

No

No

Yes - 2.9

**Testing Notes**
- Issued a cluster pair delete on the Portworx cluster using `pxctl cluster pair delete`
- Then issued a delete on the ClusterPair object in k8s. The k8s resource got deleted.


Rohit-PX (approved):
LGTM.

PR #1036, opened 2022-02-07
Always sync volumesnapshotschedule and volumesnapshot status

ram-infrac:
**What type of PR is this?**
>bug

Make sure `volumesnapshotschedule` snapshot status always in-sync with `volumesnapshots` triggered by it 

no

yes
```release-note
Issue: SnapshotSchedule have incorrect sync status with snapshot triggered if underlying driver went offline at time of SnapshotCreate API call 
User Impact: SnapshotSchedule will have Error status but Snapshot object would have successful status for same snapshot
Resolution: Fixed issue to make volumesnapshotschedule and volumesnapshot object status always in-sync
```

yes, 2.9.0

### DIFF
--- a/pkg/snapshot/controllers/snapshotschedule.go
+++ b/pkg/snapshot/controllers/snapshotschedule.go
@@ -193,10 +193,22 @@ func (s *SnapshotScheduleController) updateVolumeSnapshotStatus(snapshotSchedule
 			if snapshot.Status != snapv1.VolumeSnapshotConditionReady {
 				pendingVolumeSnapshotStatus, err := getVolumeSnapshotStatus(snapshot.Name, snapshotSchedule.Namespace)
 				if err != nil {
-					return err
+					s.recorder.Event(snapshotSchedule,
+						v1.EventTypeWarning,
+						err.Error(),
+						fmt.Sprintf("Error updating snapshot (%s) status", snapshot.Name))
+					if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/snapshot/controllers/snapshotschedule.go
```diff
@@ -190,31 +190,28 @@ func (s *SnapshotScheduleController) updateVolumeSnapshotStatus(snapshotSchedule
...
+						fmt.Sprintf("Scheduled snapshot (%v) completed successfully", snapshot.Name))
+				} else {
+					s.recorder.Event(snapshotSchedule,
+						v1.EventTypeWarning,
+						string(snapv1.VolumeSnapshotConditionError),
+						fmt.Sprintf("Scheduled snapshot (%v) failed", snapshot.Name))
 				}
-				updated = true
 			}
+			updated = true
```
adityadani, line 214: Wouldn't this end up updating the VolumeSnapshotSchedule endlessly? The `updated` boolean should be set only if snapshotStatus != pendingVolumeSnapshotStatus before line 199

ram-infrac: yes,  we shouldn't retry for `ready` snapshot. I think we can omit update of `volumesnapshotschedule` is latest snapshotstatus is same as entry in `volumesnapshotschedule` object

adityadani (approved):

PR #1039, opened 2022-02-10
Integration test for clusterpairing using backuplocation

ram-infrac:
**What type of PR is this?**
>integration-test

Integration test  for  using backuplocation for clusterpair 

no

no

yes, 2.9


test/integration_test/common_test.go
```diff
@@ -558,12 +569,11 @@ func scheduleClusterPair(ctx *scheduler.Context, skipStorage, resetConfig bool,
 		}
 	}
 
-	info, err := volumeDriver.GetClusterPairingInfo()
+	info, err = volumeDriver.GetClusterPairingInfo()
```
lpitstickpstg, line 572: This will overwrite the info we modified above?

Rohit-PX (approved):
Looks good as long as @lpitstickpstg 's comments are addressed.

PR #1046, opened 2022-02-21
PWX-22971 : Fix application clone failure at resource stage

ram-infrac:
**What type of PR is this?**
>bug

ApplicationClone fails with error - 
```
Error cloning resources: VolumeDriver with UID/Name:  not found
```
This PR fix issue where driverName is always empty for non backup/restore, `PrepareResourceForApply` call

no

no

yes2.9
2.8 -> in case of backup release of 2.8.x 

Jenkins Job : 
```
healthmonitor : http://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%202.9-dev%20%20C7/job/stork-master-px-2.9-dev-healthmonitor/45/console
dev job from 43-45 : http://jenkins.pwx.dev.purestorage.com/job/Stork/job/stork-dev-test-2/
application clone : http://jenkins.pwx.dev.purestorage.com/job/Stork/job/stork-dev-test-2/48/console
```


test/integration_test/health_monitor_test.go
```diff
@@ -19,7 +19,7 @@ import (
 const (
 	// node offline timeout just above 4.5 minutes
 	// which is the max time stork could take to delete a app pod.
-	nodeOfflineTimeout = 271 * time.Second
+	nodeOfflineTimeout = 295 * time.Second
```
lpitstickpstg, line 22: Did performance degrade?

ram-infrac: No, health monitor switch period is still 4.5 mins.
Torpedo StopDriver() labels node to stop Portworx driver, but does not actually waits for driver to shutdown only grace period of 10 Secs for some reason. that's why this increase timeout is needed

ram-infrac: @Rohit-PX  I think we should fix in torpedo will need it eventually https://github.com/portworx/torpedo/blob/master/drivers/volume/portworx/portworx.go#L1374

adityadani (approved):

PR #1048, opened 2022-02-27
pb-2172: Added fixes related to handling of rule command executor image.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-2172: Added fixes related to handling of rule command executor image.

        - Added CMD-EXECUTOR-IMAGE-REGISTRY and CMD-EXECUTOR-IMAGE-REGISTRY-SECRET environment variable
          to get the custom regitry name and secret for rule command executor pod image.
        - Added the annotation to get the image registry secret value,
          if some one pass the custom image registry value as annotation in the rule CR.
        - Added logic to take the image registry name and secret value
          from the stork deployment if the both the env and rule annotation is missing.
        - Also, add a fix to fail the backup and delete the rule command
          executor pod, if the rule command executor pod is struck in Pending phase for more
          than five minutes. Some times, it gets struck in pending state,
          if there is any issue with the image registry or secret value.
        - Minor fix in printing a err in handle function.
```

No

```release-note
Issue: If the rule command executor image is not passed as annotation in rule CR, the rule command executor pod hangs forever.
User Impact: The used does not know that there is a issue in the image setting for rule command executor.
Resolution: Now we will fails the backup and delete the rule command executor pod, if it is struck in pending state for more than five mintues.

```

Yes, Next release of stork that will happen for 2.2.0 px-backup.

**Testing:**
- With the debug statements, verified the registry name are picked correctly from the stork deployment spec as a default value.
- Stimulated the backup with incorrect rule command executor image name:
Screenshot 2022-02-27 at 8 39 42 PM

### DIFF
--- a/pkg/k8sutils/k8sutils.go
+++ b/pkg/k8sutils/k8sutils.go
@@ -3,9 +3,11 @@ package k8sutils
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/portworx/sched-ops/k8s/apiextensions"
+	"github.com/portworx/sched-ops/k8s/apps"
 	"github.com/portworx/sched-ops/k8s/core"
 	v1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -19,6 +21,10 @@ import (
 const (
 	crdTimeout    = 1 * time.Minute
 	retryInterval = 5 * time.Second
+	// StorkDeploymentName - stork deployment name
+	StorkDeploymentName = "stork"
+	// DefaultAdminNamespace - default admin namespace, where stork will be installed
+	DefaultAdminNamespace = "kube-system"
 )
 
 // GetPVCsForGroupSnapshot returns all PVCs in given namespace that match the given matchLabels. All PVCs need to be bound.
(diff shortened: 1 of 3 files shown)


pkg/k8sutils/k8sutils.go
```diff
@@ -154,3 +158,17 @@ func CreateCRD(resource apiextensions.CustomResource) error {
 	}
 	return nil
 }
+
+// GetImageRegistryFromStorkDeployment - extract image registry and image registry secret from stork deployment spec
+func GetImageRegistryFromStorkDeployment() (string, string, error) {
```
ram-infrac, line 163: lets pass in name/namespace as well so that we can use this call for other importing libs where stork is not installed in default ns as well as get any deployment image details

pkg/rule/rule.go
```diff
@@ -629,6 +674,29 @@ func runBackgroundCommandOnPods(pods []v1.Pod, container, cmd, taskID, cmdExecut
...
+		if p.Status.Phase == v1.PodPending {
+			if i == (maxRetry - 1) {
+				var err error
+				if len(p.Status.ContainerStatuses) != 0 {
+					err = fmt.Errorf("rule command executor is struck in pending state, reason: %v", p.Status.ContainerStatuses[0].State.Waiting.Reason)
+				} else {
+					err = fmt.Errorf("rule command executor is struck in pending state")
+				}
+				logrus.Errorf("%v", err)
+				return err
```
ram-infrac, line 693: missing pod delete before return?

ram-infrac (approved):
lgtm, after minor nit fixes. 
Lets make sure to run `TestApplicationBackup` suite to make sure no regression before merging

pkg/rule/rule.go
siva-portworx: There is a defer func that takes care of deleting the pod.

pkg/k8sutils/k8sutils.go
siva-portworx: Fixed. Made the API to generic by taking deployment name and namespace.

PR #1051, opened 2022-03-10
Pb 2219: Add changes to use the registry name and image secret for kopia executor from stork deployment spec.

siva-portworx:
**What type of PR is this?**
bug

```
    pb-2219: Add changes to use the registry name and image secret for kopia
    executor from stork deployment spec.

        - Add GetStorkPodNamespace to get the stork pod namespace, even
          if it is not installed in the kube-system.
```

No

```release-note
Issue: In the current releases, customer need to configure the kopia executor image registry and secret details in the configmap after installing the stork.
User Impact: User need to configure the registy and image secret in the configmap separately.
Resolution: Now these details will be picked up from the strok deployment spec. If it is different registry for kopia executor, customer need to set the details in ENV variable in stork deployment spec.
KOPIA-EXECUTOR-IMAGE-REGISTRY
KOPIA-EXECUTOR-IMAGE-REGISTRY-SECRET

```

for now it will be in master.

### DIFF
--- a/pkg/snapshotter/snapshotter_csi.go
+++ b/pkg/snapshotter/snapshotter_csi.go
@@ -12,9 +12,9 @@ import (
 
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1beta1"
 	kSnapshotClient "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
-	"github.com/libopenstorage/stork/drivers"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/crypto"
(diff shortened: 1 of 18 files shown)


pkg/snapshotter/snapshotter_csi.go
```diff
@@ -1163,7 +1163,25 @@ func buildJobSpec(
 	}, " ")
 
 	jobPodBackOffLimit := int32(1)
-
+	storkPodNs, err := k8sutils.GetStorkPodNamespace()
```
diptiranjanpx, line 1166: need to have a check for storkPodNs too. it can be empty string.

diptiranjanpx (approved):

pkg/snapshotter/snapshotter_csi.go
siva-portworx: Added a check in GetStorkPodNamespace() function itself to return err, if the ns is empty.

PR #1052, opened 2022-03-10
PWX-23026: Check if underlying storage pair is used by another cluste…

ram-infrac:
**What type of PR is this?**
>bug

If multiple Cluster-pairs are pointing to same storage, deleting one of them can result in deletion of 
storage cluster-pair. 
This PR add finaliser check for cluster-pair to hold deletion if another cluster-pair has the same storage
id

no

yes
```release-note
Issue: Underlying storage pair gets deleted if multiple ClusterPair's are pointing to same RemoteStorageID and the user deletes any one of them
User Impact: System will have ClusterPair in ready state, but underlying storage pair may not exist, which can result in migration failure
Resolution: ClusterPair will be deleted and underlying Storage Pair will be kept as it is, if another ClusterPair is pointing to same RemoteStorageID
```

yes ,2.9.0

### DIFF
--- a/pkg/migration/controllers/clusterpair.go
+++ b/pkg/migration/controllers/clusterpair.go
@@ -218,7 +218,24 @@ func getClusterPairSchedulerStatus(clusterPairName string, namespace string) (st
 }
 
 func (c *ClusterPairController) cleanup(clusterPair *stork_api.ClusterPair) error {
+	skipDelete := false
 	if clusterPair.Status.RemoteStorageID != "" {
+		// verify if any other cluster pair using the same RemoteStorageID
+		cpList, err := storkops.Instance().ListClusterPairs("")
+		if err != nil {
+			return err
+		}
+		for _, cp := range cpList.Items {
+			// No need to handle current reconciled clusterpair
+			// Since clusterPair will have deleteTimeStamp set and will be ignored
+			if cp.Status.RemoteStorageID == clusterPair.Status.RemoteStorageID &&
+				cp.DeletionTimestamp == nil {
+				skipDelete = true
+				break
+			}
+		}
+	}
+	if !skipDelete {
 		return c.volDriver.DeletePair(clusterPair)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


lpitstickpstg:
Cleaned up the release note. @ram-infrac For the resolution, does the entire ClusterPair CRD not get deleted, or is it just the "storage pair" (I'm not sure exactly what this is, does the customer?) That doesn't get deleted?

> Issue:  Underlying storage pair gets deleted if multiple ClusterPair's are pointing to same RemoteStorageID and the user deletes any one of them
> User Impact:  System will have ClusterPair in `ready` state, but underlying storage pair may not exist, which can result in migration failure 
> Resolution: ClusterPair will not be deleted if another ClusterPair is pointing to same RemoteStorageID

lpitstickpstg (approved):
Code looks good, see release note comment

pkg/migration/controllers/clusterpair.go
```diff
@@ -218,10 +218,29 @@ func getClusterPairSchedulerStatus(clusterPairName string, namespace string) (st
 }
 
 func (c *ClusterPairController) cleanup(clusterPair *stork_api.ClusterPair) error {
+	skipDelete := false
+	cpName := ""
```
adityadani, line 222: nit: Collect all the clusterPairs and their namespaces so that we can show them in the log message and/or event.

pkg/migration/controllers/clusterpair.go
```diff
@@ -218,10 +218,29 @@ func getClusterPairSchedulerStatus(clusterPairName string, namespace string) (st
...
+				cpName = cp.Namespace + "/" + cp.Name
+				break
+			}
+		}
+	}
+	if !skipDelete {
 		return c.volDriver.DeletePair(clusterPair)
 	}
-	return nil
+	return fmt.Errorf("underlying storage pair is being used by another clusterpair: %s", cpName)
```
adityadani, line 243: If you return an error here, the stork ClusterPair CR will not deleted. IMO we should let the CR get deleted. Usually someone would be deleting a namespace and the cluster pair not getting deleted will block it.

adityadani (approved):
lgtm after comments are addressed

adityadani (approved):

PR #1054, opened 2022-03-15
pb-2179: Added changes required for the supporting object lock for backupschedule deletion.

siva-portworx:
**What type of PR is this?**
>bug
```
pb-2179: Added changes required for the supporting object lock for backupschedule deletion.

        - Adding the "object-lock-retention-period" annotation to the
          applicationbackup CR created by applicationbackupschedule in stork
        - This retention period annotation value will be used in the backupscync
          logic to update the retention period of the synced backup object.
        - Added GetObjLockInfo in stork objectstore as it needs to be called in
          stork as well as px-backup
```

no

no ( From stork, we are not claiming the object lock support, So nothing need to documented.)

Currently pushing it to master branch.

Testing:
Will post the testing result in the px-backup PR.

### DIFF
--- a/pkg/objectstore/s3/s3.go
+++ b/pkg/objectstore/s3/s3.go
@@ -2,6 +2,7 @@ package s3
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -9,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
+	"github.com/libopenstorage/stork/pkg/objectstore/common"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/s3blob"
 )
@@ -62,7 +64,7 @@ func CreateBucket(backupLocation *stork_api.BackupLocation) error {
 }
 
 // GetObjLockInfo fetches the object lock configuration of a S3 bucket
-func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*s3.GetObjectLockConfigurationOutput, error) {
+func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*common.ObjLockInfo, error) {
 	sess, err := getSession(backupLocation)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/objectstore/s3/s3.go
```diff
@@ -82,5 +84,21 @@ func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*s3.GetObjectLock
 		}
 		return nil, err
 	}
-	return out, err
+	objLockInfo := &common.ObjLockInfo{}
```
lalat-das, line 87: While we have brought  the object store changes from px-backup to stork, can we assess how good bad ugly it would be to bring back S3 minio related error handling(_Method not supported_) that occurs for a lower version of minio server( no support for object lock) . It will be something to be added after line 83.

pkg/objectstore/s3/s3.go
```diff
@@ -82,5 +84,21 @@ func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*s3.GetObjectLock
...
+		}
+		if out.ObjectLockConfiguration.Rule != nil &&
+			out.ObjectLockConfiguration.Rule.DefaultRetention != nil {
+			objLockInfo.LockMode = aws.StringValue(out.ObjectLockConfiguration.Rule.DefaultRetention.Mode)
+			objLockInfo.RetentionPeriodYears = aws.Int64Value(out.ObjectLockConfiguration.Rule.DefaultRetention.Years)
+			objLockInfo.RetentionPeriodDays = aws.Int64Value(out.ObjectLockConfiguration.Rule.DefaultRetention.Days)
+		} else {
+			//This is an invalid object-lock config, no default-retention but object-loc enabled
+			objLockInfo.LockEnabled = false
+			return nil, fmt.Errorf("only object lock is enabled but no retention policy is set on the bucket")
```
lalat-das, line 100: Can we please remove that word **_"only"_** from the front, it makes the error statement little informal. or Please reword it if you think it can be rewritten differently.

lalat-das (approved):

pkg/objectstore/s3/s3.go
siva-portworx: Added it. Will take care of remove it in px-backup code.

pkg/objectstore/s3/s3.go
siva-portworx: Fixed.

prashanthpx:
Had a look at the PR, LGTM

PR #1055, opened 2022-03-15
PWX-23212 Treat StorageDown nodes as degraded instead of Online

lpitstickpstg:
**What type of PR is this?**
bug

Classifies StorageDown as Degraded node state for scheduling

no

no

2.9

**Notes**
Still not working because we don't actually receive the right status (PWX-23234)


adityadani (approved):
Changes looks good from stork's perspective

PR #1058, opened 2022-03-20
pb-2236: Added logic to force full backup for locked period backup from schedulebackup.

siva-portworx:
**What type of PR is this?**
>bug
```
    pb-2236: Added logic to force full backup for locked period backup from schedulebackup

        - Starting of every day time slot, forcing to have first backup of schedulebackup to
          full backup always.
        - This is to take care, if the incremental backup was overlapping between two day's
          time slot and avoid having incremental backup as first backup in starting of day.
```

No

Since we are not going to claim the object lock support in direct stork backup, we will not release note it. 

no

Testing:
For testing reduced dayInSec to 900 sec ( 15 mints) and tested that it takes fully backup for every 15 mint cycle, even though the default increment count is 5.
Screenshot 2022-03-20 at 11 38 30 PM
Screenshot 2022-03-20 at 11 38 17 PM
Screenshot 2022-03-20 at 11 38 04 PM
Screenshot 2022-03-20 at 11 37 43 PM
Screenshot 2022-03-20 at 11 37 21 PM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackupschedule.go
+++ b/pkg/applicationmanager/controllers/applicationbackupschedule.go
@@ -41,6 +41,8 @@ const (
 	// ApplicationBackupObjectLockRetentionAnnotation - object lock retention period annotation
 	// Since this annotation is used in the px-backup, creating with portworx.io annotation prefix.
 	ApplicationBackupObjectLockRetentionAnnotation = "portworx.io/" + "object-lock-retention-period"
+	incrementalCountAnnotation                     = "portworx.io/cloudsnap-incremental-count"
+	dayInSec                                       = 86400
 )
 
 // NewApplicationBackupSchedule creates a new instance of ApplicationBackupScheduleController.
@@ -125,7 +127,6 @@ func (s *ApplicationBackupScheduleController) handle(ctx context.Context, backup
 			log.ApplicationBackupScheduleLog(backupSchedule).Error(msg)
 			return nil
 		}
-
 		// Start a backup for a policy if required
 		if start {
 			err := s.startApplicationBackup(backupSchedule, policyType)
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackupschedule.go
```diff
@@ -41,6 +41,8 @@ const (
 	// ApplicationBackupObjectLockRetentionAnnotation - object lock retention period annotation
 	// Since this annotation is used in the px-backup, creating with portworx.io annotation prefix.
 	ApplicationBackupObjectLockRetentionAnnotation = "portworx.io/" + "object-lock-retention-period"
+	incrementalCountAnnotation                     = "portworx.io/cloudsnap-incremental-count"
+	dayInSec                                       = 5184000
```
lalat-das, line 45: Siva can you please explain this number, a day has 86400 seconds in it..

lalat-das (approved):

pkg/applicationmanager/controllers/applicationbackupschedule.go
siva-portworx: 86400 (60*60*24)is correct. Fixed it.

diptiranjanpx (approved):

pkg/applicationmanager/controllers/applicationbackupschedule.go
```diff
@@ -284,6 +285,28 @@ func getBackupLocationCR(backupSchedule *stork_api.ApplicationBackupSchedule) (*
...
+// isPeriodicSchedule - isPeriodicSchedule will check whether the sp has interval type is set.
+func isPeriodicSchedule(schedulePolicyCR *stork_api.SchedulePolicy) bool {
+	return schedulePolicyCR.Policy.Interval != nil
+}
+
+// getLatestSuccessfullbackupTime - will check return the last successful backup for backupschedule.
+// It will check the backup list present in the applicationbackupschedule CR.
+func getLatestSuccessfullbackupTime(backupSchedule *stork_api.ApplicationBackupSchedule) int64 {
+	var lastSuccessfulBackupCreateTime int64
+	for _, statusItems := range backupSchedule.Status.Items {
```
prashanthpx, line 297: If there are all bacups succesful, then how do we guarnatee that this gives us the last one as the list can be in any order? For all successful case, it can bail out in the first success itself and not go till the last one. My thinking was we sort the list based on Time and get the last or the first one (depending on sorting)

prashanthpx (approved):

pkg/applicationmanager/controllers/applicationbackupschedule.go
siva-portworx: As discussed the list of backup in the applicationbackupschedule CR is order list and start search from the end of the list for a successful backup.

PR #1061, opened 2022-03-23
pb-2260: Return default objLockInfo struct instead of error for gke and azure bucket in GetObjLockInfo

siva-portworx:
**What type of PR is this?**
>bug

```
pb-2260: Return default objLockInfo struct instead of error for gke and azure bucket in GetObjLockInfo
```

no

no

no

Testing:
Before fix:
Screenshot 2022-03-24 at 12 41 25 AM

After fix:
Screenshot 2022-03-24 at 1 22 17 AM

### DIFF
--- a/pkg/objectstore/azure/azure.go
+++ b/pkg/objectstore/azure/azure.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/objectstore/common"
+	"github.com/sirupsen/logrus"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/azureblob"
 )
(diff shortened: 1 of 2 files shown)


diptiranjanpx (approved):

pkg/objectstore/google/google.go
```diff
@@ -64,5 +64,7 @@ func CreateBucket(backupLocation *stork_api.BackupLocation) error {
 
 // GetObjLockInfo fetches the object lock configuration of a bucket
 func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*common.ObjLockInfo, error) {
-	return nil, fmt.Errorf("object lock is not supported for google provider")
+	logrus.Infof("object lock is not supported for azure provider")
```
diptiranjanpx, line 67: typo : azure

pkg/objectstore/azure/azure.go
```diff
@@ -60,5 +61,7 @@ func CreateBucket(backupLocation *stork_api.BackupLocation) error {
 
 // GetObjLockInfo fetches the object lock configuration of a bucket
 func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*common.ObjLockInfo, error) {
-	return nil, fmt.Errorf("object lock is not supported for azure provider")
+	logrus.Infof("object lock is not supported for azure provider")
+	objLockInfo := &common.ObjLockInfo{}
+	return objLockInfo, nil
```
ram-infrac, line 66: not retrub &common.ObjLockInfo{}, nil

ram-infrac (approved):

pkg/objectstore/google/google.go
siva-portworx: Fixed

pkg/objectstore/azure/azure.go
siva-portworx: Fixed.

PR #1063, opened 2022-03-29
Upgrade openssl libs

ram-infrac:
**What type of PR is this?**
>bug

Upgrade openssl libs to latest for RHEL 8 Ubi image for CVE reported `CVE-2022-0778`

no

no

yes,2.9


sharma-tapas (approved):
We should file a task to check this in the next release and remove this as and when required when the fix is available in the base image.

PR #1064, opened 2022-03-29
pb-2267: fixed issue in object lock configmap name

lalat-das:
Changed the name of object lock configMap from stork-objLock-config to
stork-objlock-config. This is to avoid any capital letter in it.

bug

**What this PR does / why we need it**: The object lock config Map name has a capital letter in it which is causing the failure in creation of configMap itself and eventually failing to create the schedule backup pertaining to locked bucket. 

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**:No

Unit Test
a  schedule backup named "5min-schedule-bkp" is created and retention period is chnaged aftera successful backup, the next backup can be seen as failed with insufficient retention period message.

Pasted Graphic
 The failure reason mentioned below
image
The config map can seen as below
stork-objlock-config

### DIFF
--- a/pkg/k8sutils/k8sutils.go
+++ b/pkg/k8sutils/k8sutils.go
@@ -29,7 +29,7 @@ const (
 	// DefaultAdminNamespace - default admin namespace, where stork will be installed
 	DefaultAdminNamespace = "kube-system"
 	// StorkConfigMapName holds any generic config specific to stork module
-	StorkConfigMapName = "stork-objLock-config"
+	StorkConfigMapName = "stork-objlock-config"
 	// ObjectLockIncrBackupCountKey defines scheduleBackup's incremental backup count
 	ObjectLockIncrBackupCountKey = "object-lock-incr-backup-count"
 	// ObjectLockDefaultIncrementalCount defines default incremental backup count


diptiranjanpx (approved):

pkg/k8sutils/k8sutils.go
```diff
@@ -29,7 +29,7 @@ const (
 	// DefaultAdminNamespace - default admin namespace, where stork will be installed
 	DefaultAdminNamespace = "kube-system"
 	// StorkConfigMapName holds any generic config specific to stork module
-	StorkConfigMapName = "stork-objLock-config"
+	StorkConfigMapName = "stork-objlock-config"
```
ram-infrac, line 32: can we use one config map for all stork related configuration like kdmp,version and objectlog. maintaining multiple configmap will be troublesome. Lets setup call and sync up online since kdmp,version being used by px-backup already.

For now lets this get merged in

ram-infrac (approved):

pkg/k8sutils/k8sutils.go
siva-portworx: Agreed, In fact this object lock config option is only for testing purpose. Customer is not going to use it. That we can add it in kdmp_config configmap itself and ask QA to use it.

lalat-das: My initial thought was same, but then I didn't have much context of other config and we needed this to be in little early. Hence went that path. Particularly in "stork-config"  config map there is something related to stork-scheduler exist which need to be discussed. But I totally agree to merge them all.

lalat-das:
Kindly merge this PR. The travis build has passed. I don't have adequate permission.

PR #1066, opened 2022-03-31
pb-2266: Adding backupType in the applicationbackup CR created by schedule backups.

siva-portworx:
**What type of PR is this?**
bug

pb-2266

No

All add the release note in px-backup as we are not claiming the generic backup from stork.

No

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackupschedule.go
+++ b/pkg/applicationmanager/controllers/applicationbackupschedule.go
@@ -47,6 +47,8 @@ const (
 	ObjectLockDefaultIncrementalCount = 5
 	// InsufficientRetentionPeriod status message signifies user has set a improper retention period on the bucket
 	InsufficientRetentionPeriod = "Failed due to insufficient bucket retention period"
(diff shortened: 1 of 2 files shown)


pkg/applicationmanager/controllers/applicationbackupschedule.go
```diff
@@ -47,6 +47,8 @@ const (
 	ObjectLockDefaultIncrementalCount = 5
 	// InsufficientRetentionPeriod status message signifies user has set a improper retention period on the bucket
 	InsufficientRetentionPeriod = "Failed due to insufficient bucket retention period"
+	backupTypeKey               = "portworx.io/" + "backup-type"
```
diptiranjanpx, line 50: Let's make it a single string value.

diptiranjanpx (approved):

pkg/applicationmanager/controllers/applicationbackupschedule.go
siva-portworx: Fixed

PR #1068, opened 2022-04-11
pb-2037: Making snapshot timeout configurable

lalat-das:
Added a configMap which can alter the snapshot timeout period this helps
in addressing OCS cluster issues wherein it takes genuinely more time.

**What type of PR is this?**Bug
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: In certain OCP version the local snapshot takes a lot of time to complete it hence we need to keep the timeout configurable. So that user can experiment with it based on the OCP version and types of Node config it is running the OCP setup. 

**Is a release note needed?**: NO

**Does this change need to be cherry-picked to a release branch?**: 2.10

** UnitTesting**
The configMap feild in kdmp-config looks as below with name SNAPSHOT_TIMEOUT
rror xcurs
Currently set it as 3second to simulate the timeout scenario, because on OCP 4.9 and with a PVC of 100GB and data worth of 50GB in it takes 14 seconds. When this bug was raised in OCP 4.8.0 it took more than 17 minutes.
image

OCP backup taken as below 
Create Backup

Backup failed with a configured timeout 
Pasted Graphic 4

IN CLUSTER

### DIFF
--- a/pkg/snapshotter/snapshotter_csi.go
+++ b/pkg/snapshotter/snapshotter_csi.go
@@ -13,6 +13,7 @@ import (
 	kSnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1beta1"
 	kSnapshotClient "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
+	"github.com/libopenstorage/stork/drivers"
 	storkapi "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/crypto"
 	"github.com/libopenstorage/stork/pkg/k8sutils"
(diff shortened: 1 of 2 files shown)


lalat-das:
I will raise the review request once finish my Unit testing. For now it is parked in PR pipeline. Currently Waiting for an OCS setup to verify the time out in action.

pkg/snapshotter/snapshotter_csi.go
```diff
@@ -283,6 +286,14 @@ func (c *csiDriver) SnapshotStatus(name, namespace string) (SnapshotInfo, error)
 	if contentName != "" {
 		vscError = c.getSnapshotContentError(contentName)
 	}
+
+	snapshotTimeout, err := getSnapshotTimeout()
```
prashanthpx, line 290: If this fails, should we log a msg  and fall back to the default one and not fail the backup?

lpitstickpstg: Is there somewhere we can indicate this to the customer? I agree breaking all snapshots backups is a drastic reaction to a bad value. If the customer is having issues still with the default value they should still notice, and the damage would be contained

lalat-das: @lpitstickpstg Other than logging to stork POD logs we don't have any other way to report it back. And the default value is 5 mins hard-coded , unless someone tampers that strings we have no way of failing in here.  If developer tampers that value it eventually fails in switch case.
Hence I feel we can adopt the default value in case the user provided timeout  value is a malformed one.
@prashanthpx @lpitstickpstg Kindly let me know your opinion

lalat-das:
> I will raise the review request once finish my Unit testing. For now it is parked in PR pipeline. Currently Waiting for an OCS setup to verify the time out in action.

Updated the PR with Unit test cases executed on OCP 4.9

lpitstickpstg (approved):

PR #1080, opened 2022-04-23
pb-2330: Made fix to support image name to have custom repo as well for cmdexecutor image

siva-portworx:
**What type of PR is this?**
>bug

```
pb-2330: Made fix to support image name to have custom repo as well for cmdexecutor image
```
no
Planning to add seperate documentation for the custom image in air-gapped environment.
2.10

### DIFF
--- a/pkg/k8sutils/k8sutils.go
+++ b/pkg/k8sutils/k8sutils.go
@@ -180,10 +180,10 @@ func GetImageRegistryFromDeployment(name, namespace string) (string, string, err
 	}
 	imageFields := strings.Split(deploy.Spec.Template.Spec.Containers[0].Image, "/")
 	// Here the assumption is that the image format will be <registry-name>/<extra-dir-name>/<repo-name>/image:tag
-	// or <repo-name>/image:tag. If repo name contains any path (<registry-name>/<repo-name>/<extra-dir-name>/image:tag), below logic will not work.
+	// or <repo-name>/image:tag or <registry-name>/<repo-name>/<extra-dir-name>/image:tag).
 	// Customer might have extra dirs before the repo-name as mentioned above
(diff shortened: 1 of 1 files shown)


prashanthpx (approved):

pkg/k8sutils/k8sutils.go
```diff
@@ -180,10 +180,10 @@ func GetImageRegistryFromDeployment(name, namespace string) (string, string, err
 	}
 	imageFields := strings.Split(deploy.Spec.Template.Spec.Containers[0].Image, "/")
 	// Here the assumption is that the image format will be <registry-name>/<extra-dir-name>/<repo-name>/image:tag
-	// or <repo-name>/image:tag. If repo name contains any path (<registry-name>/<repo-name>/<extra-dir-name>/image:tag), below logic will not work.
+	// or <repo-name>/image:tag or <registry-name>/<repo-name>/<extra-dir-name>/image:tag).
 	// Customer might have extra dirs before the repo-name as mentioned above
-	// here minus 2 is for image name and repo name exclusion
-	registryFields := imageFields[0 : len(imageFields)-2]
+	// here minus 1 is for image name
+	registryFields := imageFields[0 : len(imageFields)-1]
```
adityadani, line 186: This is being used from the rules package to determine the image name for cmdexecutor. In this case if a custom registry is not used then the cmdexecutor image would end up being `openstorage/openstorage/cmdexecutor:0.1`

siva-portworx: Good catch, I will make the defaultCmdExecutorImage to be "cmdexecutor:0.1" (https://github.com/libopenstorage/stork/blob/master/pkg/rule/rule.go#L43)
Hope it is fine. I will raise a PR for it.

PR #1088, opened 2022-04-28
pb2328: Support both v1beta1 and v1 for CSI volume snapshot APIs

lalat-das:
pb-2328: Support both v1 & v1beta1 version of volumeSnapshot feature
- Added support for V1 VolumeSnapshot APIs which is GA from
  kubernetes 1.20 version.

**What type of PR is this?** BUG
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: From k8s 1.20 onwards the volumesnapshot feature is promoted to GA, hence we need to support corresponding APIs at the same time maintain backward compatibility with v1beta1 till it is removed in future version. 

**Does this PR change a user-facing CRD or CLI?**: It doesn't change any existing CRDs rather expects user to installs new CRDs with respect to volumesnapshots, volumesnapshotClass, volumesnapshotContent & necessary controllers. Additionally  enable adequate support in code to use them

**Is a release note needed?**: Yes

```release-note
Issue: The stork code will fails to take backup in systems having only v1 based volumesnapshot CRDs & it will expect user to install v1beta1 CRDs as a workaround for k8s beyond 1.20. Additionally in future version of k8s wherein v1beta1 support will be removed this work around will also not work.
User Impact: Backup/restore will fail if Application cluster has only v1 based volumesnapshot CRDs and user reluctant to install v1beta1
Resolution: The code enables support for v1 volumesnapshot CRDs & controllers via this PR. Hence stork need to be upgraded to reflect this change in customer's application cluster for a gracefull backup/restore operation via px-backup 

```

**Does this change need to be cherry-picked to a release branch?**: 2.10

Detailed unit test related screenshots are attached in the bug report **pb-2328** (https://portworx.atlassian.net/browse/PB-2328)

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	kSnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	kSnapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1beta1"
 	kSnapshotClient "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
 	snapv1 "github.com/kubernetes-incubator/external-storage/snapshot/pkg/apis/crd/v1"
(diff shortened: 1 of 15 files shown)


pkg/version/version.go
```diff
@@ -53,6 +60,39 @@ func RequiresV1CSIdriver() (bool, error) {
 	return false, nil
 }
 
+// RequiresV1VolumeSnapshot returns true if V1 version of VolumeSnapshot APIs need to be called
+func RequiresV1VolumeSnapshot() (bool, error) {
```
ggriffiths, line 64: Same comment as your px-backup PR.
https://github.com/portworx/px-backup/pull/949#discussion_r896118336

drivers/volume/csi/csi.go
```diff
@@ -747,22 +965,21 @@ func (c *csi) recreateSnapshotForDeletion(
 	if err != nil {
 		return fmt.Errorf("failed to restore VolumeSnapshotClass for deletion: %s", err.Error())
 	}
-	log.ApplicationBackupLog(backup).Debugf("created volume snapshot class %s for backup %s deletion", vs.Name, snapshotID)
+	//log.ApplicationBackupLog(backup).Debugf("created volume snapshot class %s for backup %s deletion", vs.Name, snapshotID)
```
ggriffiths, line 968: any reason for commenting these out? I think we should uncomment these

drivers/volume/csi/csi.go
```diff
@@ -975,13 +1235,42 @@ func (c *csi) getCSIBackupObject(backupName, backupNamespace string) (*csiBackup
 		return nil, err
 	}
 
-	cbo := &csiBackupObject{}
-	err = json.Unmarshal(backupObjectBytes, cbo)
+	cboCommon := &csiBackupObject{}
+	if c.v1SnapshotRequired {
+		type CsiBackupObjectv1 struct {
```
ggriffiths, line 1240: As mentioned in the px-backup PR, let's add these to `csicommon`

drivers/volume/csi/csi.go
```diff
@@ -1033,59 +1322,106 @@ func (c *csi) findPVCInResources(resources []runtime.Unstructured, pvcName, pvcN
...
-		if k8s_errors.IsAlreadyExists(err) {
-			return vs, nil
+	vs interface{},
+	vsc interface{},
+) (interface{}, error) {
+	if c.v1SnapshotRequired {
+		vs.(*kSnapshotv1.VolumeSnapshot).ResourceVersion = ""
+		vs.(*kSnapshotv1.VolumeSnapshot).Spec.Source.PersistentVolumeClaimName = nil
+		vs.(*kSnapshotv1.VolumeSnapshot).Spec.Source.VolumeSnapshotContentName = &vsc.(*kSnapshotv1.VolumeSnapshotContent).Name
+		vs.(*kSnapshotv1.VolumeSnapshot).Namespace = namespace
```
ggriffiths, line 1361: In general, let's type case once up front and edit that object.

drivers/volume/csi/csi.go
```diff
@@ -1033,59 +1322,106 @@ func (c *csi) findPVCInResources(resources []runtime.Unstructured, pvcName, pvcN
...
+		snapshotHandle := *vsc.(*kSnapshotv1.VolumeSnapshotContent).Status.SnapshotHandle
+
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).ResourceVersion = ""
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.DeletionPolicy = kSnapshotv1.VolumeSnapshotContentDelete
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.Source.VolumeHandle = nil
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.Source.SnapshotHandle = &snapshotHandle
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.VolumeSnapshotRef.Name = vs.(*kSnapshotv1.VolumeSnapshot).Name
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.VolumeSnapshotRef.Namespace = namespace
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.VolumeSnapshotRef.UID = vs.(*kSnapshotv1.VolumeSnapshot).UID
+		vsc.(*kSnapshotv1.VolumeSnapshotContent).Spec.DeletionPolicy = kSnapshotv1.VolumeSnapshotContentRetain
```
ggriffiths, line 1400: Same as above

ggriffiths (changes requested):
Mostly looks good, let's do some of the changes I mentioned in px-backup/stork and I'll take another look after.

drivers/volume/csi/csi.go
lalat-das: Commented it by mistake.

drivers/volume/csi/csi.go
lalat-das: stork doesn't define csiCommon package neither import it, it is defined in px-backup, And possibly we do not wanted create dependency of stork on px-backup. Not sure of the history behind csiCommon though.

I have shifted the definition of structs  to the place where in csiBackupObject{} is defined to make the code looks cleaner. Please let me know ur thoughts on this.

lalat-das:
@ggriffiths I have submittded review comment chnages as a separate commit which I will squash once the PR is approved. Hope you would have got notification.. just reminding in case notification is not recvd

drivers/volume/csi/csi.go
ggriffiths: Sounds good to me.

ggriffiths (approved):

PR #1090, opened 2022-05-03
Skip deleted namespaces from a migration schedule

Rohit-PX:

**What type of PR is this?**
improvement

* In a migration schedule, if  a namespace that is present in the migration schedule is deleted, then we throw an error and fail subsequently triggered migrations
* This results in stoppage of migrations for namespaces that are still present and which the user would like to be migrated

Improvement can be as below:
* Even if a deleted namespace is present in a migration schedule, add warning messages but proceed with migration of other namespaces.
* Add flag in the migration schedule spec to proceed with migrations for other namespaces even if some namespaces are deleted.

```release-note
Issue:  https://portworx.atlassian.net/browse/PWX-23459
User Impact: 
Namespaces that are deleted but are present in the migration schedule namespaces list those migrations were being marked as failed. 
Resolution:
Migration schedule spec will have a new bool field: SkipDeletedNamespaces that will determine if migrations are to be marked as failed in case some namespaces are deleted.
```

2.11

### DIFF
--- a/pkg/apis/stork/v1alpha1/migration.go
+++ b/pkg/apis/stork/v1alpha1/migration.go
@@ -25,6 +25,7 @@ type MigrationSpec struct {
 	PreExecRule                  string            `json:"preExecRule"`
 	PostExecRule                 string            `json:"postExecRule"`
 	IncludeOptionalResourceTypes []string          `json:"includeOptionalResourceTypes"`
+	SkipDeletedNamespaces        *bool             `json:"skipDeletedNamespaces"`
 }
 
 // MigrationStatus is the status of a migration operation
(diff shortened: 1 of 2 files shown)


pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -25,6 +25,7 @@ type MigrationSpec struct {
 	PreExecRule                  string            `json:"preExecRule"`
 	PostExecRule                 string            `json:"postExecRule"`
 	IncludeOptionalResourceTypes []string          `json:"includeOptionalResourceTypes"`
+	SkipDeletedNamespaces        bool              `json:"skipDeletedNamespaces"`
```
lpitstickpstg, line 28: Does this default to False?
The other bools here are *bool

lpitstickpstg (approved):

pkg/apis/stork/v1alpha1/migration.go
Rohit-PX: Yes it will default to false

ram-infrac (approved):

PR #1091, opened 2022-05-03
pb-2371: Handle non-existent bucket scenario for object-lock

lalat-das:
Handle error code "NoSuchBucket" while fetching object lock info for a
AWS S3 based bucket.

**What type of PR is this?**Bug
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: When a user creates a backuplocation object without validation flag set then code fails to fetch the object lock info for the bucket as the bucket wouldn't have created during validation process. Hence "NoSuchBucket" error is discounted in the object lock info fetch logic. It enables stork to create the bucket in later point in time. This is found in integration test for AWS S3 bucket. This issue never occurs when experimented via UI as UI always sets validation flag for the bucket in px-backup.

```release-note
Issue:
User Impact:
Resolution

```

### DIFF
--- a/pkg/objectstore/s3/s3.go
+++ b/pkg/objectstore/s3/s3.go
@@ -78,10 +78,15 @@ func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*common.ObjLockIn
 	out, err := s3.New(sess).GetObjectLockConfiguration(input)
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok {
-			// When a minio server doesn't have object-lock implemented
-			// then Minio backed backuplocation throws this error for
-			// normal buckets too hence we need to ignore this for now.
-			if awsErr.Code() == "ObjectLockConfigurationNotFoundError" || awsErr.Code() == "MethodNotAllowed" {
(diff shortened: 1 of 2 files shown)


pkg/objectstore/s3/s3.go
```diff
@@ -81,7 +81,9 @@ func GetObjLockInfo(backupLocation *stork_api.BackupLocation) (*common.ObjLockIn
 			// When a minio server doesn't have object-lock implemented
 			// then Minio backed backuplocation throws this error for
 			// normal buckets too hence we need to ignore this for now.
-			if awsErr.Code() == "ObjectLockConfigurationNotFoundError" || awsErr.Code() == "MethodNotAllowed" {
+			if awsErr.Code() == "ObjectLockConfigurationNotFoundError" ||
+				awsErr.Code() == "MethodNotAllowed" ||
```
siva-portworx, line 85: Can u please add a comment saying we are getting NoSuchBucket from minio and ObjectLockConfigurationNotFoundError error from aws S3 when the bucket does not exists

siva-portworx (approved):

PR #1100, opened 2022-05-12
linstor: support for WaitingForFirstConsumer

kvaps:
**What type of PR is this?**
bug

PWX-23693 - when using stork scheduler with WaitForFirstConsumer, stork refuses to schedule without a bound PVC, which is waiting for the pod to be scheduled, causing deadlock. This allows stork scheduling to continue, ignoring pending pvc error if it is the above scenario.

no

no

2.11


drivers/volume/linstor/linstor.go
```diff
@@ -268,9 +271,13 @@ func (l *linstor) GetPodVolumes(podSpec *v1.PodSpec, namespace string, includePe
 				continue
 			}
 
-			if pvc.Status.Phase == v1.ClaimPending {
-				return nil, nil, &storkvolume.ErrPVCPending{
-					Name: volume.PersistentVolumeClaim.ClaimName,
+			if !includePendingWFFC && pvc.Status.Phase == v1.ClaimPending {
+				if includePendingWFFC && isWaitingForFirstConsumer(pvc) {
```
lpitstickpstg, line 275: This will never evaluate to true because we only enter if includePendingWFFC is false from the above if statement. Review logic

lpitstickpstg (changes requested):

drivers/volume/linstor/linstor.go
kvaps: Yeah, sorry. Just a little copy/paste bug here 😅

lpitstickpstg:
Thanks for the quick turnaround

PR #1102, opened 2022-05-23
DS-2051: Add `selector` and `namespace` args.

jindrazak:
**What type of PR is this?**
improvement

This PR adds `-selector` and `-namespace` flags. These flags provide user with another way to specify the list of pods where to run the command. 

The `-selector` and `-namespace` args cannot be combined with the `-pod` args as this flag already requires the namespace to be included (`-pod my-namespace/my-pod`).

The behavior of the `selector` is intended to be similar [to the one in `kubectl` itself](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering).

**Example usage**: 

```shell
cmdexecutor -cmd <cmd> -taskid <taskid> -namespace my-namespace -selector label1=value1,label2=value2
```

Yes, without breaking changes. The behavior of `-pod` args remains unchanged.

Improvement: cmd-executor can now select pods to run custom commands using namespace name and label selectors. 

No.

### DIFF
--- a/cmd/cmdexecutor/cmdexecutor.go
+++ b/cmd/cmdexecutor/cmdexecutor.go
@@ -10,7 +10,10 @@ import (
 	"github.com/libopenstorage/stork/pkg/cmdexecutor"
 	"github.com/libopenstorage/stork/pkg/cmdexecutor/status"
 	"github.com/libopenstorage/stork/pkg/version"
+	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/cmdexecutor/cmdexecutor.go
```diff
@@ -50,7 +66,18 @@ func main() {
 	logrus.Infof("Running pod command executor: %v", version.Version)
 	flag.Parse()
 
-	if len(podList) == 0 {
+	podListSpecified := len(podList) > 0
+	selectorAndNamespaceSpecified := labelSelector != "" && namespace != ""
+
+	if labelSelector != "" && namespace == "" {
```
ram-infrac, line 72: nit: can use `selectorAndNamespaceSpecified` directly here

ram-infrac (approved):
lgtm

cmd/cmdexecutor/cmdexecutor.go
jindrazak: Note the `namespace == ""` which is different from the `namespace != ""` used for `selectorAndNamespaceSpecified`.

I can introduce a variable `selectorWithoutNamespaceSpecified := labelSelector != "" && namespace == ""` if you think it might improve readability.

lpitstickpstg (approved):

adityadani (approved):
We need a release note for this:

```release-note
Improvement: cmd-executor can now select pods to run custom commands using namespace name and label selectors. 
```

PR #1108, opened 2022-06-14
add options to collect all network policy

ram-infrac:
**What type of PR is this?**
>improvement

Stork collect only network policies which does not have CIDR set, This PR add option to  collect all network policies regardless of CIDR check

yes, MigrationSchedule/Migration spec now has `skipNetworkPolicyCheck` option to ignore network policies check
```
kind: MigrationSchedule
metadata:
  name: mysql-migration-schedule-weekly-invalid
spec:
  schedulePolicyName: default-interval-policy
  template:
    spec:
      # This should be the name of the cluster pair
      clusterPair: test-dr
      # If set to false this will migrate only the volumes. No PVCs, apps, etc will be migrated
      includeResources: true
      # If set to false, the deployments and stateful set replicas will be set to
      # 0 on the destination. There will be an annotation with
      # "stork.openstorage.org/migrationReplicas" to store the replica count from the source
      startApplications: false
      skipNetworkPolicyCheck: true
      # List of namespaces to migrate
      namespaces:
      - test-dr
```

yes

```release-note
Issue: Stork only collect network policy which does not CIDR set
User Impact: DR system which has same CIDR network may wants to migrate all of the network policy but unable to do so with current stork resource collection restriction
Resolution: Stork now collect all of the network policies if `skipNetworkPolicyCheck` is set in migration specs

```

yes, TBD

### DIFF
--- a/pkg/apis/stork/v1alpha1/migration.go
+++ b/pkg/apis/stork/v1alpha1/migration.go
@@ -21,6 +21,7 @@ type MigrationSpec struct {
 	StartApplications            *bool             `json:"startApplications"`
 	PurgeDeletedResources        *bool             `json:"purgeDeletedResources"`
 	SkipServiceUpdate            *bool             `json:"skipServiceUpdate"`
+	IncludeNetworkPolicyWithCIDR *bool             `json:"includeNetworkPolicyWithCIDR"`
 	Selectors                    map[string]string `json:"selectors"`
 	PreExecRule                  string            `json:"preExecRule"`
 	PostExecRule                 string            `json:"postExecRule"`
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ gocyclo:
 	GO111MODULE=off go get -u github.com/fzipp/gocyclo
 	gocyclo -over 15 $(GO_FILES)
 
-pretest: check-fmt lint vet errcheck staticcheck
+pretest: check-fmt vet errcheck staticcheck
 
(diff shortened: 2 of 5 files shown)


pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -21,6 +21,7 @@ type MigrationSpec struct {
 	StartApplications            *bool             `json:"startApplications"`
 	PurgeDeletedResources        *bool             `json:"purgeDeletedResources"`
 	SkipServiceUpdate            *bool             `json:"skipServiceUpdate"`
+	SkipNetworkPolicyCheck       *bool             `json:"skipNetworkPolicyCheck"`
```
adityadani, line 24: may be we can rename this  to `includeNetworkPolicyWithCIDR` ? Goes well with other configs like `includeResources`. `SkipNetworkPolicyCheck` doesn't indicate what check is being called out in the Migration object

adityadani, line 24: Need to add this field in MigrationSchedule as well.

ram-infrac: migrationschedule points to `migration` spec, there is no need to add it separately  in migrationschedule

ram-infrac: agree, renamed it to `includeNetworkPolicyWithCIDR`

PR #1113, opened 2022-06-30
ProjectMappings for Rancher

adityadani:
**What type of PR is this?**
>feature

PlatformOptions in ClusterPair allow users to specify any configurations required for
kubernetes platform providers like Rancher / EKS / AKS etc

Currently the PlatformOptions are only being used for Rancher. Each platform can define their
own spec within this PlatformOptions.

RancherSpec:
- ProjectMappings: Allows users to configure source and target cluster projectID mappings
so that stork can apply the correct annotations on the destination project where it
is migrating the k8s resources.
- RancherSecret (FUTURE): Proposal for specifying a kubernetes secret to specify rancher api keys that would
be used to send REST APIs to Rancher endpoint for creating/deleting/getting project details

NetworkPolicy / All Affinity referencing objects - Deployment/StatefulSet etc
- Parse a pod spec and if NamespaceSelectors are set, check if there are any rancher project labels
and replace them with the target project ID mapping.

Yes

```release-note
Rancher Project Mapping Support: You can now specify project mappings between the source and destination cluster in the ClusterPair object. Stork will migrate resources between the clusters while translating the project related information by using the project mappings. Currently annotations, labels and namespace selectors on NetworkPolicy and any pod spec are handled by Stork while transforming the resources across projects between two clusters.

```

2.11

### DIFF
--- a/pkg/resourcecollector/resourcecollector.go
+++ b/pkg/resourcecollector/resourcecollector.go
@@ -42,6 +42,8 @@ const (
 	StorkResourceHash = "stork.libopenstorage.org/resource-hash"
 	// SkipModifyResources is the annotation used to skip update of resources
 	SkipModifyResources = "stork.libopenstorage.org/skip-modify-resource"
+	// ProjectMappingsOption is a resource collector option to provide rancher project mappings
+	ProjectMappingsOption = "ProjectMappings"
 	// ServiceKind for k8s service resources
 	ServiceKind = "Service"
 	// NetworkPolicyKind for network policy resources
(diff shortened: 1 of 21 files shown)


pkg/resourcecollector/resourcecollector.go
```diff
@@ -42,6 +42,8 @@ const (
 	StorkResourceHash = "stork.libopenstorage.org/resource-hash"
 	// SkipModifyResources is the annotation used to skip update of resources
 	SkipModifyResources = "stork.libopenstorage.org/skip-modify-resource"
+	// ProjectMappingsOption is a resource collector option to provide rancher project mappings
+	ProjectMappingsOption = "ProjectMappings"
```
ram-infrac, line 46: I don't see this options being passed anywhere in migration controller when clusterpair has Rancher project mapping set, could you check if this is missed

adityadani: Good point! However that made me realize is there is another bug in the way this options are handled

- The resource collector instance is a common instance used in backup and migration controllers. A common instance cannot dictate the options used for different migration or backup objects. 
- Instead of a global map on the resource collector options we should use an actual Options object and pass it as an argument to every API.

I have made that change as a subsequent commit on this PR.

ram-infrac (approved):

PR #1115, opened 2022-07-04
PWX-24431 : Support manually created endpoint resources for migration and backup

ram-infrac:
**What type of PR is this?**
>enhancement

Support manually created endpoint resources for migration and backup

no

yes

```release-note
Issue: Endpoint resources are not migrated by stork
Resolution: stork will now migrate manually created endpoint resources, user can also set annotation `"stork.libopenstorage.org/include-resource"` to collect auto-created endpoint

```

2.11.2

### DIFF
--- a/pkg/resourcecollector/endpoint.go
+++ b/pkg/resourcecollector/endpoint.go
@@ -1,0 +1,28 @@
+package resourcecollector
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func (r *ResourceCollector) endpointsToBeCollected(
+	object runtime.Unstructured,
+) (bool, error) {
+	var endpoint v1.Endpoints
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.UnstructuredContent(), &endpoint); err != nil {
+		return false, fmt.Errorf("error converting to endpoint: %v", err)
+	}
+
+	if endpoint.Annotations != nil {
+		if _, ok := endpoint.Annotations["last_applied_annotation"]; ok {
+			return true, nil
+		}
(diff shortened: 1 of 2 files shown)


lpitstickpstg:
We should document explicitly how `"stork.libopenstorage.org/include-resource"` is used. So users know it's only relevant to endpoints (at this time). They may try to use it for other resources

pkg/resourcecollector/endpoint.go
```diff
@@ -0,0 +1,36 @@
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.UnstructuredContent(), &endpoint); err != nil {
+		return false, fmt.Errorf("error converting to endpoint: %v", err)
+	}
+	if endpoint.Annotations != nil {
+		// collect manually created endpoint resources
+		if _, ok := endpoint.Annotations[v1.LastAppliedConfigAnnotation]; !ok {
+			return false, nil
+		}
+		// collect endpoint which has include-resources annotation applied
+		if _, ok := endpoint.Annotations[IncludeResources]; ok {
```
lpitstickpstg, line 23: Also check for skip resources?

ram-infrac: we do have check for ` skip resource` for all of the resources in `objectsToBeCollected`

ram-infrac:
good point, let me update release note section

pkg/resourcecollector/endpoint.go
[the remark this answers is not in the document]
adityadani: Also the selector check should be done first. So basically:
- 1st check: IncludeResources
- 2nd check: selectors (doing this as a 2nd check ensures if someone applied  an Endpoints object with selectors using `kubectl` )
- 3rd check: last applied configuration

adityadani (approved):
lgtm after comments are addressed

PR #1117, opened 2022-07-06
PWX-24637: mutate the virt-launcher container to intercept statfs()

pureneelesh:
**What type of PR is this?**
>feature

Live migration of KubeVirt VM fails if the VM is using
a bind-mounted sharedv4 volume. The root cause is that libvirt
uses a statfs() call to check the file system type and
incorrectly concludes that the volume is not shared.

This patch addresses this problem as described below.

We use a shared library px_statfs.so that intercepts libvirt's statfs call.
If the input path is backed by a PX volume, we change the file system
type returned by the real statfs call. This shared library is bundled with
the stork container.

If a virt-launcher pod is being created and is using a PX volume, stork's
mutating webhook creates a ConfigMap in the pod's namespace.

This configMap has 2 keys that represent the 2 files that we want to inject
into the virt-launcher container's /etc directory:

  - ld.so.preload
  - px_statfs.so

The stork webhook then mutates the virt-launcher pod's spec to mount
the configMap as a volume and inject the 2 files above in /etc dir on the
container's file system. This makes linux load px_statfs.so in the libvirt
process that is running inside the virt-launcher container and intercept
libvirts' statfs() call.

Signed-off-by: Neelesh Thakur <neelesh.thakur@purestorage.com>

no

```release-note
Issue: Kubevirt live migration fails if VM is using Portworx volume.
User Impact: Can't use live migration feature of Kubevirt.
Resolution: The problem was fixed by intercepting libvirt's statfs() call.
```

Yes. 2.12.0.

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -3,9 +3,12 @@ package portworx
 import (
 	"context"
 	"encoding/csv"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"math"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -167,6 +170,12 @@ const (
 	pureBackendParam     = "backend"
 	pureFileParam        = "pure_file"
 	pureBlockParam       = "pure_block"
+
+	statfsSOName              = "px_statfs.so"
+	statfsSODirInStork        = "/"
+	statfsSODirInVirtLauncher = "/etc"
+	statfsConfigMapName       = "px-statfs"
+	statfsVolName             = "px-statfs"
 )
 
 type cloudSnapStatus struct {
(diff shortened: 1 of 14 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -3956,4 +4068,11 @@ func init() {
 	if err := storkvolume.Register(storkvolume.PortworxDriverName, p); err != nil {
 		logrus.Panicf("Error registering portworx volume driver: %v", err)
 	}
+
+	soPath := path.Join(statfsSODirInStork, statfsSOName)
+	statfsSOContents, err = ioutil.ReadFile(soPath)
```
lpitstickpstg, line 4073: Why read this in as a global var, keeping it in memory the whole time instead of just when we need it?

drivers/volume/portworx/portworx.go
```diff
@@ -3946,6 +3958,106 @@ func (p *portworx) CleanupRestoreResources(*storkapi.ApplicationRestore) error {
...
+
+	if err := p.createStatfsConfigMap(podNamespace); err != nil {
+		return nil, fmt.Errorf("failed to create config map: %w", err)
+	}
+
+	podName := pod.Name
+	if podName == "" {
+		podName = pod.GenerateName + "***"
+	}
+	logrus.Infof("Mutating virt-launcher pod %s/%s", podNamespace, podName)
```
lpitstickpstg, line 4034: Maybe some expansion on how/why it's mutating for someone seeing it in the logs

drivers/volume/portworx/px-statfs/px_statfs.c
```diff
@@ -0,0 +1,180 @@
...
+int statfs(const char *path, struct statfs *buf)
+{
+    int rc = -1;
+
+    // Perform the actual system call
+    rc = real_statfs(path, buf);
+
+    if (rc == 0)
+    {
+        // TODO: do we need to check buf->f_type in (ext4 | XFS) ?
```
lpitstickpstg, line 152: todo comment

lpitstickpstg (commented):
Go code looks good.

Does setting `webhook-controller: false` disable this? Is that a problem?

pureneelesh:
> Go code looks good.
> 
> Does setting `webhook-controller: false` disable this? Is that a problem?

I suppose. I am thinking that we should honor that setting for all mutations.

pkg/webhookadmission/webhook.go
```diff
@@ -161,13 +182,27 @@ func (c *Controller) processMutateRequest(w http.ResponseWriter, req *http.Reque
 		}
 	} else {
 		// create patch
-		log.Debugf("Updating scheduler to stork for Resource:%s, Name: %s, Namespace:%s", arReq.Kind.Kind, resourceName, arReq.Namespace)
-		patch := createPatch(schedPath)
+		log.Debugf("Updating scheduler to stork for Resource:%s, Name: %s, Namespace:%s",
+			arReq.Kind.Kind, resourceName, arReq.Namespace)
+
+		patches = append(patches, k8sutils.JSONPatchOp{
```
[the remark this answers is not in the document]
pureneelesh: createPatch was doing more than what we need here. We want to merge the other patches before marshalling into []byte. Anyway, let me try to resurrect createPatch().

pkg/webhookadmission/webhook.go
```diff
@@ -161,13 +182,27 @@ func (c *Controller) processMutateRequest(w http.ResponseWriter, req *http.Reque
...
+			Operation: "replace",
+			Path:      schedPath,
+			Value:     []byte(strconv.Quote(storkScheduler)),
+		})
+		patchBytes, err := json.Marshal(&patches)
+		if err != nil {
+			log.Errorf("Could not marshal the patch object: %v", err)
+			c.Recorder.Event(webhookConfig, v1.EventTypeWarning, "could not marshal the patch object", err.Error())
+			http.Error(w, "Encode error", http.StatusInternalServerError)
+			return
```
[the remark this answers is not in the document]
pureneelesh: Currently, the patches are generated only if storkResource is true. Even in future, I am thinking that we don't want to mutate a resource which is not a storkResource. WDYT?

drivers/volume/portworx/portworx.go
pureneelesh: Didn't want to read the same file over and over unnecessarily. Do you see a problem here?

lpitstickpstg: Are kubevirt pods common enough for us to keep in in memory for every single instance of the portworx driver? If someone is using them do they start new pods very frequently? If statfsSOContents is small I guess it's not a big deal, but I'm suspicious of the pattern. Reading a file occasionally isn't a big deal, but ballooning memory footprint could be.

lpitstickpstg:
> I suppose. I am thinking that we should honor that setting for all mutations.

For scheduling they can use `schedulerName: stork`, as long as that workaround works, and is documented somewhere I think that makes sense

pkg/webhookadmission/webhook.go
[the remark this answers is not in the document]
lpitstickpstg: I agree with that.

What I mean here is that our old approach would log an error if there was an issue marshalling, but send the admission response anyways. Now if either patch (kubevirt, sched) marshal fails we return without sending an admissionResponse at all. 

Do we want to move to this hard failure behavior from soft failure?

drivers/volume/portworx/portworx.go
pureneelesh: Yeah, I see your point. Will change it.

drivers/volume/portworx/px-statfs/px_statfs.c
```diff
@@ -0,0 +1,180 @@
...
+        int rc2 = get_major_minor_nr(path, &major, &minor);
+        if (rc2 != 0)
+        {
+            return rc;
+        }
+
+        int is_pxd = is_px_device(major, minor);
+        if (is_pxd == 1)
+        {
+            buf->f_type = 0x6969; // NFS_SUPER_MAGIC
```
prabirpaul, line 164: Is the incoming argument always a shared mountpath?

drivers/volume/portworx/px-statfs/px_statfs.c
prabirpaul: I think we can add this check as another precondition. PX allows only ext4/xfs formatted volumes.

drivers/volume/portworx/portworx.go
```diff
@@ -3946,6 +3955,113 @@ func (p *portworx) CleanupRestoreResources(*storkapi.ApplicationRestore) error {
...
+	cm := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      statfsConfigMapName,
+			Namespace: cmNamespace,
+		},
+		BinaryData: map[string][]byte{
+			statfsSOName: statfsSOContents,
+		},
+		Data: map[string]string{
+			"ld.so.preload": path.Join(statfsSODirInVirtLauncher, statfsSOName),
```
prabirpaul, line 4052: Where is the px_statfs.so getting injected into ld.so.preload?

drivers/volume/portworx/px-statfs/px_statfs.c
pureneelesh: > PX allows only ext4/xfs formatted volumes.
Why do we need to add the check, then? If we support another file system in future, we will have to fix that check. Would be an unnecessary busy-work in future.

drivers/volume/portworx/px-statfs/px_statfs.c
pureneelesh: Kubevirt has a check that prevents "live migration" if the VM is not using a shared volume. Therefore, we should not get here.

In addition, I don't think we allow mounting a non-shared volume on 2 hosts at the same time. Therefore, I am guessing that this is safe for non-shared volumes as well.

drivers/volume/portworx/portworx.go
pureneelesh: We add this configMap as a volume to the pod and inject the 2 keys in this configMap as files into the container's `/etc` dir. The configmap key becomes the name of the file, and value becomes the content.

See: https://docs.google.com/document/d/1HR3U3KWgite71jj3o5t44gvgiCG-Pc2veWzdWXbKtqU/edit

adityadani (approved):
looks good!

drivers/volume/portworx/px-statfs/px_statfs.c
prabirpaul: That's fair. The device being a px device is a good enough check, fs on it is not really relevant for the work being done here.

prabirpaul (approved):

PR #1132, opened 2022-08-02
Pwx 18788

lpitstickpstg:
**What type of PR is this?**
improvement

PWX-18788 stork make tls connection to px 

no

no

no


drivers/volume/portworx/portworx.go
```diff
@@ -72,7 +73,7 @@ import (
 const (
 
 	// default API port
-	defaultAPIPort = 9001
+	defaultAPIPort = 9023
```
ram-infrac, line 76: why  the change of default rest api for PX ? If it's for tls you will need to make changes in `BuildClientsEndpoints` method to recognise this for from px-api service

ram-infrac (commented):
Kindly change header of PR to match with PR purpose

drivers/volume/portworx/portworx.go
lpitstickpstg: Endpoints change is here: https://github.com/libopenstorage/openstorage/pull/2092/files

I've rolled this back, it was a holdover from the old PR I picked up. I'll deal with cluster pairs in a different PR/ticket PWX-26101

`defaultAPIPort` is only used for CreatePair if not overridden, how is this port used in a clusterPair? I don't know much about the connectoin between 2 clusters.

ggriffiths (approved):
Looks good

adityadani (approved):
Overall looks good to me. Can you run a stork integration job on your branch to ensure there are no regressions. You can use this job as a reference - https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%20master/job/stork-master-px-2.10-all-tests/68/

PR #1144, opened 2022-08-18
Update object to be migrated as per resource transformation specs

ram-infrac:
**What type of PR is this?**
>feature

Add updating resource as per transformation specs

yes 
Migration spec now accept `updateResourceSpecs` param to update resource as per transformation rule before migrating 
resources.
```
kind: MigrationSchedule
metadata:
  name: mysql-migration-transform-interval
spec:
  schedulePolicyName: migrate-every-5m
  template:
    spec:
      # This should be the name of the cluster pair
      clusterPair: remoteclusterpair
      # If set to false this will migrate only the volumes. No PVCs, apps, etc will be migrated
      includeResources: true
      # If set to false, the deployments and stateful set replicas will be set to
      # 0 on the destination. There will be an annotation with
      # "stork.openstorage.org/migrationReplicas" to store the replica count from the source
      startApplications: false
      # Update service resource as per transformation specs
      transformSpecs:
      - mysql-service-transform
      namespaces:
      - mysql-1-pvc-mysql-migration-transform-interval

  ```
Transformation CR :
```
kind : ResourceTransformation
metadata:
   name: test
specs:
  transformSpecs:
  - paths:
      - path: "spec.type"
        value: "LoadBalancer"
        type: "string"
        operation: "add"
      - path: "metadata.labels"
        value: "handler: project1,app:mysql-2"
        type: "keypair"
        operation: "modify"
    resource: "/v1/Service"
```
yes
```release-note
Migration can now accept resource transformation rules by which customer can modify resources before applying it on DR cluster. This transformation rules has to be created as part of `ResourceTransformation` CR and passed in to migration specification using `updateResourceSpecs` option.

```

Note: 

Jenkins for regression : https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%202.12-dev%20%20C7/job/2.12-px-2.11-migration-csi/135/consoleFull

```
time="2022-08-26T15:49:44Z" level=info msg="[mysql-1-pvc] Destroyed PVC: mysql-data"
=== RUN   TestMigration/testMigrationFailoverFailback
--- PASS: TestMigration (340.97s)
    --- PASS: TestMigration/testMigration (337.92s)
        --- PASS: TestMigration/testMigration/transformResourceTest (334.91s)
    --- PASS: TestMigration/testMigrationFailoverFailback (0.04s)
PASS

DONE 6 tests in 363.544s
```

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -373,6 +373,7 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 			}
 			// Make sure if transformation CR is in ready state
 			if len(migration.Spec.TransformSpecs) != 0 {
+				// Check if multiple transformation specs are provided
 				if len(migration.Spec.TransformSpecs) > 1 {
 					errMsg := fmt.Sprintf("providing multiple transformation specs is not supported in this release %v, err: %v", migration.Spec.TransformSpecs, err)
 					log.MigrationLog(migration).Errorf(errMsg)
@@ -386,9 +387,10 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 					}
 					return nil
 				}
+				// verify if transform specs are created
 				resp, err := storkops.Instance().GetResourceTransformation(migration.Spec.TransformSpecs[0], ns)
-				if err != nil && !errors.IsNotFound(err) {
-					errMsg := fmt.Sprintf("unable to retrive transformation %s, err: %v", migration.Spec.TransformSpecs, err)
+				if err != nil {
+					errMsg := fmt.Sprintf("unable to retrieve transformation %s, err: %v", migration.Spec.TransformSpecs, err)
 					log.MigrationLog(migration).Errorf(errMsg)
 					m.recorder.Event(migration,
 						v1.EventTypeWarning,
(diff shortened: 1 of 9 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1240,6 +1287,7 @@ func (m *MigrationController) prepareApplicationResource(
 	migration *stork_api.Migration,
 	clusterPair *stork_api.ClusterPair,
 	object runtime.Unstructured,
+	resPatch []stork_api.TransformResourceInfo,
```
adityadani, line 1290: Don't you need to use this while applying?

pkg/migration/controllers/migration.go
```diff
@@ -364,7 +366,37 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 						err.Error())
 					err = m.updateMigrationCR(context.Background(), migration)
 					if err != nil {
-						log.MigrationLog(migration).Errorf("Error updating")
+						log.MigrationLog(migration).Errorf("Error updating CR, err: %v", err)
+					}
+					return nil
+				}
+			}
+			// Make sure if transformation CR is in ready state
```
adityadani, line 374: What if we did the dryRun and after a that a few resources got created? The migration could still potentially if a transform rule does not apply to this new resource.

pkg/resourcecollector/resourcetransformation.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	logrus.Infof("resource patch : %v", patch)
+	return patch, nil
+}
+
+// this method transform object as per resource transformation specified in each namespaces
+func TransformResources(
+	object runtime.Unstructured,
+	resPatch []stork_api.TransformResourceInfo,
+	name, namespace string,
```
adityadani, line 45: nit: lets rename these to objName and objNamespace. Also we should be able to determine the name and namespace from the unstructured object right?

pkg/resourcecollector/resourcetransformation.go
```diff
@@ -0,0 +1,138 @@
...
+		return patch, nil
+	}
+	for _, namespace := range namespaces {
+		resp, err := storkops.Instance().GetResourceTransformation(transformName, namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			logrus.Infof("not found in namespace: %v", transformName)
+			return nil, err
+		}
+		resMap := make(map[string][]stork_api.TransformResourceInfo)
+		for _, resource := range resp.Status.Resources {
```
adityadani, line 32: Pre-determining which resources to transform when we are validating Transform CR won't work if new objects get added to the namespaces which are getting migrated. I think we should simply validate the CR and check if the input arguments are correct. Once the migration triggers we could do this:
- collect all the objects
- dry run apply all the objects using the transformation
- fail or proceed the migration based on the dry run

pkg/resourcecollector/resourcetransformation.go
```diff
@@ -0,0 +1,138 @@
...
+			resMap[resource.Kind] = append(resMap[resource.Group], *resource)
```
adityadani, line 33: shouldn't this be?
```suggestion
			resMap[resource.Kind] = append(resMap[resource.Kind], *resource)
```

adityadani (commented):
I have a few comments across both the PRs. Lets merge all those PRs into this one.

pkg/resourcecollector/resourcetransformation.go
ram-infrac: thats good point. Now we are deleting resource and setting transform CR status to `Initial` that way it will re-run all checks before we start migration

pkg/migration/controllers/migration.go
```diff
@@ -366,11 +366,79 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
...
+						err.Error())
+					err = m.updateMigrationCR(context.Background(), migration)
+					if err != nil {
+						log.MigrationLog(migration).Errorf("Error updating CR, err: %v", err)
+					}
+					return nil
+				}
+				resp, err := storkops.Instance().GetResourceTransformation(migration.Spec.TransformSpecs[0], ns)
+				if err != nil && !errors.IsNotFound(err) {
+					errMsg := fmt.Sprintf("unable to retrive transformation %s, err: %v", migration.Spec.TransformSpecs, err)
```
lpitstickpstg, line 391: spelling: retrieve

pkg/migration/controllers/migration.go
```diff
@@ -366,11 +366,79 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
...
+					if err != nil && !errors.IsNotFound(err) {
+						errMsg := fmt.Sprintf("Error updating transformation CR: %v", err)
+						log.MigrationLog(migration).Errorf(errMsg)
+						m.recorder.Event(migration,
+							v1.EventTypeWarning,
+							string(stork_api.MigrationStatusFailed),
+							err.Error())
+						err = m.updateMigrationCR(context.Background(), migration)
+						if err != nil {
+							if err != nil {
```
lpitstickpstg, line 419: Double err check

pkg/migration/controllers/migration.go
```diff
@@ -366,11 +366,79 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
...
+					if err := storkops.Instance().ValidateResourceTransformation(transform.Name, ns, 1*time.Minute, 5*time.Second); err != nil {
+						errMsg := fmt.Sprintf("transformation %s is not in ready state: %s", migration.Spec.TransformSpecs, resp.Status.Status)
+						log.MigrationLog(migration).Errorf(errMsg)
+						m.recorder.Event(migration,
+							v1.EventTypeWarning,
+							string(stork_api.MigrationStatusFailed),
+							err.Error())
+						err = m.updateMigrationCR(context.Background(), migration)
+						if err != nil {
+							if err != nil {
```
lpitstickpstg, line 434: double err check

pkg/resourcecollector/resourcetransformation.go
```diff
@@ -0,0 +1,137 @@
...
+					}
+					err := unstructured.SetNestedField(content, value, strings.Split(path.Path, ".")...)
+					if err != nil {
+						logrus.Errorf("Unable to apply patch path %s on resource kind: %s/,%s/%s,  err: %v", path, patch.Kind, patch.Namespace, patch.Name, err)
+						return err
+					}
+				}
+			}
+			object.SetUnstructuredContent(content)
+			logrus.Infof("updated resource of kind %v with patch , resource: %v", patch.Kind, object)
```
lpitstickpstg, line 112: nit: include Operation?

lpitstickpstg (approved):

PR #1159, opened 2022-09-19
PWX-26900: Set the resource migration finish timestamp once pruning is done

adityadani:
**What type of PR is this?**
bug

Stork would report incorrect resource migration time since it did not consider the time taken to prune deleted resources from the target cluster.

No

Yes
```release-note
Issue: Stork reported incorrect resource migration elapsed time when `purgeDeletedResources` is set on Migration.
User Impact: Resource migration elapsed time would be lesser than the actual time taken to complete the resource migraiton.
Resolution: Stork calculates the time taken to perform resource migration once it is done purging the deleted resources.

```

Yes - 2.12

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -1021,15 +1021,6 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration, v
 		return err
 	}
 
-	migration.Status.ResourceMigrationFinishTimestamp = metav1.Now()
-	migration.Status.Stage = stork_api.MigrationStageFinal
-	migration.Status.Status = stork_api.MigrationStatusSuccessful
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1572,11 +1573,11 @@ func (m *MigrationController) applyResources(
 
 			res.ResourceVersion = ""
 			// if crds is applied as v1beta on k8s version 1.16+ it will have
-			// preservedUnkownField set and api version converted to v1 ,
+			// preservedUnknownField set and api version converted to v1 ,
```
lpitstickpstg, line 1576: nit: plural?

lpitstickpstg (approved):

PR #1170, opened 2022-09-30
Update migration resources with resource transformation events

ram-infrac:
**What type of PR is this?**
>enhancement

Update resource transformation CR to accept clusterpair info, so that transformation can be dry-run on remote cluster pointed by ClusterPair object

no

yes, should be part of Parent resource transformation CR

2.12


pkg/migration/controllers/migration.go
```diff
@@ -402,6 +402,10 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 					}
 					return nil
 				}
+
+				// to run transformation validation on remote cluster
+				resp.Spec.ClusterPairName = migration.Spec.ClusterPair
```
adityadani, line 407: I think we should probably avoid doing a dry-run during Migration. This adds to more "GetResourceForType" k8s API calls. If the initial validation was successful, then lets just try the transformation and if it fails then fail the Migration.  Any way if the Transform CR dry-run validation fails here we fail the migration here.

adityadani (changes requested):

pkg/migration/controllers/migration.go
```diff
@@ -402,26 +402,8 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
...
-						err.Error())
-					err = m.updateMigrationCR(context.Background(), migration)
-					if err != nil {
-						log.MigrationLog(migration).Errorf("Error updating CR, err: %v", err)
-					}
-					return nil
-				}
 				// wait for re-run of dry-run resources
-				if err := storkops.Instance().ValidateResourceTransformation(transform.Name, ns, 1*time.Minute, 5*time.Second); err != nil {
+				if err := storkops.Instance().ValidateResourceTransformation(resp.Name, ns, 1*time.Minute, 5*time.Second); err != nil {
```
adityadani, line 406: nit: change the comment since we dont run dry-run here anymore

pkg/migration/controllers/resourcetransformation.go
```diff
@@ -199,6 +182,43 @@ func (r *ResourceTransformationController) validateSpecPath(transform *stork_api
...
 	resourceCollectorOpts := resourcecollector.Options{}
+	config, err := clientcmd.BuildConfigFromFlags("", "")
+	if err != nil {
+		return err
+	}
+	localInterface, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	remoteOps, err := coreops.NewForConfig(config)
```
adityadani, line 193: nit: rename this to localOps

adityadani (approved):
lgtm

PR #1178, opened 2022-10-06
pb-3138: Resetting the serviceAccountUID annotation in the secret.

siva-portworx:
**What type of PR is this?**
>bug

```
    pb-3138: Resetting the serviceAccountUID annotation in the secret.

            - When applying the service account, it will get a new UID.
            - This UID need to be updated in the secret. Resetting the old
              UID to empty string, so that k8s will update the new UID of service
              account.
```

No

```release-note
Issue: The secret token for the service account was not getting applied during restore.
User Impact: Application bring fails some time, if they look for the old secret of the service account token
Resolution: Resetting the service account UID annotation of the secret, so that k8s will update the new UID of the restored service account.

```

2.12

**Testing:**
Tested the fix on the QA setup, where minio app was failing.

### DIFF
--- a/pkg/resourcecollector/secret.go
+++ b/pkg/resourcecollector/secret.go
@@ -8,6 +8,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	serviceAccountUIDKey = "kubernetes.io/service-account.uid"
+)
+
 func (r *ResourceCollector) secretToBeCollected(
 	object runtime.Unstructured,
 ) (bool, error) {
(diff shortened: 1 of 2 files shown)


ram-infrac (approved):

pkg/resourcecollector/secret.go
```diff
@@ -26,3 +30,30 @@ func (r *ResourceCollector) secretToBeCollected(
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.UnstructuredContent(), &secret); err != nil {
+		logrus.Errorf("Error converting Secret object %v: %v", object, err)
+		return err
+	}
+	// Reset the " kubernetes.io/service-account.uid" annotation,
+	// so that it will update the uid of the newly created SA after restoring
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.UnstructuredContent(), &secret)
+	if err != nil {
+		return err
+	}
```
adityadani, line 47: You are already doing this on line 38-40

PR #1180, opened 2022-10-07
pb-3123: Added logic to trim the group name, if crds have last three part of groups in common.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3123: Added logic to trim the group name, if crds have last three
    part of groups in common.

            - For example, some of the crds have group as agent.k8s.elastic.co, apm.k8s.elastic.co and beat.k8s.elastic.co
            - These case, we group them while taking backup as they have last three part og the group name in common.
```

no

```release-note
Issue: Some of CRDs, which have last three parts of groups as common name were not included in the backup and that makes app not to come up properly after restore.
User Impact: Missing of the CRDs in the backup makes app not to come up properly after restore.
Resolution: Added logic to include the CRDs that have last three part of the group name as common.
```

2.12


ram-infrac (approved):
lgtm

PR #1189, opened 2022-10-17
PB-3150: Handle GCE snapshot NotFound error in DeleteBackup

adityadani:
**What type of PR is this?**
>bug

If a DeleteBackup is invoked on the GCE driver in succession, the backup deletion will fail.
- Do not fail CancelBackup / DeleteBackup API if a snapshot was not found.

No

```release-note
Issue: A repetitive call to delete a backup of GCE persistent disks would silently fail the backup deletion
User Impact: After deleting a backup, the kubernetes resource objects would not get deleted from the objectstore.
Resolution: Stork now handles NotFound error if returned by GCE on snapshot deletion.
```

Yes - 2.12

### DIFF
--- a/drivers/volume/gcp/gcp.go
+++ b/drivers/volume/gcp/gcp.go
@@ -347,6 +347,12 @@ func (g *gcp) DeleteBackup(backup *storkapi.ApplicationBackup) (bool, error) {
 		}
 		_, err := service.Snapshots.Delete(vInfo.Options["projectID"], vInfo.BackupID).Do()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


ram-infrac (approved):

siva-portworx:
This is not a issue on AWS and Azure it is already taken care to handle the notfound  error rt? But ticket says it fails in other cloud provider as well.

siva-portworx (approved):

adityadani:
The jobs on EKS failed with the similar signature, but the backup deletion was failing because of a AWS permission issue. So this bug only exists for GKE.

PR #1191, opened 2022-10-18
pb-3162: Resetting secret token data for service account token before restoring

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3162: Resetting secret token data for service account token before restoring
```

no

```release-note
Issue: Service account secret was not updated with latest token value.
User Impact: Token value in the service account secret is old value. Using that token will with the restored application.
Resolution: Resetting the token to empty, so that new token will be update while restoring based on new service account on restore cluster.

```

2.12

### DIFF
--- a/pkg/resourcecollector/secret.go
+++ b/pkg/resourcecollector/secret.go
@@ -44,6 +44,10 @@ func (r *ResourceCollector) prepareSecretForApply(
 	if secret.Annotations != nil {
 		if _, ok := secret.Annotations[serviceAccountUIDKey]; ok {
 			secret.Annotations[serviceAccountUIDKey] = ""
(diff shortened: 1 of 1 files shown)


ram-infrac (approved):

pkg/resourcecollector/secret.go
```diff
@@ -46,6 +46,10 @@ func (r *ResourceCollector) prepareSecretForApply(
 			secret.Annotations[serviceAccountUIDKey] = ""
 		}
 	}
+	// Reset the secret token data to empty, so that new service account token will be updated by k8s, during restore.
```
ram-infrac, line 49: lets empty token only if serviceaccountUID key is set ?

siva-portworx: Fixed

PR #1192, opened 2022-10-20
Handle checks for resource transformation spec paths

ram-infrac:
**What type of PR is this?**
>bug

NestedMap does not return error if spec path is not found in k8s object, causing nil entry to map. This PR add checks around for spec path provided in RT

no

no

2.12


siva-portworx (approved):
LGTM

PR #1193, opened 2022-10-20
pb-3159: Add subpath and nfs mount option in BL

lalat-das:
The NfsConfig structure need to have subpath and mountOption to store the value passed from px-backup as part of backuplocation object.

**What type of PR is this?** Feature
>bug
feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: Adds sub path and mount option to nfsConfig so that NFS based backuplocation can propagate the same for further processing.

```release-note
Issue:
User Impact:
Resolution

```

### DIFF
--- a/pkg/apis/stork/v1alpha1/backuplocation.go
+++ b/pkg/apis/stork/v1alpha1/backuplocation.go
@@ -127,7 +127,9 @@ type GoogleConfig struct {
 }
 
 type NfsConfig struct {
-	NfsServerAddr string `json:"nfsServerAddr"`
+	ServerAddr  string `json:"serverAddr"`
+	SubPath     string `json:"subPath"`
+	MountOption string `json:"mountOption"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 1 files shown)


pkg/apis/stork/v1alpha1/backuplocation.go
```diff
@@ -128,6 +128,8 @@ type GoogleConfig struct {
 
 type NfsConfig struct {
 	NfsServerAddr string `json:"nfsServerAddr"`
```
siva-portworx, line 130: Minor comment, it is better to change NfsServerAddr to ServerAddr :)

siva-portworx (approved):

pkg/apis/stork/v1alpha1/backuplocation.go
lalat-das: I tried that but that is causing a circular dependencies... Like if I change that then in the utils file of kdmp expects in this file the variable to be Nfsbased
../../vendor/github.com/portworx/kdmp/pkg/drivers/utils/utils.go:362:73: backupLocation.Location.NfsConfig.ServerAddr undefined (type *"github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1".NfsConfig has no field or method ServerAddr)

Now I need to push this stork path forcefully if I chnage this variable and then vendor it kdmp to mitigatecompilation  issue in kdmp then vendor it back in stork  for stork to compile fine..So  someone has to merge with a broken patch..

lalat-das: It can be done but there will be commit for which stork branch will not compile.. Which I thought i should aoid

PR #1195, opened 2022-10-24
pb-3174: Added csi section in the case portworx volume restore.

siva-portworx:
**What type of PR is this?**
>bug

pb-3174

no

```release-note
Issue: Restore of portworx volume to portworx CSI volume was not adding csi section in the PV spec
User Impact: Some of environment are identifying the restore portworx CSI volume as csi volume.
Resolution: Adding csi section details for portworx csi restored volume.

```

2.12

Testing:
Covered following cases:
1) restoring non-secure portworx volume to secure portworx csi volume.
2) restoring secure portworx volume to secure portworx csi volume.
3) restoring non-secure portworx volume to secure portworx csi volume, where the storage class was having template reference for the secret name and namespace.
4) restoring secure portworx volume to secure portworx csi volume, where the storage class was having template reference for the secret name and namespace.

Note:
```
Note able to test with RKE cluster. Installing portworx operator with px-security enabled is not working. 
Facing below error:
Events:
  Type     Reason                  Age                     From     Message
  ----     ------                  ----                    ----     -------
  Warning  FailedCreatePodSandBox  13s (x2028 over 7h23m)  kubelet  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "1fec59be470f6d45ecf27e720c08a21929104d9fb032378da4a61b4909edcd07": plugin type="calico" failed (add): error getting ClusterInformation: connection is unauthorized: Unauthorized
```

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -166,17 +166,24 @@ const (
 	noToken = "notoken"
 	// templatizedNamespace is the CSI templatized parameter for namespace
 	templatizedNamespace = "${pvc.namespace}"
-	proxyEndpoint        = "proxy_endpoint"
-	proxyPath            = "proxy_nfs_exportpath"
-	pureBackendParam     = "backend"
-	pureFileParam        = "pure_file"
-	pureBlockParam       = "pure_block"
+	// templatizedName is the CSI templatized parameter for name
+	templatizedName  = "${pvc.name}"
+	proxyEndpoint    = "proxy_endpoint"
+	proxyPath        = "proxy_nfs_exportpath"
+	pureBackendParam = "backend"
+	pureFileParam    = "pure_file"
+	pureBlockParam   = "pure_block"
 
 	statfsSOName              = "px_statfs.so"
 	statfsSODirInStork        = "/"
 	statfsSODirInVirtLauncher = "/etc"
 	statfsConfigMapName       = "px-statfs"
 	statfsVolName             = "px-statfs"
+
+	nodePublishSecretName           = "csi.storage.k8s.io/node-publish-secret-name"
+	controllerExpandSecretName      = "csi.storage.k8s.io/controller-expand-secret-name"
+	nodePublishSecretNamespace      = "csi.storage.k8s.io/node-publish-secret-namespace"
+	controllerExpandSecretNamespace = "csi.storage.k8s.io/controller-expand-secret-namespace"
 )
 
 type cloudSnapStatus struct {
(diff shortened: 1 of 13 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -2608,7 +2615,59 @@ func (p *portworx) CancelMigration(migration *storkapi.Migration) error {
...
+				pv.Spec.CSI.ControllerExpandSecretRef.Name = vInfo.PersistentVolumeClaim
+			} else {
+				pv.Spec.CSI.ControllerExpandSecretRef.Name = val
+			}
+			// In the case of portworx volume backup, we will have namespace mapping always.
+			// So no need to check for the template string as we are going to change the value to destination namepsace always.
+			// If user does not change the namespace, the source and destionation namespacein the mapping will be same.
+			if _, ok := sc.Parameters[controllerExpandSecretNamespace]; ok {
+				pv.Spec.CSI.ControllerExpandSecretRef.Namespace = dstNamespace
+			}
```
adityadani, line 2651: Lets set the ControllerExpandSecretRef only if sc.Parameters has it. 
```suggestion
			// Update the controller expand secret

			if val, ok := sc.Parameters[controllerExpandSecretName]; ok {
				if pv.Spec.CSI.ControllerExpandSecretRef == nil {
				                   pv.Spec.CSI.ControllerExpandSecretRef = &v1.SecretReference{}
			         } 
			         if val == templatizedName {
				               pv.Spec.CSI.ControllerExpandSecretRef.Name = vInfo.PersistentVolumeClaim
			         } else {
				               pv.Spec.CSI.ControllerExpandSecretRef.Name = val
			         }
			         // In the case of portworx volume backup, we will have namespace mapping always.
			        // So no need to check for the template string as we are going to change the value to destination namespace always.
			// If user does not change the namespace, the source and destination namespace in the mapping will be same.
			if _, ok := sc.Parameters[controllerExpandSecretNamespace]; ok {
				pv.Spec.CSI.ControllerExpandSecretRef.Namespace = dstNamespace
			}
	      }
```

adityadani: nit: also fix the typos in comment

drivers/volume/portworx/portworx.go
```diff
@@ -2608,7 +2615,59 @@ func (p *portworx) CancelMigration(migration *storkapi.Migration) error {
...
+
+			// Update the node publish secret
+			if pv.Spec.CSI.NodePublishSecretRef == nil {
```
adityadani, line 2654: same here: only set the NodePublishSecretRef object if you find those in the StorageClass parameters.

adityadani: This will ensure we don't set the "Namespace"  field in *SecretRef object even if there are not security params specified in the storage class.

adityadani (approved):
looks good after comment is fixed

drivers/volume/portworx/portworx.go
siva-portworx: Fixed  typos.

drivers/volume/portworx/portworx.go
siva-portworx: Fixed as suggested.

PR #1200, opened 2022-11-03
pb3206: Not returning error if the storageclass is missing on destination cluster.

siva-portworx:
**What type of PR is this?**
>bug

```
pb3206: Not returning error if the storageclass is missing on destination cluster
```

no

No, will be covered with pb-3174

yes. 2.12

Testing
Started the failing jenkins job with the  fixed image:
https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%202.12-dev%20%20C7/job/2.12-backup-k8s-1-23-0/376/console

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2621,67 +2621,69 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
 	if len(pv.Spec.StorageClassName) != 0 {
 		sc, err := storage.Instance().GetStorageClass(pv.Spec.StorageClassName)
 		if err != nil {
-			return nil, fmt.Errorf("failed in getting the storage class [%v]: %v", pv.Spec.StorageClassName, err)
+			logrus.Warnf("failed in getting the storage class [%v]: %v", pv.Spec.StorageClassName, err)
 		}
-		if isCsiProvisioner(sc.Provisioner) {
-			// add csi section in the pv spec
-			if pv.Spec.CSI == nil {
-				pv.Spec.CSI = &v1.CSIPersistentVolumeSource{}
-			}
-			// get the destinationNamespace
-			var dstNamespace string
-			var exists bool
-			if dstNamespace, exists = namespaceMapping[vInfo.SourceNamespace]; !exists {
-				dstNamespace = vInfo.SourceNamespace
-			}
-			// Update the controller expand secret
-			if val, ok := sc.Parameters[controllerExpandSecretName]; ok {
-				if pv.Spec.CSI.ControllerExpandSecretRef == nil {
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -2621,67 +2621,69 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
 	if len(pv.Spec.StorageClassName) != 0 {
 		sc, err := storage.Instance().GetStorageClass(pv.Spec.StorageClassName)
 		if err != nil {
-			return nil, fmt.Errorf("failed in getting the storage class [%v]: %v", pv.Spec.StorageClassName, err)
+			logrus.Warnf("failed in getting the storage class [%v]: %v", pv.Spec.StorageClassName, err)
```
prashanthpx, line 2624: Don't we need to return error from here ?

drivers/volume/portworx/portworx.go
```diff
@@ -2621,67 +2621,69 @@ func (p *portworx) UpdateMigratedPersistentVolumeSpec(
...
 				}
-				if val == templatizedNamespace {
-					pv.Spec.CSI.ControllerExpandSecretRef.Namespace = dstNamespace
-				} else {
-					pv.Spec.CSI.ControllerExpandSecretRef.Namespace = val
+				if val, ok := sc.Parameters[controllerExpandSecretNamespace]; ok {
+					if pv.Spec.CSI.ControllerExpandSecretRef == nil {
+						pv.Spec.CSI.ControllerExpandSecretRef = &v1.SecretReference{}
+					}
+					if val == templatizedNamespace {
```
prashanthpx, line 2653: Here asumption is `templatizedNamespace="${pvc.name}"` but if a user chooses something else as this is user choice, one suggestion is to have a Regex here to check for anything starting with `$` and use the val as is. 
When i was searching for this on csi website, i found an example https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#per-volume-secrets where it is different.

prashanthpx (approved):

drivers/volume/portworx/portworx.go
siva-portworx: No, if the storageclass is not present, we will return error. In that case, we will not add the CSI section. That is the fix

drivers/volume/portworx/portworx.go
siva-portworx: Good point. As per our document, we are suggesting this. 
https://docs.portworx.com/operations/operate-kubernetes/storage-operations/csi/security/
To support $(pvc.Name)-xxxxxxx, we need to change in other places as well. We can take it as separate fix.

adityadani (approved):

PR #1213, opened 2022-11-10
PWX-27549: Monitor csi-ext pods and ensure they run on online portwor…

pp511:
…x nodes

**What type of PR is this?**
>feature

If Stork detects that px-csi-ext pods are running on an offline PX node, it can delete the pod.
When Stork gets a scheduling request for such a pod, it can place the pod on the node where PX is operational. 

no

```release-note

Issue: 
When CSI is enabled, PVC provisioning goes through a deployment of 3 replicas of px-csi-ext-* pods. If PX on 
any of these nodes goes down, none of the PVC creations would succeed even if the PX cluster is operational. 

User Impact:
If PX goes down on a node where CSI Pod is deployed,  PVC requests on that Pod does not succeed.

Resolution:
Set Stork as the default scheduler for px-csi-ext-* pods. If Stork detects that these pods are running on an offline Portworx node, delete the CSI Pod. When Stork gets a scheduling request for such a Pod, it can place the Pods on the Node where PX is operational. 

```
yes 2.12.2

### DIFF
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -30,7 +30,6 @@ const (
 	nodeWaitFactor       = 2
 	nodeWaitSteps        = 5
 
-	CSIPodNamePrefix           = "px-csi-ext"
 	storageDriverOfflineReason = "StorageDriverOffline"
 )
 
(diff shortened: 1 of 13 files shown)


pkg/monitor/monitor.go
```diff
@@ -29,14 +30,15 @@ const (
 	nodeWaitFactor       = 2
 	nodeWaitSteps        = 5
 
+	CSIPodNamePrefix           = "px-csi-ext"
```
adityadani, line 33: We have tried to not mention PX specific naming in the common stork code. I would suggest adding a new API in the Driver interface - `GetCSIPodPrefix` and return `px-csi-ext` from Portworx driver implementation. For other drivers you can return an empty value and handle that in monitor and extender code

adityadani (approved):

PR #1223, opened 2022-11-19
pb-3312: Added steps to delete the resourceexport CR, when restore fails with timeout value.

siva-portworx:
```
pb-3312: Added steps to delete the resourceexport CR, when restore fails
    with timeout value.
```

no

no

no
Testing:
Tested restore after removing the resource.json file and checked that RE, Job and job pods are getting deleted.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -1477,7 +1477,7 @@ func (a *ApplicationRestoreController) restoreResources(
 	} else {
 		// Check whether ResourceExport is present or not
 		crName := getResourceExportCRName(utils.PrefixRestore, string(restore.UID), restore.Namespace)
-		resourceExport, err := kdmpShedOps.Instance().GetResourceExport(crName, a.restoreAdminNamespace)
+		resourceExport, err := kdmpShedOps.Instance().GetResourceExport(crName, restore.Namespace)
 		if err != nil {
 			if k8s_errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -1695,6 +1695,22 @@ func (a *ApplicationRestoreController) cleanupRestore(restore *storkapi.Applicat
 			return fmt.Errorf("cancel restore: %s", err)
 		}
 	}
+	// Directly calling DeleteResourceExport with out checking backuplocation type.
+	// For other backuplocation type, expecting Notfound
+	crName := getResourceExportCRName(utils.PrefixRestore, string(restore.UID), restore.Namespace)
```
diptiranjanpx, line 1700: Is the intention to delete from both AdminNamespace and restore.Namespace?
Currently both the deletes are deleting from restore.Namespace

siva-portworx: Fixed resource's RE also to use restore CR namespace instead of Adminnamespace.

diptiranjanpx, line 1700: We can just loop the same delete steps for both the CRs.

diptiranjanpx (approved):

pkg/applicationmanager/controllers/applicationrestore.go
siva-portworx: Fixed

PR #1226, opened 2022-11-22
PWX-27385: Honor IncludeResources flag when entering Application stage.

adityadani:
**What type of PR is this?**
>bug

- The migration controller was not honoring the IncludeResources flag and was
  trying to migrate all the resources even if IncludeResources was set to false.

- Add comments to the MigrationSpec.

No

Yes
```release-note
Will handle release note for this bug PWX-27385 in a follow up PR which involves more changes to reduce the overall time for migration. 

```

Yes - 2.12

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -487,7 +487,12 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 			}
 		}
 	case stork_api.MigrationStageApplications:
-		err := m.migrateResources(migration, false)
+		// Check if we need to handle only volume related resources such as PV, PVCs or all resources
+		var volumesOnly bool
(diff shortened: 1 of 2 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -2131,30 +2137,28 @@ func (m *MigrationController) getMigrationSummary(migration *stork_api.Migration
 		migrationSummary.ElapsedTimeForVolumeMigration = elapsedTimeVolume
 	}
 
-	if migration.Spec.IncludeResources == nil || *migration.Spec.IncludeResources {
```
adityadani, line 2134: We should calculate the MigrationSummary and populate the time taken for migration k8s resources,  even if IncludeResources is set to false, since we would be migrating the PV and PVC objects.

diptiranjanpx (approved):

pp511 (approved):
lgtm

PR #1234, opened 2022-12-07
pb-3358: In case of NFS BL prevent local snapshot

lalat-das:
While selecting a volumesnapshot class in case of CSI, stork should avoid taking local snapshot. this is specifically for NFS backuplocation.

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: We need to avoid taking local snapshot for NFS backuplocation.

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: No

Unit test:
Used pure-block csi provisioner and 5.0.1 volumesnapshotclass driver from k8s and cluster is 1.20 and above.
Without fix it fails in backup path:
Pasted Graphic

Now after the fix both ways i.e offload option and without offload option it passes whenever we specifically mention a volumesnapshot class to be used.
Pasted Graphic 3
with offload 
Pasted Graphic 2

Pasted Graphic 4


drivers/volume/kdmp/kdmp.go
```diff
@@ -321,7 +322,16 @@ func (k *kdmp) StartBackup(backup *storkapi.ApplicationBackup,
 		}
 		snapshotClassRequired := isCSISnapshotClassRequired(&pvc)
 		if snapshotClassRequired {
-			dataExport.Spec.SnapshotStorageClass = k.getSnapshotClassName(backup)
+			backupLocation, err := storkops.Instance().GetBackupLocation(backup.Spec.BackupLocation, backup.Namespace)
```
diptiranjanpx, line 325: Add a comment that it is temporary.

diptiranjanpx (approved):

PR #1236, opened 2022-12-08
Pod spread

Rohit-PX:
**What type of PR is this?**
integration-test

https://portworx.atlassian.net/browse/PWX-28054

no

no

2.12


Rohit-PX:
@pp511 I tested this out on `portworx/px-operator:1.10.2-dev` and it passed.

pp511:
@Rohit-PX To reproduce this issue, we have to use settings such that 3 nodes get equally score by Stork. You have used repl1 volume. One node will be scored higher by Stork and a Pod will get scheduled on that node. Remaining two pods might get spread equally. Did you try running with a previous operator version and the same test fails? 
I reproduced the issue with repl3 volume in which 3 nodes are scored equally and Pods end up chosing one or two of the 3 equally scored nodes.

Rohit-PX:
@pp511 I was able to reproduce the issue in `portworx/px-operator:2.10.1` and verified it in `portworx/operator:2.10.2-dev` using this test.

pp511 (approved):
LGTM

PR #1239, opened 2022-12-12
PWX-28159: Fix health monitor and extender integration test failures

pp511:

**What type of PR is this?**
>integration-test

Fix integration test failures
1. extender_test : defer has function scope and uncordon was not being executed for all the nodes due to which subsequent tests were failing
2. health_monitor_test:  We have a 4 worker node config in our integration test jobs. 3 nodes to run CSI pods and one node for failover. So the logic to stop volume driver on other nodes is not needed to test failover so commented that code.
Absence of KubeSchedulerConfiguration based stork scheduler, make multiple pods to run on single node sometime and we end up stopping px on more nodes than we intended and can not find the node to failover. This can be re enabled once the fix is integrated in operator so keeping the code commented here.

no

no

yes 2.12.2


Rohit-PX (approved):
Looks good except for the one suggestion.

PR #1240, opened 2022-12-13
Increase wait time for inactivity for travis

Rohit-PX:

**What type of PR is this?**
integration-test

Increase travis wait time since we are now running some commands inside docker containers.

no

no

2.12


pp511 (approved):
This change looks ok to me. We are past the problem where travis was stuck duting vet. However CBT tests in your PR build still seem to be stuck. May need to check if it doesn't succeed.

Rohit-PX:
> This change looks ok to me. We are past the problem where travis was stuck duting vet. However CBT tests in your PR build still seem to be stuck. May need to check if it doesn't succeed.

Actually, CBT tests seem to be passing as well, after this change.

PR #1245, opened 2022-12-16
PWX-28300: Add opsmanagers as the Plural form of MongoDBOpsManager CR plural set.

adityadani:
**What type of PR is this?**
>improvement

    - MongoDB defines "opsmanagers" as the plural form for MongoDBOpsManager in its CRD definition.
    - However the default plural set provided by k8s simply converts the Kind name "MongoDBOpsManager" to its plural form as "mongodbopsmanagers"
    - This causes migrations to fail, since this object is not identified by k8s.
    - Explicitly set the plural form of MongoDBOpsManager as "opsmanagers"

No

```release-note
Issue: Stork fails to apply MongoDBOpsManager objects on the target cluster since it fails with the error: "Error applying resource: the server could not find the requested resource". This happens since the plural form of MongoDBOpsManager CRD is defined as "opsmanagers" instead of "mongodbopsmanager"
User Impact: Migrations of MongoDB app deployed using the MongoDBOpsManager CR fails.
Resolution: Explicitly set the plural form of MongoDBOpsManager kind as "opsmanagers" in the plural set

```

Yes - 2.12

**Testing Notes**

Migrations failed with the below error without this change:
```
    - group: mongodb.com
      kind: MongoDBOpsManager
      name: myopsmanager
      namespace: default
      reason: Error applying resource: the server could not find the requested resource
      status: Failed
      transformedBy: ""
      version: v1

```
With this change migration of MongoDBOpsManager CR passes
```
    - group: mongodb.com
      kind: MongoDBOpsManager
      name: myopsmanager
      namespace: default
      reason: Resource migrated successfully
      status: Successful
      transformedBy: ""
      version: v1
```

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -1051,6 +1051,8 @@ func (a *ApplicationBackupController) uploadCRDResources(backup *stork_api.Appli
 	ruleset.AddPlural("quota", "quotas")
 	ruleset.AddPlural("prometheus", "prometheuses")
 	ruleset.AddPlural("mongodbcommunity", "mongodbcommunity")
+	ruleset.AddPlural("mongodbopsmanager", "opsmanagers")
+	ruleset.AddPlural("mongodb", "mongodb")
 	v1CrdApiReqrd, err := version.RequiresV1Registration()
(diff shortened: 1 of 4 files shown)


pp511 (approved):
looks good.

PR #1250, opened 2023-01-11
PWX-28055: Update stork integration tests to use kube scheduler confi…

pp511:
…g for K8s version 1.23+

**What type of PR is this?**
>integration-test

Update Stork integration tests to use kube scheduler configuration when they are using daemonset px installs.

no

no

yes

**Notes**:
-Stork test container is running a dash shell and  I couldn't find a straightforward way to do k8 version comparison. So I am breaking the string into the major and minor version and comparing.  Another option was to use  > with [[]] which are "bash" compatible syntaxes but we would need to update all scripts to use /bin/bash in that case and isn't portable. So I went ahead with the current approach.
- Reference Operator PR: https://github.com/libopenstorage/operator/pull/862

### DIFF
--- a/specs/stork-deployment-deprecated.yaml
+++ b/specs/stork-deployment-deprecated.yaml
@@ -1,0 +1,135 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: stork-config
+  namespace: kube-system
+data:
+  # The default list of predicates and priorities can change depending on your version
+  # of Kubernetes, so please update those as required. The requirement to manually specify this
+  # list will go away with Kubernetes v1.10 where it will use the defaults if nothing is
+  # specified.
+  policy.cfg: |-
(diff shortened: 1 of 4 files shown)


Rohit-PX (changes requested):
Looks good to me, but can we run a couple of integration jobs to make sure this works.

Rohit-PX (approved):

pp511:
Extender tests passed https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%202.12-dev%20%20C7/job/2.12-dev-extender-k8s-1-25-0/510/parameters/

Have not been able to execute other integration tests due to ongoing infra issues. Also CBTs are also not able to get triggereted due to the same issue https://portworx.atlassian.net/browse/PTX-14749 . Merging the PR.

PR #1258, opened 2023-01-28
PB-3397: Backed Nameapecs based on namespace label selector

sgajawada1:
- These feature is supported with manual and schedule backups
	- Currently Namespaces come along the applicationbackup spec consider to be backedup, Now these namespace are selected using the namespace selectors provided as part of the same spec at exection time.
	- Filtered namespaces that to be backed up are also updated to the applicationbackup resource status.

**What type of PR is this?**
>bug
feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

Namespaces that need to be backup can be filtered via label selector at execution time bot for manual and schedule backups

ApplicationBackups and ApplicationBackupSchedule CRD are updated to take namespace label selectors as input and save the filtered namespaces as part of the applicationbackup status CRD

```release-note
Issue:
User Impact:
Resolution

```


pkg/apis/stork/v1alpha1/applicationbackup.go
```diff
@@ -66,6 +67,7 @@ type ApplicationBackupStatus struct {
 	LastUpdateTimestamp metav1.Time                      `json:"lastUpdateTimestamp"`
 	FinishTimestamp     metav1.Time                      `json:"finishTimestamp"`
 	TotalSize           uint64                           `json:"totalSize"`
+	Namespaces          []string                         `json:"namespaces"`
```
siva-portworx, line 70: We already have the namespaces list in the applicationbackup spec definition itself. Cann't we us that list itself?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -245,6 +246,21 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		return nil
 	}
 
+	if labelSelector := backup.Spec.NamespaceSelector; len(labelSelector) != 0 {
+		namespaces, err := core.Instance().ListNamespaces(labelSelector)
+		if err != nil {
```
siva-portworx, line 251: Instead of returning error, it is better to record the error as event in the applicationbackup CR as it is done in other places.

siva-portworx (approved):

pkg/apis/stork/v1alpha1/applicationbackup.go
sgajawada1: During reconcile at px-backup we updated the datastore object with the CR status. So added the same in application backup status too.
Will verify the same once

pkg/applicationmanager/controllers/applicationbackup.go
sgajawada1: ok

lalat-das (approved):

lalat-das (approved):

siva-portworx (approved):

PR #1262, opened 2023-02-02
build binaries inside container

Rohit-PX:

**What type of PR is this?**
ntegration-test

https://portworx.atlassian.net/browse/PWX-28448

no

no

yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,9 @@ test:
 
 integration-test:
 	@echo "Building stork integration tests"
-	@cd test/integration_test && GOOS=linux go test -tags integrationtest $(BUILD_OPTIONS) -v -c -o stork.test
+	docker run --rm -v $(shell pwd):/go/src/github.com/libopenstorage/stork  $(DOCK_BUILD_CNT) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,17 +132,23 @@ codegen:
 
 stork:
 	@echo "Building the stork binary"
-	@cd cmd/stork && CGO_ENABLED=0 GOOS=linux go build $(BUILD_OPTIONS) -o $(BIN)/stork
+	docker run --rm -v $(shell pwd):/go/src/github.com/libopenstorage/stork  $(DOCK_BUILD_CNT) \
+           /bin/bash -c 'cd /go/src/github.com/libopenstorage/stork/cmd/stork && \
+		   CGO_ENABLED=0 GOOS=linux go build $(BUILD_OPTIONS) -o /go/src/github.com/libopenstorage/stork/bin/stork;'
```
diptiranjanpx, line 137: Will it not be a problem for "make container" where we do docker build using the dockerfile and there it needs to copy the stork binaries into the container from the host path itself ?

diptiranjanpx (approved):

PR #1265, opened 2023-02-06
pb-3338: Added support for handling default storage class request in the storageclass mapping

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3338: Added support for handling default storage class request in the
    storageclass mapping
```

no

```release-note
Issue: User are not able to select the default storageclass set on the destination cluster during restore.
User Impact: User should know the storageclass name on the destination cluster and was not able to select the storageclass that was set as default on the cluster.
Resolution: User can now select "use-default-storage-class" option int he storageclass mapping and internal, it will fetch the default storageclass set on the cluster for pvc provisioning. 

```

No.

Testing:
1) Tested by having two default storage class. The restore will fail and will update the CR with event.
```
Events:
  Type     Reason  Age                From   Message
  ----     ------  ----               ----   -------
  Warning  Failed  13s (x8 over 74s)  stork  more than one storageclass is set as default on destination cluster
```
2) Tested by not having default storage class. The restore will fail and will update the CR with event.
```
Events:
  Type     Reason  Age               From   Message
  ----     ------  ----              ----   -------
  Warning  Failed  4s (x6 over 44s)  stork  no storageclass is set as default on destination cluster
[root@ssubramani-pso-1 ~]#
```
3) Tested with one storageclass and restore completed with default storageclass.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -21,6 +21,7 @@ import (
 	"github.com/libopenstorage/stork/pkg/version"
 	"github.com/portworx/sched-ops/k8s/apiextensions"
 	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/portworx/sched-ops/k8s/storage"
 	storkops "github.com/portworx/sched-ops/k8s/stork"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -286,6 +302,49 @@ func (a *ApplicationRestoreController) handle(ctx context.Context, restore *stor
...
+		if len(scList.Items) > 1 {
+			errMsg := "more than one storageclass is set as default on destination cluster"
+			log.ApplicationRestoreLog(restore).Errorf(errMsg)
+			a.recorder.Event(restore,
+				v1.EventTypeWarning,
+				string(storkapi.ApplicationRestoreStatusFailed),
+				errMsg)
+			return nil
+		}
+		// If more than one storageclass is set as default storageclass, update error event
```
prashanthpx, line 328: Comment needs to change as here no storage class is found.

prashanthpx (approved):

pkg/applicationmanager/controllers/applicationrestore.go
siva-portworx: fixed

PR #1269, opened 2023-02-07
[PWX-27076, PWX-26970]: Fix confluent kafka async dr migration issues

pp511:
**What type of PR is this?**
>bug
>improvement

[DR: Apply CRD objects from the same group on the destination cluster even though there are no CRs  
](https://portworx.atlassian.net/browse/PWX-27076)

[Async-DR Migration not migrating all resources  of  "Confluent Platform Kafka application
](https://portworx.atlassian.net/browse/PWX-26970)

no

```release-note
Improvements:
Stork will migrate all the CRDs under the same group for which CR exists if it is for a different kind

```

```release-note
Improvements2: 
Stork will update owner reference for PVC objects on the destination cluster.
```

```release-note
Bug Fixes - 1 : 
Issue: Confluent Kafka operator not able to recognize service sub resources during Async-DR migration
User Impact: Not able to start application pods for Confluent Kafka
Resolution: Stork will not migrate Service resources if owner reference field is set
```

yes

**Notes for reviewer**:
* Update CRD if under group for which a CR exists  -> migration.go
* strimzi.kafka.io.Kafka was conflicting with   platform.confluent.io.Kafka due to which a separate applicationregistration 
 entry  for platform.confluent.io.Kafka is required. Creating an appreg entry with appReg.Name + "-" + newResource.Grou -> applicationregistration.go
* Do not migrate Service if owner reference is set -> resourcecollector.go
* Update PVC owner reference on destination cluster
   *   Update func "pruneOwnedResources" to return pvc with object owner reference -> resourcecollector.go
   *   Creating a copy of this list before calling "prepareResourcesForCollection".  prepareResourcesForCollection" updates metadata so a copy is required. Not returning the same from resourcecollector to keep to make sure we don't increase memory footprint unless needed.
   *  Added a new fun "updateOwnerReferenceOnPVC" to update PVC with owner references.
   * Note: Updating incorrect ownerReference can lead to deletion of the PVC.


pkg/applicationmanager/controllers/applicationregistration.go
```diff
@@ -71,6 +71,37 @@ func createAppReg(regCRD map[string][]stork_api.ApplicationResource) error {
...
+				logrus.Errorf("unable to get an existing application registration")
+			} else {
+				needsRegistration := false
+				appRegName := appReg.Name
+				for _, newResource := range appReg.Resources {
+					found := false
+					for _, existingResource := range existingAppReg.Resources {
+						if newResource.Group == existingResource.Group &&
+							newResource.Kind == existingResource.Kind {
+							found = true
```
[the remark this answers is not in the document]
pp511: done

pkg/applicationmanager/controllers/applicationregistration.go
```diff
@@ -71,6 +71,37 @@ func createAppReg(regCRD map[string][]stork_api.ApplicationResource) error {
...
+						}
+					}
+					if !found {
```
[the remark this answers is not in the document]
pp511: I have added a break condition inside !found

pkg/applicationmanager/controllers/applicationregistration.go
```diff
@@ -71,6 +71,37 @@ func createAppReg(regCRD map[string][]stork_api.ApplicationResource) error {
...
+						// Found a resource whose Group+Kind does not match with an existing app reg
+						// For ex. strimzi.kafka.io.Kafka vs platform.confluent.io.Kafka
+						needsRegistration = true
+						appRegName = appReg.Name + "-" + newResource.Group
```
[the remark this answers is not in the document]
pp511: Appending kind + group name. This should be unique.

pkg/applicationmanager/controllers/applicationregistration.go
```diff
@@ -71,6 +71,39 @@ func createAppReg(regCRD map[string][]stork_api.ApplicationResource) error {
...
+			if err != nil {
+				logrus.Errorf("unable to get an existing application registration")
+			} else {
+				needsRegistration := false
+				appRegName := appReg.Name
+				for _, newResource := range appReg.Resources {
+					found := false
+					for _, existingResource := range existingAppReg.Resources {
+						if newResource.Group == existingResource.Group &&
+							newResource.Kind == existingResource.Kind {
```
adityadani, line 86: Should we check for equality of versions as well?

pkg/migration/controllers/migration.go
```diff
@@ -1499,10 +1513,120 @@ func (m *MigrationController) getParsedLabels(
...
+		return err
+	}
+	remoteAdminInterface := remoteInterface
+	if migration.Spec.AdminClusterPair != "" {
+		remoteAdminInterface, err = dynamic.NewForConfig(remoteAdminConfig)
+		if err != nil {
+			return err
+		}
+	}
+
```
adityadani, line 1551: nit: Lets move this into a separate function and pass all the objects to this function and applyResources

pkg/migration/controllers/migration.go
```diff
@@ -1499,10 +1513,120 @@ func (m *MigrationController) getParsedLabels(
...
+	ruleset := m.getDefaultRuleSet()
+	for _, srcPvc := range pvcsWithOwnerRef {
+		destPvc, err := adminClient.CoreV1().PersistentVolumeClaims(srcPvc.GetNamespace()).Get(context.TODO(), srcPvc.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		// Skip updating if destination PVC has owner reference enabled
+		if len(destPvc.GetOwnerReferences()) > 0 {
```
adityadani, line 1559: Why do we need this check? I think we should keep making sure that the PVC on destination side always points to the right owner.

pkg/migration/controllers/migration.go
```diff
@@ -1499,10 +1513,120 @@ func (m *MigrationController) getParsedLabels(
...
+					} else {
+						dynamicClient = remoteAdminInterface.Resource(
+							o.GetObjectKind().GroupVersionKind().GroupVersion().WithResource(resource.Name))
+					}
+					break
+				}
+			}
+
+			if len(ownerName) == 0 {
+				return fmt.Errorf("could not find owner reference for pvc : %v", srcPvc.GetName())
```
adityadani, line 1597: Instead of returning an error lets continue. This is to handle a case where some user explicitly disabled migrating CRs by stork and only wants to migrate PV, PVC objects. In such cases we can skip updating the ownership for now.

adityadani (approved):
lgtm after comments are handled

pkg/migration/controllers/migration.go
```diff
@@ -1023,6 +1025,18 @@ func (m *MigrationController) migrateResources(migration *stork_api.Migration, v
 		return err
 	}
 
+	if !volumesOnly {
```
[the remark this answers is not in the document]
pp511: when volumeOnly is true, we will migrate PV and PVCs but no other objects so there is no owner reference to set. If Owner CRs for these PVCs might not be present on the destination. If we remove this check, we will get error when trying to get  destinationParentResourceUnstructured.

pkg/applicationmanager/controllers/applicationregistration.go
pp511: I think we should. Updating it to check the version too. Also updating the appReg name to include the version.

pkg/migration/controllers/migration.go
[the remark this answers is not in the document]
pp511: We are checking len(ownerName) == 0. If a valid ownerName is present in this condition, and we don't error out from if len(ownerName) == 0 we should be fine removing this check. I will remove it.

pkg/migration/controllers/migration.go
pp511: Done

pkg/resourcecollector/resourcecollector.go
```diff
@@ -613,30 +642,45 @@ func (r *ResourceCollector) objectToBeCollected(
...
+			if objectType.GetKind() == "PersistentVolumeClaim" {
+				for _, owner := range owners {
+					// Collect PVC objects separately. If OwnerReferences is available
+					// but the corresponding resource is not being collected, then
+					// no need to collect resource and we will not manually patch those PVCs
+					if _, exists := resourceMap[owner.UID]; exists {
+						var pvc v1.PersistentVolumeClaim
+						if err = runtime.DefaultUnstructuredConverter.FromUnstructured(o.UnstructuredContent(), &pvc); err != nil {
+							return nil, nil, err
+						}
```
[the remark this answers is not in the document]
pp511: fixed this

pkg/migration/controllers/migration.go
pp511: Agree. removing this check.

pkg/migration/controllers/migration.go
pp511: Moved this to a function and returning a struct of all the params. We can extract the respective param from the struct.

PR #1275, opened 2023-02-08
Revert "PWX-28818: Update libxml for vulnerability fixes"

pp511:
Reverting this change as it is breaking travis.


Rohit-PX (approved):
LGTM. Let's merge after travis and CBT pass.

PR #1276, opened 2023-02-09
Removed clonepvctest

gejain:
**What type of PR is this?**
>integration-test

We are having some issue with test, which causing migration test failure. Removing it as of now and add back later.

CC: @pp511


gejain:
Creating other PR.

PR #1248, opened 2023-01-05
Added test for Clone PVCs after DR on destination cluster

gejain:
**What type of PR is this?**
>integration-test for - https://portworx.atlassian.net/browse/PTX-9117
**What this PR does / why we need it**:
Resolves - https://portworx.atlassian.net/browse/PTX-9117

PR #1279, opened 2023-02-09
Revert "Added test for Clone PVCs after DR on destination cluster"

Rohit-PX:
Reverts libopenstorage/stork#1248

* I believe this is based on a discussion between @pp511  and @gejain 
* Doing this since @gejain does not have access to `reverts`


pp511 (approved):
@Rohit-PX Our migration jenkins jobs were red due to this test. Let's check what needs to be done for this test and have a successful run and we can merge this test in again. 
LGTM. Let's merge this PR travis build success.

PR #1283, opened 2023-02-11
Updating the version to 23.2

pp511:

**What type of PR is this?**
> branch update

Update the release version to 23.2

no

no

no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ifeq ($(BUILD_TYPE),debug)
 BUILDFLAGS += -gcflags "-N -l"
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,7 +31,7 @@ ifeq ($(BUILD_TYPE),debug)
 BUILDFLAGS += -gcflags "-N -l"
 endif
 
-RELEASE_VER := 2.7.0
```
adityadani, line 34: Lets change this to 23.2.0  and the branch name should be 23.2 - This was the standard we have been following in the past

PR #1290, opened 2023-02-16
PWX-28751: Fix svc and sa update issue on secondary cluster

pp511:
**What type of PR is this?**
>bug

PWX-28751

no

```release-note
Issue: Service and Service Account updates not getting reflected on destination cluster
Resolution: Allow service updates to be synced,  secrets associated with service account to be merged and service account param AutomountServiceAccountToken to be updated on the destination cluster

```

yes

**Notes**:
- We do not allow ServiceAccount and ClusterRoleBinding deletion during migration as we "mergeSupportedForResource" is true doe these resources.
- I am still waiting to hear from customer regarding what issues were they facing with ServiceAccount update https://portworx.atlassian.net/jira/servicedesk/projects/CEE/queues/custom/112/CEE-710  I noticed SA issue while reviewing the code.


pkg/migration/controllers/migration.go
```diff
@@ -1320,6 +1319,24 @@ func (m *MigrationController) checkAndUpdateDefaultSA(
 			destSA.ImagePullSecrets = append(destSA.ImagePullSecrets, s)
 		}
 	}
+
+	for _, s := range sourceSA.Secrets {
+		if !m.secretToBeMigrated(s.Name) {
```
diptiranjanpx, line 1324: Is this required as the secrets with the prefixes in AutoCreatedPrefixes are anyway not collected in resourcecollector?

diptiranjanpx (approved):

pkg/migration/controllers/migration.go
pp511: Secrets with AutoCreatedPrefixes are not collected or migrated. But if service has an associated secret that isn't auto created, it will be migrated and we should migrate it's reference in SA too.

PR #1294, opened 2023-02-18
PB-3413 - Added changes to support rancher project mapping

px-kesavan:
What type of PR is this? feature

**What this PR does / why we need it**: Add supports to perform project mapping for rancher platform
-> Extracts the associated rancher projects from k8s resources and updates the ApplicationBackup CR
-> Based on the project mapping available in the ApplicationRestore CR, it replaces the projects IDs in the required annotations, labels , namespace selectors with the target project ID

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -1031,14 +1031,9 @@ func updateRancherProjects(
 				projects[val] = ""
 			}
 		}
-		for key, val := range ns.Labels {
-			if strings.Contains(key, utils.CattleProjectPrefix) {
-				projects[val] = ""
-			}
-		}
 	}
 
-	if err := updateProjectDiplayNames(projects, platformCredential); err != nil {
+	if err := updateProjectDisplayNames(projects, platformCredential); err != nil {
 		return err
 	}
 
@@ -1047,8 +1042,8 @@ func updateRancherProjects(
 	return nil
 }
 
-// updateProjectDiplayNames updates the projectIDs with project display names
-func updateProjectDiplayNames(
+// updateProjectDisplayNames updates the projectIDs with project display names
+func updateProjectDisplayNames(
 	projects map[string]string,
 	platformCredential *stork_api.PlatformCredential,
 ) error {
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -23,6 +23,7 @@ import (
 	"github.com/libopenstorage/stork/pkg/k8sutils"
 	"github.com/libopenstorage/stork/pkg/log"
 	"github.com/libopenstorage/stork/pkg/objectstore"
+	"github.com/libopenstorage/stork/pkg/platform/rancher"
 	"github.com/libopenstorage/stork/pkg/resourcecollector"
 	"github.com/libopenstorage/stork/pkg/rule"
 	"github.com/libopenstorage/stork/pkg/utils"
@@ -33,11 +34,13 @@ import (
 	"github.com/sirupsen/logrus"
 	"gocloud.dev/gcerrors"
 	v1 "k8s.io/api/core/v1"
+	v1networking "k8s.io/api/networking/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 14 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -963,6 +966,272 @@ func (a *ApplicationBackupController) prepareResources(
...
+		ns, err := core.Instance().GetNamespace(namespace)
+		if err != nil {
+			return err
+		}
+		for key, val := range ns.Annotations {
+			if strings.Contains(key, utils.CattleProjectPrefix) {
+				projects[val] = ""
+			}
+		}
+		for key, val := range ns.Labels {
```
prashanthpx, line 1034: As i understand when a NS is created and added to project, both annotation and Labels should have same project name. Do we foresee any case where these values are different?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -963,6 +966,272 @@ func (a *ApplicationBackupController) prepareResources(
...
+			if strings.Contains(key, utils.CattleProjectPrefix) {
+				projects[val] = ""
+			}
+		}
+	}
+
+	if err := updateProjectDiplayNames(projects, platformCredential); err != nil {
```
prashanthpx, line 1041: Typo: updateProjectDiplayNames

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -963,6 +966,272 @@ func (a *ApplicationBackupController) prepareResources(
...
+	rancherClient.Init(*platformCredential.Spec.RancherConfig)
+	projectList, err := rancherClient.ListProjectNames()
+	if err != nil {
+		return err
+	}
+	for key, val := range projectList {
+		if _, ok := projects[key]; ok {
+			projects[key] = val
+		}
+		data := strings.Split(key, ":")
```
prashanthpx, line 1065: I assume this is for the case where clusterID:ProjectID is present for NS annotation right?
Rather than assigning at line#1063, can't we do the assignment either in if(len) or else case and avoid doing a delete.
When would data[1] be present in projectList?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -963,6 +966,272 @@ func (a *ApplicationBackupController) prepareResources(
...
+		}
+	}
+
+	for _, namespace := range backup.Spec.Namespaces {
+		ns, err := core.Instance().GetNamespace(namespace)
+		if err != nil {
+			return err
+		}
+		for key, val := range ns.Annotations {
+			if strings.Contains(key, utils.CattleProjectPrefix) {
```
prashanthpx, line 1030: As annotation will have both cluster:project,  where do we stip of clusterID from the annotation and extract only project?

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -116,9 +118,27 @@ func (a *ApplicationRestoreController) setDefaults(restore *storkapi.Application
...
+
 	return nil
 }
 
+func getRancherProjectMapping(restore *storkapi.ApplicationRestore) map[string]string {
+	rancherProjectMapping := map[string]string{}
+	if restore.Spec.RancherProjectMapping != nil {
+		for key, value := range restore.Spec.RancherProjectMapping {
+			rancherProjectMapping[key] = value
+			dataKey := strings.Split(key, ":")
```
prashanthpx, line 130: As i understand during the backup itself, we have taken relevant projects and stored in object. under which condition would this have value with `:` as delimeter?

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Fixed

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Yes both the annotation and label should be pointing to the same project.
If you create the namespace through management UI, the label and annotation for projectID is as given below:
```
labels:
    field.cattle.io/projectId: p-62xs2
```
```
annotation: ( In annotation, it contains the clusterID:projectID)
field.cattle.io/projectId: c-xg2ls:p-62xs2
```   
Some times, if it is created by applying the yaml directly, it is possible, it contains only one of them. So we are taking the value, if it is present in either in label or annotation.
@px-kesavan, But one thing, I noticed during the test is that if we have only label and not annotation, the management UI does not associate the namespace to a project, We need to have annotation to associate the ns to a project. Label is for just informational. So I feel, we should pick up the project only when it is present in the annotation.
For now, I will remove the label loop.

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: One more thing that I noticed on my setup is that, if we have the annotation with out clusterID:ProjectID, the ns is associated to the project. That means to associated a ns to a project, we must have the annotation with the in the format of "clusterID:ProjectID"
If this assumption is correct, we can very well avoid getting the  projectID from label. That means the project will always have x:y format.
@px-kesavan Will sync up with you and will make the code change accordingly.

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Currently the projectmap does not strip the clusterID.
projects[clustID:ProjectID] = projectDisplayableName
```
rancher_projects:
c-xg2ls:p-62xs2: "siva-project"

pkg/applicationmanager/controllers/applicationbackup.go
px-kesavan: @prashanthpx @siva-portworx In the namespace, both annotation and label will have the same projectID unless someone removes or modifies the annotation manually. 
If namespace is again move to proper project from UI , the rancher controller will automatically updates the annotation and label with correct values. 
if the namespace has valid projectID, the rancher controller will automatically update the label. 

Yes, we can remove the label check for namespace and get it from annotation .

pkg/applicationmanager/controllers/applicationbackup.go
px-kesavan: Extracted projects might have projectID with clusterID and without cluster ID . For instance let say namespace has `c-gw7q8:p-lmz7v` and namespace selector in network policy has `p-lmz7v`. Both belongs to same project and the project list has [`c-gw7q8:p-lmz7v` ,`p-lmz7v` ], instead of having both entries , the code is to combine both in a single entry (i.e) `c-gw7q8:p-lmz7v` and deleting `p-lmz7v`. So rancherproject list with always be in clusterid:projectid format. so it will be useful in project mapping , as we dont want to do for two entries , instead we can just do for one , as both belongs to same project.

pkg/applicationmanager/controllers/applicationbackup.go
px-kesavan: We strip of cluster id only during the restore, `getRancherProjectMapping`, will do the stripping , Lets say `c-gw7q8:p-lmz7v` : `c-greq1:p-zgs91` is the project mapping we got from restore request. `getRancherProjectMapping` will strip the clusterID and maintain two entries one as `c-gw7q8:p-lmz7v` : `c-greq1:p-zgs91` and the other with `p-lmz7v` : `p-zgs91`, so first will be used annotation replace and the later for namespace selector replace.

pkg/applicationmanager/controllers/applicationrestore.go
px-kesavan: No, during the backup we store in the format clusterID:projectID, please refer the previous comment.

pkg/applicationmanager/controllers/applicationbackup.go
px-kesavan: removed the label loop

pkg/applicationmanager/controllers/applicationbackup.go
px-kesavan: Fixed

pkg/applicationmanager/controllers/applicationbackup.go
prashanthpx: Was discussing with @siva-portworx what would be the use case where we need to store source cluster ID aka `c-gw7q8:p-lmz7v`. On destination cluster ID would be fetched by Rancher and the source cluster ID is not where used.

prashanthpx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: As discussed, it is need to do the direct matching the restore process.

PR #1296, opened 2023-02-21
pb-3557: Added check to return from applicationbackup handler, if stage is already final.

siva-portworx:
**What type of PR is this?**
> bug

```
pb-3557: Added check to return from applicationbackup handler, if stage is already final.
```
no

```release-note
Issue: Some the restore fails to restore the volumes as the applicationbackup CR get over written during restore.
User Impact: Some time the restoring of volumes faces failure.
Resolution: Added check not to process any thing on the applicationbackup CR, if it is already in final stage.

```

23.2.1

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -248,6 +248,10 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		return nil
 	}
 
+	// If the backup is already in final stage, return with out doing anything.
+	if backup.Status.Stage == stork_api.ApplicationBackupStageFinal {
+		return nil
+	}
 	if labelSelector := backup.Spec.NamespaceSelector; len(labelSelector) != 0 {
 		namespaces, err := core.Instance().ListNamespaces(labelSelector)
 		if err != nil {


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -220,6 +220,11 @@ func (a *ApplicationBackupController) createBackupLocationPath(backup *stork_api
 
 // handle updates for ApplicationBackup objects
 func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_api.ApplicationBackup) error {
+	// If the backup is already in final stage, return with out doing anything.
+	if backup.Status.Stage == stork_api.ApplicationBackupStageFinal {
```
diptiranjanpx, line 224: lets move it after the deletion timestamp check

diptiranjanpx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Fixed

siva-portworx:
Bypassing the branch protection as previous cycle the tests passed. I need to update the branch as one PR got merged.

PR #1309, opened 2023-02-24
PWX-24071: Add NamespacedAction CRD for Horntail failover

dgoel-px:
**What type of PR is this?**
feature

This PR adds a new CustomResourceDefinition called `NamespacedAction`. This is required to implement changes to promote near-sync volumes to sync volumes via the `NamespacedAction.Spec.Action` field that can take different `NamespacedActionType` like `failover` which will trigger the respective code path in the controller. 

No

No

No

**Notes for reviewer**
1. Add a new utility method `CreateCRDV1` to abstract away duplicate logic across controllers
2. All changes related to adding a new CRD, namely changes to api, controller and the ops (under porx).


pkg/apis/stork/v1alpha1/action.go
```diff
@@ -0,0 +1,60 @@
...
+type Action struct {
+	meta.TypeMeta   `json:",inline"`
+	meta.ObjectMeta `json:"metadata,omitempty"`
+	Spec            ActionSpec   `json:"spec"`
+	Status          ActionStatus `json:"status"`
+}
+
+// ActionSpec specifies the type of Action
+type ActionSpec struct {
+	ActionType ActionType `json:"actiontype"`
```
adityadani, line 37:
```suggestion
	ActionType ActionType `json:"actionType"`
```

pkg/action/action.go
```diff
@@ -0,0 +1,101 @@
...
+			logrus.Errorf("%s: %s: %s", reflect.TypeOf(ac), action.Name, err)
+			return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
+		}
+	}
+
+	return reconcile.Result{RequeueAfter: controllers.DefaultRequeue}, nil
+}
+
+func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) error {
+	ac.updateStatus(action, storkv1.ActionStatusInProgress)
```
adityadani, line 71: What is the point of updating the status to InProgress if we are anyways updating it to Successful immediately on line 75? 

I guess it is fine for now until we add the actual logic to check the status as well and then based on status perform an operation.

adityadani (approved):
lgtm - Can you create a PR that merges these changes into a feature branch instead of master?

PR #1310, opened 2023-02-24
PB-3579 - Added changes to create config map to hold info on custom admin namespace

px-kesavan:

**What this PR does / why we need it**: Added changes to create config map to hold info on custom admin namespace

**Testing SS:**
Screenshot 2023-02-24 at 7 14 44 PM
Screenshot 2023-02-27 at 9 16 49 AM

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -238,6 +238,21 @@ func run(c *cli.Context) {
 	} else if err != nil {
 		log.Warnf("Unable to create stork version configmap: %v", err)
 	}
+	// create configmap to store stork admin namespace details.
+	storkControllerCm := &api_v1.ConfigMap{}
+	storkControllerCm.Name = k8sutils.StorkControllerConfigMapName
+	storkControllerCm.Namespace = defaultAdminNamespace
+	storkControllerCm.Data = make(map[string]string)
+	storkControllerCm.Data[k8sutils.AdminNsKey] = c.String("admin-namespace")
+	_, err = schedops.Instance().CreateConfigMap(storkControllerCm)
+	if k8s_errors.IsAlreadyExists(err) {
+		_, err := schedops.Instance().UpdateConfigMap(storkControllerCm)
+		if err != nil {
+			log.Warnf("unable to update stork controller configmap: %v", err)
+		}
+	} else if err != nil {
+		log.Warnf("Unable to create stork controller configmap: %v", err)
+	}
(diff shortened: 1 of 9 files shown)


siva-portworx (approved):

siva-portworx:
@px-kesavan Let us not merge this PR to master. Need to confirm with @adityadani on which stork release, we will be landing up in.

cmd/stork/stork.go
```diff
@@ -238,6 +238,25 @@ func run(c *cli.Context) {
...
+	// create configmap to store stork admin namespace details when
+	// it differs from default admin namespace
+	adminNs := c.String("admin-namespace")
+	if adminNs != defaultAdminNamespace {
+		storkControllerCm := &api_v1.ConfigMap{}
+		storkControllerCm.Name = k8sutils.StorkControllerConfigMapName
+		storkControllerCm.Namespace = defaultAdminNamespace
+		storkControllerCm.Data = make(map[string]string)
+		storkControllerCm.Data[k8sutils.AdminNsKey] = c.String("admin-namespace")
+		_, err = schedops.Instance().CreateConfigMap(storkControllerCm)
```
diptiranjanpx, line 250: Why is the cm getting created only if adminNs != defaultAdminNamespace?

siva-portworx: Otherwise, it will be kube-system namespace by default. So we might be needing it.

siva-portworx:
@px-kesavan We need to modify the namespaceBackupAllowed function as well to allow both the kube-system namespace and admin namespace, So that existing schedule created in the kube-system will be continue to work. Please add a function test case for this as well.

cmd/stork/stork.go
px-kesavan: Yes, we wont be creating configmap if we don't have a custom admin namespace. As kube-system is the default one

px-kesavan:
> @px-kesavan We need to modify the namespaceBackupAllowed function as well to allow both the kube-system namespace and admin namespace, So that existing schedule created in the kube-system will be continue to work. Please add a function test case for this as well.

Fixed Siva

diptiranjanpx (approved):

cmd/stork/stork.go
px-kesavan: I was checking on the later modification of admin namespace, so when user removes the custom admin namespace. The config map was still holding older one. So modified the code , to create configmap even there is no custom admin namespace. 

By default, the configmap now will hold `kube-system` when admin namespace is not set

PR #1312, opened 2023-02-27
PWX-29253: Adding support for gke-gcloud-auth-plugin

pp511:
**What type of PR is this?**
>bug

Add support for gke-gcloud-auth-plugin https://portworx.atlassian.net/browse/PWX-29253

no

```release-note
Added support for gke-gcloud-auth-plugin required for authenticating with GKE.
```

yes 23.2

**Notes for the reviewer**:
- PWX-29253 reports error "configuration: interactiveMode must be specified". However this error is not present when using the storkctl binary from 23.2 build. This is due to the updated K8s APIs. We do not need to set an explicit Exec.InteractiveMode
- Changing the installation directory to /usr/lib for google-cloud-sdk as it should not be present in root directory
- platform/gsutil was being deleted from previous versions too. Also deleting RELEASE_NOTES(1 MB)
- python3 -m pip install to update third_party shouldn't be needed. 
- USE_GKE_GCLOUD_AUTH_PLUGIN should be set to true as per the doc [https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke ] I didn't observe any issues when it was not set on K8s v1.24.9-gke. But setting it nevertheless as the doc suggests it and we might run into issues on older K8s versions
- Creating symlink as the legacy auth plugin requires us to pass the path of gcloud https://github.com/libopenstorage/stork/blob/master/pkg/storkctl/clusterpair.go#:~:text=/google%2Dcloud%2Dsdk/bin/gcloud and we have set it to ./google-cloud-sdk/bin/gcloud. Symlink is required so that existing clusterpairs using this setting do not break with upgrade to the new Stork version
- Also updating integration-test container

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"flag"
-	"github.com/libopenstorage/stork/pkg/cache"
 	"net/http"
 	"os"
 	"os/signal"
@@ -23,6 +22,7 @@ import (
 	_ "github.com/libopenstorage/stork/drivers/volume/portworx"
 	"github.com/libopenstorage/stork/pkg/apis"
 	"github.com/libopenstorage/stork/pkg/applicationmanager"
+	"github.com/libopenstorage/stork/pkg/cache"
 	"github.com/libopenstorage/stork/pkg/clusterdomains"
 	"github.com/libopenstorage/stork/pkg/dbg"
 	"github.com/libopenstorage/stork/pkg/extender"
@@ -208,6 +208,10 @@ func main() {
 			Usage: "Start the resource transformation controller (default: true)",
 		},
 	}
+
+	// Export USE_GKE_GCLOUD_AUTH_PLUGIN=True
+	log.Warnf("Export USE_GKE_GCLOUD_AUTH_PLUGIN=True")
+	os.Setenv("USE_GKE_GCLOUD_AUTH_PLUGIN", "True")
 
 	if err := app.Run(os.Args); err != nil {
 		log.Fatalf("Error starting stork: %v", err)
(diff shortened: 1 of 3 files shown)


cmd/stork/stork.go
```diff
@@ -209,6 +209,10 @@ func main() {
 		},
 	}
 
+	// Export USE_GKE_GCLOUD_AUTH_PLUGIN=True
+	log.Warnf("Export USE_GKE_GCLOUD_AUTH_PLUGIN=True")
```
adityadani, line 213: What happens if we use this on the oldest supported k8s version GKE cluster? I am assuming it wont cause any problem since this is only being used to authenticate with Google cloud and generate a kubeconfig. We should test it out.

cmd/stork/stork.go
```diff
@@ -209,6 +209,10 @@ func main() {
...
+	os.Setenv("USE_GKE_GCLOUD_AUTH_PLUGIN", "True")
```
adityadani, line 214: Does it require "True" or "true"

adityadani (approved):

cmd/stork/stork.go
pp511: It is "True" https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke

cmd/stork/stork.go
pp511: I tested with  GCloud CLI version 399, Storkctl from 2.12, GK8s 1.24.9-gke. This created a kubeconfig which doesn't have any reference to gke-gcloud-auth-plugin so it ends up using deprecated gcp auth plugin and it worked.
I think this encironment variable should just be ignored by GCP on older K8s versions. But we can test it out in System Tests.

cmd/stork/stork.go
diptiranjanpx: is setting same env var also required in integration test container?

pp511: good point. I actually didn't face any issues when I didn't set this env variable. Will keep this in mind when we run integration tests on GKE. If we are facing issues in fetching the kubeconfig, we can set it then.

PR #1313, opened 2023-02-27
PWX-29253: Adding support for gke-gcloud-auth-plugin (#1312)

pp511:
**What type of PR is this?**
>bug

Add support for gke-gcloud-auth-plugin https://portworx.atlassian.net/browse/PWX-29253

no

```release-note
Added support for gke-gcloud-auth-plugin required for authenticating with GKE.
```

yes 23.2

**Notes for the reviewer**:
- PWX-29253 reports error "configuration: interactiveMode must be specified". However this error is not present when using the storkctl binary from 23.2 build. This is due to the updated K8s APIs. We do not need to set an explicit Exec.InteractiveMode
- Changing the installation directory to /usr/lib for google-cloud-sdk as it should not be present in root directory
- platform/gsutil was being deleted from previous versions too. Also deleting RELEASE_NOTES(1 MB)
- python3 -m pip install to update third_party shouldn't be needed. 
- USE_GKE_GCLOUD_AUTH_PLUGIN should be set to true as per the doc [https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke ] I didn't observe any issues when it was not set on K8s v1.24.9-gke. But setting it nevertheless as the doc suggests it and we might run into issues on older K8s versions
- Creating symlink as the legacy auth plugin requires us to pass the path of gcloud https://github.com/libopenstorage/stork/blob/master/pkg/storkctl/clusterpair.go#:~:text=/google%2Dcloud%2Dsdk/bin/gcloud and we have set it to ./google-cloud-sdk/bin/gcloud. Symlink is required so that existing clusterpairs using this setting do not break with upgrade to the new Stork version
- Also updating integration-test container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,16 +18,19 @@ RUN python3 -m pip install awscli  && python3 -m pip install rsa --upgrade
 RUN curl -q -o /usr/local/bin/aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.10.3/2018-07-26/bin/linux/amd64/aws-iam-authenticator && \
     chmod +x /usr/local/bin/aws-iam-authenticator
 
+#Install Google Cloud SDK
 ARG GCLOUD_SDK=google-cloud-sdk-418.0.0-linux-x86_64.tar.gz
-# Remove the test directories
-# Also don't need gsutil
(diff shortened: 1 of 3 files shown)


Rohit-PX (approved):
LGTM.

PR #1319, opened 2023-03-02
PWX-29487 - Added changes to fix nil dereferencing in PrepareResourceForApply

px-kesavan:

**What this PR does / why we need it**: Added changes to fix nil dereferencing in PrepareResourceForApply

**Does this change need to be cherry-picked to a release branch?**: 23.2


pkg/applicationmanager/controllers/applicationclone.go
```diff
@@ -545,7 +546,7 @@ func (a *ApplicationCloneController) prepareResources(
 			pvNameMappings,
 			clone.Spec.IncludeOptionalResourceTypes,
 			nil,
-			nil,
+			&opts,
```
diptiranjanpx, line 549: Can we fix it in the prepareResources function itself?
All the callers need not be modified because of this.

diptiranjanpx (approved):

PR #1320, opened 2023-03-03
PWX-29354: Update StorageClass on PV post PVC migration

pp511:
**What type of PR is this?**
>improvement

https://portworx.atlassian.net/browse/PWX-29354 
https://portworx.atlassian.net/browse/CEE-804
Some operators require StorageClass to be present on PV to be able to consume the resources.

no

yes
```
* Update StorageClass on PersistentVolumes during migration.
```

yes 23.3

**Jenkins Job for migration regressions**
https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.2-dev%20%20C7/job/23-2-migration-k8s-1-24-0/384/parameters/


pkg/migration/controllers/migration.go
```diff
@@ -1564,6 +1575,82 @@ func (m *MigrationController) getParsedLabels(
...
+				sc, err := storkcache.Instance().GetStorageClass(scName)
+				if err != nil {
+					return err
+				}
+				sc.UID = ""
+				sc.ResourceVersion = ""
+				log.MigrationLog(migration).Infof("Applying %v %v", sc.Kind, sc.Name)
+				for retries := 0; retries < maxApplyRetries; retries++ {
+					_, err = remoteClient.adminClient.StorageV1().StorageClasses().Create(context.TODO(), sc, metav1.CreateOptions{})
+					if err == nil {
```
diptiranjanpx, line 1624: Can we add err == nil || errors.IsAlreadyExists(err) ?

diptiranjanpx: Also what will happen if there exists another storageclass with same name but different parameters or even provisioners? this case is very unlikely though.

diptiranjanpx (approved):

pkg/migration/controllers/migration.go
pp511: 1. We are only creating the SC if it doens't exist. So I think we will never get "IsAlreadyExists". However, it still is logical to handle this condition. Updated as you mentioned.
2. We don't want to migrate SC if it already exists on destination. So if Sc exists with different params, we will not migrate the SC. We would still update the SC name to the PV.

pkg/migration/controllers/migration.go
```diff
@@ -1564,6 +1575,82 @@ func (m *MigrationController) getParsedLabels(
...
+				sc, err := storkcache.Instance().GetStorageClass(scName)
+				if err != nil {
+					return err
+				}
+				sc.UID = ""
+				sc.ResourceVersion = ""
+				log.MigrationLog(migration).Infof("Applying %v %v", sc.Kind, sc.Name)
+				for retries := 0; retries < maxApplyRetries; retries++ {
+					_, err = remoteClient.adminClient.StorageV1().StorageClasses().Create(context.TODO(), sc, metav1.CreateOptions{})
+					if err == nil || errors.IsAlreadyExists(err) {
```
diptiranjanpx, line 1624: We may hit the nil pointer derereference. 
Lets make it like 
if err == nil || ( err != nil && errors.IsAlreadyExists(err))

pp511: rhs of || will only be computed if err is not nil. So we don't need to check err != nil  condition explicitly.

diptiranjanpx: right. Ignore the comment.

PR #1321, opened 2023-03-06
PWX-29219: Use cache for getting pod list in health monitor

pp511:
**What type of PR is this?**
>improvement

https://portworx.atlassian.net/browse/PWX-29219 
Use Cache interface to get Pod list in health monitor

no

yes
```release-note - Improvement
Use Cache to pull Pod list in health monitor to optimize the total stork memory consumption.
```

yes. 23.3

**Notes**
- Updated the existing cache to using controllercache.Cache instead of manager client.
- Using controllercache.TransformByObject to modify Pod and only store information that is necessary. We can expand it to use other Params if needed later.
- healthmonitor [https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.2-dev%20%20C7/job/23-2-dev-healthmonitor-k8s125-0/366/] and migration jobs [https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.2-dev%20%20C7/job/23-2-migration-k8s-1-24-0/394/parameters/] to make sure we don't have any regressions. 
- Either need to come up with a fake cache or design UTs specifically for cache. For failing UTs, I had to add  !reflect.ValueOf(storkcache.Instance()).IsNil() so we don't use the cache insterfaces.


pkg/monitor/monitor.go
```diff
@@ -241,7 +242,8 @@ func (m *Monitor) cleanupDriverNodePods(node *volume.NodeInfo) {
 	if err == nil {
 		return
 	}
-	pods, err := core.Instance().GetPods("", nil)
+
+	pods, err := storkcache.Instance().ListPods()
```
diptiranjanpx, line 246: Is it ideal to list Pods from cache where pods are the objects which are most likely change?

pp511: Cache is always eventually consistent. However in this case we need to optimize the total space we are using to store Pod lists.

pkg/cache/cache.go
```diff
@@ -30,26 +32,67 @@ type SharedInformerCache interface {
...
+	transformMap := controllercache.TransformByObject{
+		&corev1.Pod{}: func(obj interface{}) (interface{}, error) {
+			podResource, ok := obj.(*corev1.Pod)
+			if !ok {
+				return nil, fmt.Errorf("unexpected object type: %T", obj)
+			}
+			currPod := corev1.Pod{}
+			currPod.Name = podResource.Name
+			currPod.Namespace = podResource.Namespace
+			currPod.Namespace = podResource.Namespace
```
pp511, line 70: noticed that Namespace is being assigned twice. Will update along with other review comments..if any.

pkg/cache/cache.go
```diff
@@ -30,26 +32,67 @@ type SharedInformerCache interface {
...
+
+			currPod.Spec.Volumes = podResource.Spec.Volumes
+			currPod.Spec.Containers = podResource.Spec.Containers
```
adityadani, line 73: Do we need Containers?

pkg/monitor/monitor.go
```diff
@@ -241,9 +243,17 @@ func (m *Monitor) cleanupDriverNodePods(node *volume.NodeInfo) {
 	if err == nil {
 		return
 	}
-	pods, err := core.Instance().GetPods("", nil)
+
+	var pods *v1.PodList
+	if !reflect.ValueOf(storkcache.Instance()).IsNil() {
```
adityadani, line 248: nit: if storkcache.Instance() != nil should work as well

adityadani (approved):
looks good

pkg/cache/cache.go
pp511: We do not need it in health-monitor which is the only place where we are using this Cache for listing Pods. However, we are using  Containers in webhook and might later want to update that interface to use the cache instead. So I will keep this for now.

pkg/monitor/monitor.go
pp511: done

pkg/cache/cache.go
```diff
@@ -30,26 +32,67 @@ type SharedInformerCache interface {
 
 	// ListApplicationRegistrations lists the application registration CRs from the cache
 	ListApplicationRegistrations() (*storkv1alpha1.ApplicationRegistrationList, error)
+
+	// ListPods lists the all the Pods from the cache
```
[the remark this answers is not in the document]
pp511: done

pkg/cache/cache.go
pp511: done

adityadani (approved):

PR #1322, opened 2023-03-07
pb-3633: fixed the v1/v1beta1 version call for volumesnapshot get in RestoreVolumeClaim()

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3633: fixed the v1/v1beta1 version call for volumesnapshot get in RestoreVolumeClaim()
```

no

NA.

23.2 branch

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 )
 
 require (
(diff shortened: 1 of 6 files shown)


lalat-das (approved):
lgtm

PR #1326, opened 2023-03-09
PWX-29480: Fixing plurals for WKC crds.

diptiranjanpx:
**What type of PR is this?**
improvement

Support migrations of watson knowledge catalog crs.

```release-note
Issue: Unable to find resource of Watson knowledge catalog.
User Impact: Migration was failing for Watson knowledge catalog.
Resolution: With right plurals for the crds, now migration is successful for Watson knowledge catalog.

```

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-openapi/inflect"
 	"github.com/libopenstorage/stork/drivers"
 	"github.com/libopenstorage/stork/drivers/volume"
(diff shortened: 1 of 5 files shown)


adityadani (approved):

diptiranjanpx:
> This works. Nit: I think we should move all these custom rulesets we are maintaining in one place and reference it from there. We could probably move getDefaultRuleSet from migration.go to resourcecollector.go and reference that function in all other packages.

Addressed.

pp511:
> > This works. Nit: I think we should move all these custom rulesets we are maintaining in one place and reference it from there. We could probably move getDefaultRuleSet from migration.go to resourcecollector.go and reference that function in all other packages.
> 
> Addressed.

Thanks for doing this :)

PR #1328, opened 2023-03-09
PWX-29534: Add integration test for Action CR

dgoel-px:
**What type of PR is this?**
integration-test

This PR adds the basic integration test for the new Action controller. It creates a migration, followed by a Action CR and checks to see if the application has come up on the destination as expected.

No

No

No

### DIFF
--- a/test/integration_test/action_test.go
+++ b/test/integration_test/action_test.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"github.com/sirupsen/logrus"
 	"testing"
-	"time"
 
 	"github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/portworx/sched-ops/k8s/core"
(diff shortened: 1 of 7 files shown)


test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func createSecret() *v1.Secret {
+	return &v1.Secret{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      "volume-secrets",
+			Namespace: "kube-system",
+		},
+		StringData: map[string]string{
+			"mysql-secret": "supersecretpassphrase",
```
Rohit-PX, line 41: can have the `key` and `value` as params to the `createSecret()` method, that way it is possible call this method for other apps or with other `passphrases` in other tests.

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+		logrus.Infof("creating secret in destination")
+		err := setDestinationKubeConfig()
+		require.NoError(t, err, "failed to set kubeconfig to destination cluster: %v", err)
+
+		defer func() {
+			err := setSourceKubeConfig()
+			require.NoError(t, err, "failed to set kubeconfig to source cluster: %v", err)
+		}()
+	} else {
+		logrus.Infof("creating secret in source")
```
Rohit-PX, line 57:
```suggestion
		logrus.Infof("creating secret in destination")
```

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	_, err := core.Instance().CreateSecret(secret)
+	if !errors.IsAlreadyExists(err) {
+		require.NoError(t, err, "failed to create secret for volumes")
```
Rohit-PX, line 62:
```suggestion
		require.NoError(t, err, "failed to create secret on %s for volumes", cluster)
```

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+			err := setSourceKubeConfig()
+			require.NoError(t, err, "failed to set kubeconfig to source cluster: %v", err)
+		}()
+	} else {
+		logrus.Infof("deleting namespace %v in source", namespace)
+	}
+
+	err := core.Instance().DeleteNamespace(namespace)
+	if err != nil {
+		logrus.Infof("Error deleting namespace: %v\n", err)
```
Rohit-PX, line 82:
```suggestion
		logrus.Infof("Error deleting namespace %s on %s: %v\n", namespace, cluster, err)
```

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func actionFailoverTest(t *testing.T) {
+
+	appKey := "mysql-enc-pvc"
+	instanceID := "mysql-action"
+	migrationAppId := "mysql-action-migration"
+	actionName := "mysql-action-failover"
+
+	namespace := fmt.Sprintf("%v-%v", appKey, instanceID)
+	deleteNamespace(t, namespace, "src")
```
Rohit-PX, line 95: any reason we are deleting the namespaces before the test?

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+	logrus.Infof("scaleFactor: %v", scaleFactor)
+
+	// TODO(dgoel): add a wrapper method that sets and unsets destKubeConfig
+	// trigger action
+	err = setDestinationKubeConfig()
+	require.NoError(t, err, "failed to set kubeconfig to destination cluster: %v", err)
+
+	createActionCR(t, actionName, namespace, ctxs[0])
+
+	err = schedulerDriver.WaitForRunning(preMigrationCtx, defaultWaitTimeout, defaultWaitInterval)
```
Rohit-PX, line 156: I think `preMigrationCtx` wont have the the `ActionCR` we should pass `ctxs[0]` here instead if we need to validate ActionCR as well here.

test/integration_test/action_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// if above call to WaitForRunning is successful,
+	// then Action validateActionCR should be successful too
+	validateActionCR(t, actionName, namespace)
+
+	err = setSourceKubeConfig()
+	require.NoError(t, err, "failed to set kubeconfig to source cluster: %v", err)
+}
+
+func createActionCR(t *testing.T, actionAppKey, namespace string, ctx *scheduler.Context) {
```
Rohit-PX, line 167: if we return the ActionCR object here, this method would be reusable in other places as well.

test/integration_test/action_test.go
Rohit-PX: can you also please address the above suggestion, if the test fails, it will help us know faster on which cluster secret creation failed.

test/integration_test/action_test.go
```diff
@@ -21,67 +20,63 @@ import (
...
-		logrus.Infof("deleting namespace %v in source", namespace)
+func cleanup(t *testing.T, namespace string) {
+	funcDeleteNamespace := func() {
+		err := core.Instance().DeleteNamespace(namespace)
+		if err != nil {
+			logrus.Infof("Error deleting namespace %s: %v\n", namespace, err)
+		}
 	}
+	funcDeleteNamespace()
+	executeOnDestination(t, funcDeleteNamespace)
```
Rohit-PX, line 65: this is a great idea, as an improvement we can also have something like `executeOnSource(...)`

Rohit-PX (approved):
Looks good except for the minor comments.

PR #1330, opened 2023-03-11
PWX-29463: Add promote method to Px

dgoel-px:
**What type of PR is this?**
feature

This PR adds an interface for `Action` to `volDriver` with a method `Failover`. It also adds implementation for `Failover` to Portworx volDriver that promotes nearsync volumes.

No

No

No

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -2440,40 +2440,47 @@ func (p *portworx) DeletePair(pair *storkapi.ClusterPair) error {
 
 func (p *portworx) Failover(action *storkapi.Action) error {
 	namespace := action.Namespace
-	logrus.Debugf("Failover: namespace %s", namespace)
+	logrus.Infof("failover: namespace %s", namespace)
 
-	// TODO(dgoel): when should we return error
+	// TODO(dgoel): when should we return error,
+	// for example if a volume promote fails
 	// and should we just log an error and continue
 
 	// TODO(dgoel): show relevant events in the Action CR
 	// return list of errors (as above) - controller can use to log events
 	// for ex: a volume promote failed
 
-	// can we directly fetch PersistentVolumes here?
-	pvList, err := core.Instance().GetPersistentVolumes()
+	pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, nil)
 	if err != nil {
-		return fmt.Errorf("error getting list of volumes to migrate: %v", err)
+		return fmt.Errorf("error fetching pvcList %v", err)
 	}
-	for _, pv := range pvList.Items {
-		if !p.OwnsPV(&pv) {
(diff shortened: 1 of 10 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -2438,6 +2438,75 @@ func (p *portworx) DeletePair(pair *storkapi.ClusterPair) error {
...
+	// can we directly fetch PersistentVolumes here?
+	pvcList, err := core.Instance().GetPersistentVolumeClaims(namespace, nil)
+	if err != nil {
+		return fmt.Errorf("error getting list of volumes to migrate: %v", err)
+	}
+	for _, pvc := range pvcList.Items {
+		if !p.OwnsPVC(core.Instance(), &pvc) {
+			continue
+		}
+		if resourcecollector.SkipResource(pvc.Annotations) { // TODO(dgoel): do we need this?
```
adityadani, line 2465: Yes we should keep this, so that if a user wants to not migrate a PVC object within a given namespace.

drivers/volume/portworx/portworx.go
```diff
@@ -2438,6 +2438,75 @@ func (p *portworx) DeletePair(pair *storkapi.ClusterPair) error {
...
+		if err != nil {
+			return err
+		}
+		volID := pv.Spec.PortworxVolume.VolumeID
+		if volID == "" {
+			return fmt.Errorf("PV contains \"\" value for Spec.PortworxVolume.VolumeID")
+		}
+		logrus.Debugf("Failover: volID %v", volID)
+
+		volumes, err := volDriver.Inspect([]string{volID})
```
adityadani, line 2485: You can query Portworx with pvName as well
```suggestion
		volumes, err := volDriver.Inspect([]string{pvName})
```

dgoel-px: Is it safe to assume that Px volume name will always be the same as PV name?

adityadani: Yes the PV name needs to be either the PX volume ID or volumeName. The Inspect() API accepts both PX volumeID and volumeName.

drivers/volume/portworx/portworx.go
```diff
@@ -2438,6 +2438,67 @@ func (p *portworx) DeletePair(pair *storkapi.ClusterPair) error {
...
+		if resourcecollector.SkipResource(pvc.Annotations) {
+			continue
+		}
+
+		pvName, err := core.Instance().GetVolumeForPersistentVolumeClaim(&pvc)
+		if err != nil {
+			return fmt.Errorf("error fetching volume for PVC: %v", err)
+		}
+
+		logrus.Infof("failover: volume %s", pvName)
```
adityadani, line 2470: Can you change this log line to be more descriptive?

adityadani (approved):

PR #1339, opened 2023-03-21
PWX-29423: Add ability to exclude resources during migration

pp511:
**What type of PR is this?**
>feature

https://portworx.atlassian.net/browse/PWX-29423 Allow user to specify "excludeResources:" to exclude resources from migration based on labels.
```
kind: MigrationSchedule
metadata:
  name: migration-schedule
  namespace: kube-system
annotations:
  # Add the below annotations when PX-Security is enabled on both the clusters
  #openstorage.io/auth-secret-namespace: <the namespace where the kubernetes secret holding the auth token resides>
  #openstorage.io/auth-secret-name: <the name of the kubernetes secret which holds the auth token>
spec:
  template:
    spec:
      clusterPair: cluster-pair
      includeResources: true
      startApplications: false
      namespaces:
      - <ns>
      selectors:
      excludeSelectors:
        app: cassandra
  schedulePolicyName: default-interval-policy
  suspend: false
  autoSuspend: false
```

yes

```
Allow user to specify resources to be excluded from migration in migrationschedule.

```

yes - 23.3


pp511:
Looking at other migrationschedule params and now that I think about it, excludeResources might not be the best naming param given we have a boolean includeResources.  Maybe we should update it to excludeResourcesSelector?

pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -23,6 +23,7 @@ type MigrationSpec struct {
 	SkipServiceUpdate            *bool             `json:"skipServiceUpdate"`
 	IncludeNetworkPolicyWithCIDR *bool             `json:"includeNetworkPolicyWithCIDR"`
 	Selectors                    map[string]string `json:"selectors"`
+	ExcludeResources             map[string]string `json:"excludeResources"`
```
adityadani, line 26: How about `ExcludeSelectors` ? The behavior is very similar to the existing `Selectors` flag, except it is reverse.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1501,7 +1502,7 @@ func (a *ApplicationBackupController) backupResources(
 				for _, resource := range resourceTypes {
 					if resource.Kind == backupResourceType || (backupResourceType == "PersistentVolumeClaim" && resource.Kind == "PersistentVolume") {
 						log.ApplicationBackupLog(backup).Tracef("GetResourcesType for : %v", resource.Kind)
-						objects, _, err := a.resourceCollector.GetResourcesForType(resource, nil, resourceTypeNsBatch, backup.Spec.Selectors, nil, true, resourceCollectorOpts)
+						objects, _, err := a.resourceCollector.GetResourcesForType(resource, nil, resourceTypeNsBatch, backup.Spec.Selectors, nil, nil, true, resourceCollectorOpts)
```
adityadani, line 1505: cc @siva-portworx @prashanthpx This is something backup can also leverage

pkg/resourcecollector/resourcecollector.go
```diff
@@ -576,6 +579,10 @@ func (r *ResourceCollector) objectToBeCollected(
 		return false, err
 	}
 
+	if exclude, err := r.excludeResource(object, excludeResourceSelectors, namespace); err != nil || exclude {
+		return false, err
```
adityadani, line 583: Can you wrap the error here, so that it is easier to understand what failed while  collecting the object?
```
return false, fmt.Errorf("failed to determine if object needs to be excluded: %v", err)
```

adityadani (approved):

pkg/resourcecollector/resourcecollector.go
pp511: There are two conditions being checked inside if. Updating it in the following way
```     
 exclude, err := r.excludeResource(object, excludeResourceSelectors, namespace)
       if err != nil {
              return false, fmt.Errorf("failed to determine if object needs to be excluded: %v", err)
       }
       if exclude {
               return false, nil
       }
```

PR #1341, opened 2023-03-23
PB-3728: GKE/EKS - Add zone information to restore object when replace policy is retain

px-kesavan:
What type of PR is this? bug

GKE - Add zone information to restore object when replace policy is retain - fetched zones from PV objects 
EKS - Add zone information to restore object when replace policy is retain - with volume id fetched from pv spec , zones are fetched using aws client

**Does this change need to be cherry-picked to a release branch?**: 23.2.1, 23.3

Testing on EKS

- The restore was successful, with policy as retain and was able to see the zone information in the restore object

Screenshot 2023-03-24 at 4 44 59 PM

Testing on GKE

- restored the app on the same zone cluster with retain replace policy, was able to see the zone information in the restore object (`us-central1-a`)

Screenshot 2023-03-24 at 11 42 36 PM

- restored the app on the different zone cluster with retain replace policy (with app already existing) , was was able to see the zone information in the restore object (`us-east1-b`)

Screenshot 2023-03-24 at 11 44 21 PM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -1188,6 +1188,29 @@ func (a *ApplicationRestoreController) skipVolumesFromRestoreList(
 			}
 			return newVolInfos, existingInfos, fmt.Errorf("erorr getting pvc %s/%s: %v", ns, pvcObject.Name, err)
 		}
+		pvName := pvc.Spec.VolumeName
+		var zones []string
+		// If PVC is present, fetch the corresponding PV spec and get the zone information
+		if driver.String() == volume.GCEDriverName || driver.String() == volume.AWSDriverName {
+			pv, err := core.Instance().GetPersistentVolume(pvName)
+			if err != nil && !errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -1124,6 +1124,7 @@ func (a *ApplicationRestoreController) skipVolumesFromRestoreList(
 		restoreVolInfo.Status = storkapi.ApplicationRestoreStatusRetained
 		restoreVolInfo.RestoreVolume = pvc.Spec.VolumeName
 		restoreVolInfo.TotalSize = bkupVolInfo.TotalSize
+		restoreVolInfo.Zones = bkupVolInfo.Zones
```
siva-portworx, line 1127: Can we test it with restoring to cluster on different zone? I feel, we are overwriting the zone value to source cluster zone.

siva-portworx (approved):

pkg/applicationmanager/controllers/applicationrestore.go
px-kesavan: As discussed , modified to fetch zones from PV objects. Implemented the changes for EKS as well

PR #1346, opened 2023-03-27
PWX-29754: Create or update the existing default-migration-policy's interval minutes to 30 instead of 1.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

As 1 minute migration policy was too aggressive, we are changing the default-migration-policy's interval to 30mins.

**Test**
Tested with fresh installation and upgrade case. 
Test results have been updated in PWX-29754


pp511 (approved):
lgtm

PR #1349, opened 2023-03-28
PWX-29594: More integration tests for failover action

dgoel-px:
**What type of PR is this?**
integration-test

This PR adds more integration tests for the Failover action
1. Failover without migration
2. Failover when volume promote fails
3. Failover of multiple namespaces
4. Failover of namespace with multiple applications

No

No

No


test/integration_test/common_test.go
```diff
@@ -928,6 +932,213 @@ func getTokenFromSecret(secretName, secretNamespace string) (string, error) {
...
+		StringData: secret_map,
+	}
+	secretObj, err := core.Instance().CreateSecret(secret)
+	if !errors.IsAlreadyExists(err) {
+		require.NoError(t, err, "failed to create secret for volumes")
+	}
+	return secretObj
+}
+
+func cleanup(t *testing.T, namespace string, storageClass string) {
```
Rohit-PX, line 950: * We can rewrite this function to take an additional param as `source` or `destination` so that it can also be used to cleanup on source.
* This will be useful during failback when source and destination roles are reversed.

test/integration_test/common_test.go
```diff
@@ -928,6 +932,213 @@ func getTokenFromSecret(secretName, secretNamespace string) (string, error) {
...
+
+	// wait for all apps to get to running state
+	for _, ctx := range ctxs {
+		err := schedulerDriver.WaitForRunning(ctx, defaultWaitTimeout, defaultWaitInterval)
+		require.NoError(t, err, "Error waiting for app to get to running state")
+	}
+	return ctxs
+}
+
+func scheduleTasksAndWait(t *testing.T, ctx *scheduler.Context, appKeys []string) {
```
Rohit-PX, line 1007:
```suggestion
func addTasksAndWait(t *testing.T, ctx *scheduler.Context, appKeys []string) {
```

test/integration_test/common_test.go
dgoel-px: On line 965, it already calls the funcDeleteNamespace for both source and destination.

Rohit-PX (approved):
LGTM.

PR #1352, opened 2023-03-29
cleanup changes

dgoel-px:
**What type of PR is this?**
cleanup

Yes, change `storkctl perform` to `storkctl do` as it more generally expresses the intent

No

No

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -76,8 +76,7 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
 			return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
 		}
 	}
-
-	ac.pruneActions(action.Namespace)
+	ac.pruneActionIfExpired(action)
 
 	return reconcile.Result{RequeueAfter: controllers.DefaultRequeue}, nil
 }
@@ -128,16 +127,12 @@ func (ac *ActionController) updateStatus(action *storkv1.Action, actionStatus st
 }
 
 // delete any action older than actionExpiryTime
-func (ac *ActionController) pruneActions(namespace string) {
-	actionList, err := storkops.Instance().ListActions(namespace)
-	if err != nil {
-		logrus.Errorf("call to fetch ActionList to prune older actions failed with error %v", err)
-	}
-	for _, action := range actionList.Items {
+func (ac *ActionController) pruneActionIfExpired(action *storkv1.Action) {
+	if action.Status == storkv1.ActionStatusFailed || action.Status == storkv1.ActionStatusSuccessful {
 		if time.Since(action.CreationTimestamp.Local()) >= actionExpiryTime {
-			err = storkops.Instance().DeleteAction(action.Name, namespace)
+			err := storkops.Instance().DeleteAction(action.Name, action.Namespace)
 			if err != nil {
-				log.ActionLog(&action).Errorf("received error when deleting expired action %v", err)
+				log.ActionLog(action).Errorf("received error when deleting expired action %v", err)
 			}
 		}
 	}
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -10,9 +10,12 @@ import (
 	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/controllers"
 	"github.com/libopenstorage/stork/pkg/k8sutils"
+	"github.com/libopenstorage/stork/pkg/log"
 	"github.com/libopenstorage/stork/pkg/resourceutils"
 	"github.com/portworx/sched-ops/k8s/apiextensions"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/rest"
@@ -25,6 +28,7 @@ import (
 const (
 	validateCRDInterval time.Duration = 5 * time.Second
 	validateCRDTimeout  time.Duration = 1 * time.Minute
+	actionExpiryTime    time.Duration = 24 * time.Hour
 )
 
 func NewActionController(mgr manager.Manager, d volume.Driver, r record.EventRecorder) *ActionController {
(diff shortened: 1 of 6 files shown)


pkg/action/action.go
```diff
@@ -93,23 +101,45 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action)
...
 	err := ac.client.Update(context.TODO(), action)
 	if err != nil {
-		logrus.Errorf("failed to update Action status %v/%v with error %v", action.Name, actionStatus, err)
+		log.ActionLog(action).Errorf("failed to update action status to %v with error %v", action.Status, err)
+	}
+}
+
+// delete any action older than actionExpiryTime
+func (ac *ActionController) pruneActions(namespace string) {
+	actionList, err := storkops.Instance().ListActions(namespace)
```
adityadani, line 132: Instead of calling ListActions you can prune the Action CR for which you originally got the reconcile event. 
Since you are requeueing every reconcile event, you should get an event for all Action CRs periodically. This will save us from making extra k8s API calls to ListActions on every reconcile period.

pkg/storkctl/action.go
```diff
@@ -7,21 +7,24 @@ import (
...
+	nameTimeSuffixFormat string        = "2006-01-02-150405"
+	actionWaitTimeout    time.Duration = 10 * time.Minute
+	actionWaitInterval   time.Duration = 10 * time.Second
 )
 
 func newPerformCommand(cmdFactory Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
 	performCommands := &cobra.Command{
 		Use:   "perform",
-		Short: "Perform actions",
+		Short: "perform actions",
```
adityadani, line 27: I thought you wanted to change this 'perform' -> 'do'

pkg/storkctl/action.go
```diff
@@ -40,10 +43,13 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
 				util.CheckErr(err)
 				return
 			}
+			var actions []storkv1.Action
 			for _, namespace := range namespaces {
 				if anyActionIncomplete(namespace) {
 					printMsg(
-						fmt.Sprintf("Cannot perform new action as an action is already scheduled/in-progress for %v", namespace),
+						fmt.Sprintf(
+							"failed to start failover for namespace %v as an action is already in scheduled/in-progress",
```
adityadani, line 51: nit: I would say lets print the status here, so that in future if we add more statuses we don't need to change the text.

pkg/storkctl/action.go
```diff
@@ -59,15 +65,54 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
...
+							"failed to start failover for namespace %v due to error %v",
+							namespace, err),
+						ioStreams.ErrOut)
+					continue
+				}
+				printMsg(fmt.Sprintf("started failover for namespace %v", namespace), ioStreams.Out)
+				actions = append(actions, action)
+			}
+			for _, action := range actions {
+				isSuccessful, err := waitForActionToComplete(action.Name, action.Namespace)
```
adityadani, line 79: If we are going to block the CLI from finishing until the action completes then we need to give some kind of status back to the user. Instead I would say print a line here indicating that the action has been triggered and check the status of the action. There is no guarantee ever that the action esp if it is a failback to finish within 10 mins

```
storkctl perform failover -n ns1
Started failover for namespace ns1. To check the status of the failover run the following command:
storkctl get failoverstatus -n ns1
```

pkg/storkctl/action.go
```diff
@@ -59,15 +65,54 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
...
+				if err != nil {
+					var errorMessage string
+					if _, ok := err.(*task.ErrTimedOut); ok {
+						errorMessage = fmt.Sprintf(
+							"timed out waiting for action %v/%v to complete\naction might still be scheduled/in-progress\n",
```
adityadani, line 84: nit: Since we are going to print this error on stdout, I feel we should start it with uppercase - `Timed out waiting.. `,  similar to how k8s does it
```
$ kubectl create -f pvc.yml
Error from server (AlreadyExists): error when creating "pvc.yml": persistentvolumeclaims "pvc" already exists
```

pkg/storkctl/action.go
```diff
@@ -59,15 +65,54 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
...
+							action.Namespace, action.Name) +
+							getDebugMessage(action.Name, action.Namespace)
+					} else {
+						errorMessage = fmt.Sprintf(
+							"received error when trying to get the action %v/%v: %v",
```
adityadani, line 89: nit: `Failed to get status of failover action for namespace %v: %v`

pkg/storkctl/action.go
```diff
@@ -86,3 +131,33 @@ func anyActionIncomplete(namespace string) bool {
...
+	if action.(*storkv1.Action).Status == storkv1.ActionStatusFailed {
+		return false, nil
+	}
+	return true, nil
+}
+
+func getDebugMessage(actionName, namespace string) string {
+	return "debug info:\n" +
+		"- for details on the action, use: " + getCmdDescribeAction(actionName, namespace) +
+		"\n- to view stork logs, use: kubectl -n kube-system logs -l name=stork"
```
adityadani, line 158: Stork could be running in any other namespace apart from `kube-system`. We can make this generic 
```
kubectl -n <stork-namespace> logs -lname=stork
```

pkg/action/action.go
dgoel-px: That's a great idea!

pkg/storkctl/action.go
dgoel-px: I do want to change. I had reverted it based on what you had said a while ago. I will revert my revert now so that it is `do`

pkg/storkctl/action.go
```diff
@@ -14,20 +14,23 @@ import (
 )
```
adityadani, line 14: Once the CLI is finalized, lets plan to add UTs to the action commands like we have for others.

pkg/storkctl/action.go
```diff
@@ -59,30 +64,45 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
...
+							namespace, err),
+						ioStreams.ErrOut)
+					continue
 				}
+				printMsg(fmt.Sprintf("Started failover for namespace %v", namespace), ioStreams.Out)
+				printMsg(getDescribeActionMessage(&action), ioStreams.Out)
 			}
 		},
 	}
 	return getClusterPairCommand
```
adityadani, line 79: nit: rename to failoverCommand

pkg/storkctl/action.go
```diff
@@ -41,9 +44,11 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
 				return
 			}
 			for _, namespace := range namespaces {
-				if anyActionIncomplete(namespace) {
+				if incompleteAction := getAnyIncompleteAction(namespace); incompleteAction != nil {
```
adityadani, line 47: On a second thought I think for all the provided namespace we should check if there is any pending action and error out instead of starting the action for a subset of namespaces. In this way the user gets a chance to fix their list of namespaces and trigger an action in one shot.

adityadani (approved):
lgtm after comments are addressed.

pkg/storkctl/action.go
dgoel-px: Created a ticket to track it https://portworx.atlassian.net/browse/PWX-30833
I will add the tests once the CLI is finalized

PR #1353, opened 2023-03-30
integration test for service and service account updates

Rohit-PX:
**What type of PR is this?**
integration-test

https://portworx.atlassian.net/browse/PWX-29221

no

no

no

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -85,6 +85,7 @@ func testMigration(t *testing.T) {
 	t.Run("operatorMigrationMongoTest", operatorMigrationMongoTest)
 	t.Run("operatorMigrationRabbitmqTest", operatorMigrationRabbitmqTest)
 	t.Run("bidirectionalClusterPairTest", bidirectionalClusterPairTest)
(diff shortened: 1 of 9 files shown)


test/integration_test/migration_test.go
```diff
@@ -85,6 +85,7 @@ func testMigration(t *testing.T) {
 	t.Run("operatorMigrationMongoTest", operatorMigrationMongoTest)
 	t.Run("operatorMigrationRabbitmqTest", operatorMigrationRabbitmqTest)
 	t.Run("bidirectionalClusterPairTest", bidirectionalClusterPairTest)
+	t.Run("serviceAccountUpdate", serviceAccountUpdate)
```
pp511, line 88: nit.. maybe we should name this serviceAndServiceAccount (or something similar. Someone lookin at the name wouldn't know we are doing service update tests in it too.

pp511 (approved):
one nit. Let's make sure cbts pass before merge.
looks good.

test/integration_test/migration_test.go
Rohit-PX: fixed.

PR #1355, opened 2023-04-03
PWX-30157: Recreate pv if its csi.spec.volumeHandle does not match with PV name during  migration.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
>bug

For px csi volumes, the pv created directly has volumeHandle as the portworx volumeID. During migration we create the pv with volumeHandle as portworx volume Name in the secondary. During failback, even though the volume gets correctly migrated, but still the pv spec points to old volumeID as volumeHandle. 
The fix is to find this case during migration and recreate the PV with volumehandle as PV name.

```release-note
Issue: Failback to primary when app uses px csi volumes fails as the volumeHandle points to old volume ID.
User Impact: App does not come up in primary after failback.
Resolution: With the fix of recreating pv and pointing the volumeHandle to volume name in its spec makes the app rightly use the pvc which gets properly bounded and app comes up fine.

```

**Test**:

1. Tested with migration with only csi pvcs.
2. Tested with migtation with both csi and non-csi pvcs.
Test results have been updated in PWX-30157

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -1881,6 +1881,51 @@ func (m *MigrationController) applyResources(
 			updatedObjects = append(updatedObjects, o)
 		}
 	}
+
+	// find out the csi PVs and if the volumeHandle does not match with pv Name , delete those.
+	// https://portworx.atlassian.net/browse/PWX-30157
+	pvToPVCMapping := getpvToPVCMappingFromPVCObjects(pvcObjects)
+	var csiPVCAndPVObjects []runtime.Unstructured
+	for _, obj := range pvObjects {
+		var pv v1.PersistentVolume
+		var err error
+		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), &pv); err != nil {
+			m.updateResourceStatus(
+				migration,
+				obj,
+				stork_api.MigrationStatusFailed,
+				fmt.Sprintf("Error unmarshalling pv resource: %v", err))
+			continue
+		}
+		if pv.Spec.CSI != nil {
+			respPV, err := remoteClient.adminClient.CoreV1().PersistentVolumes().Get(context.TODO(), pv.Name, metav1.GetOptions{})
+			if err != nil {
+				logrus.Errorf("error getting pv %s: %v", pv.Name, err)
+			}
+			if respPV.Spec.CSI != nil && respPV.Spec.CSI.VolumeHandle != pv.Name {
+				// Add the pvc object related to the PV for deleting
+				if _, ok := pvToPVCMapping[pv.Name]; ok {
+					csiPVCAndPVObjects = append(csiPVCAndPVObjects, pvToPVCMapping[pv.Name])
+				}
+				// Add the pv object to the list also for getting deleted
+				csiPVCAndPVObjects = append(csiPVCAndPVObjects, obj)
+			}
+		}
+	}
+	if len(csiPVCAndPVObjects) > 0 {
+		dynamicInterface, err := dynamic.NewForConfig(remoteClient.remoteAdminConfig)
+		if err != nil {
+			return err
+		}
+		err = m.resourceCollector.DeleteResources(
+			dynamicInterface,
+			csiPVCAndPVObjects)
+		if err != nil {
+			logrus.Errorf("error deleting csi pvcs and pvs: %v ", err)
+			return err
+		}
+	}
+
 	// create/update pv object with updated policy
 	for _, obj := range pvObjects {
 		var pv v1.PersistentVolume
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -2483,3 +2528,19 @@ func (m *MigrationController) getVolumeOnlyMigrationResources(
 	resources = append(resources, objects.Items...)
 	return resources, pvcWithOwnerRef, nil
 }
+
+func getpvToPVCMappingFromPVCObjects(pvcObjects []runtime.Unstructured) map[string]runtime.Unstructured {
```
pp511, line 2532: nit.getPVToPVC...

pkg/migration/controllers/migration.go
```diff
@@ -1881,6 +1881,51 @@ func (m *MigrationController) applyResources(
...
+	}
+	if len(csiPVCAndPVObjects) > 0 {
+		dynamicInterface, err := dynamic.NewForConfig(remoteClient.remoteAdminConfig)
+		if err != nil {
+			return err
+		}
+		err = m.resourceCollector.DeleteResources(
+			dynamicInterface,
+			csiPVCAndPVObjects)
+		if err != nil {
```
pp511, line 1923: We already have access remoteClient in this function. Instead of using NewForConfig we should try to use remoteClient.remoteAdminInterface

pkg/migration/controllers/migration.go
```diff
@@ -1881,6 +1881,51 @@ func (m *MigrationController) applyResources(
...
+				obj,
+				stork_api.MigrationStatusFailed,
+				fmt.Sprintf("Error unmarshalling pv resource: %v", err))
+			continue
+		}
+		if pv.Spec.CSI != nil {
+			respPV, err := remoteClient.adminClient.CoreV1().PersistentVolumes().Get(context.TODO(), pv.Name, metav1.GetOptions{})
+			if err != nil {
+				logrus.Errorf("error getting pv %s: %v", pv.Name, err)
+			}
```
pp511, line 1904: "continue" here

pkg/migration/controllers/migration.go
pp511:
```
err = m.resourceCollector.DeleteResources(
			remoteClient.remoteAdminInterface,
			csiPVCAndPVObjects
```

pkg/migration/controllers/migration.go
```diff
@@ -1881,6 +1881,51 @@ func (m *MigrationController) applyResources(
...
+			if respPV.Spec.CSI != nil && respPV.Spec.CSI.VolumeHandle != pv.Name {
+				// Add the pvc object related to the PV for deleting
+				if _, ok := pvToPVCMapping[pv.Name]; ok {
+					csiPVCAndPVObjects = append(csiPVCAndPVObjects, pvToPVCMapping[pv.Name])
+				}
+				// Add the pv object to the list also for getting deleted
+				csiPVCAndPVObjects = append(csiPVCAndPVObjects, obj)
```
pp511, line 1911: I think this should be ok but just trying to understand your thought process. If no PVC is found in the pvToPVCMapping(ideally this should not happen), do we still want to go ahead and deleted PV object or we want to skip?

pp511 (commented):
Mostly looks good. Need you to address some comments.

pkg/migration/controllers/migration.go
diptiranjanpx: I got your point. My understanding was to handle only PV case itself.

pkg/migration/controllers/migration.go
diptiranjanpx: good catch. will do.

pkg/migration/controllers/migration.go
diptiranjanpx: done

pkg/migration/controllers/migration.go
diptiranjanpx: done

pp511 (approved):
Looks good. Make sure to cherry-pick your changes to 23.3.1 branch in addition to 23.3.

pkg/migration/controllers/migration.go
```diff
@@ -1881,6 +1881,48 @@ func (m *MigrationController) applyResources(
...
+			if err != nil {
+				logrus.Errorf("error getting pv %s: %v", pv.Name, err)
+				continue
+			}
+			if respPV.Spec.CSI != nil && respPV.Spec.CSI.VolumeHandle != pv.Name {
+				// Add the pvc object related to the PV for deleting
+				if _, ok := pvToPVCMapping[pv.Name]; ok {
+					csiPVCAndPVObjects = append(csiPVCAndPVObjects, pvToPVCMapping[pv.Name])
+				}
+				// Add the pv object to the list also for getting deleted
```
adityadani, line 1911: Is the goal to delete the PV object if the volumeHandle is pointing to a volumeID or during every migration cycle?

adityadani (approved):

pkg/migration/controllers/migration.go
diptiranjanpx: Only if volumeHandle is pointing to volumeID not volumeName. With this in forward migration, secondary PVs are not going to be deleted once created in the first migration. And for reverse migration, it will be the first migration which will delete and recreate the PV with volumeHandle as volumeName. Subsequent reverse migrations will also not delete the PV. The goal was to not delete PV in each migration.

PR #1356, opened 2023-04-04
PWX-30112: Making default k8s-api-qps value as 1000 and k8s-api-burst as 2000.

diptiranjanpx:
Signed-Off-By: Diptiranjan

>improvement

Increasing default k8 api qps and burst in stork for better performance.


pp511 (approved):
looks good

PR #1357, opened 2023-04-04
PWX-30127: Add test for ExcludeSelectors param in MigrationSchedule

pp511:
**What type of PR is this?**
>integration-test

Adding integration test for excludeSelector param
https://portworx.atlassian.net/browse/PWX-30127

no

yes 23.4

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -68,6 +68,7 @@ func testMigration(t *testing.T) {
 	// TODO: waiting for https://portworx.atlassian.net/browse/STOR-281 to be resolved
 	if authTokenConfigMap == "" {
 		t.Run("labelSelectorTest", migrationLabelSelectorTest)
+		t.Run("labelExcludeSelectorTest", migrationLabelExcludeSelectorTest)
 		t.Run("intervalScheduleTest", migrationIntervalScheduleTest)
 		t.Run("dailyScheduleTest", migrationDailyScheduleTest)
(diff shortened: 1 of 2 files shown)


test/integration_test/migration_test.go
```diff
@@ -442,6 +443,19 @@ func migrationLabelSelectorTest(t *testing.T) {
 	)
 }
 
+func migrationLabelExcludeSelectorTest(t *testing.T) {
+	triggerMigrationTest(
+		t,
+		"migration-label-exclude-selector-test",
+		"cassandra",
+		[]string{"mysql-1-pvc"},
```
diptiranjanpx, line 451: How are we verifying if mysql app is not getting migrated?

pp511: I am just following the existing framework.

appKey = cassandra and additionalAppKeys = mysql, migrateAllAppsExpected = false, startAppsOnMigration = true

func() validateAndDestroyMigration 
: allAppsCtx -> Cassandra (Cassandra migrated)
if (migrationSuccessExpected) -> true 
 if !migrateAllAppsExpected -> true
   require.Error(t, err, "All apps shouldn't have been migrated")
This means if mysql got migrated and scheduled we would report and error and test would fail.

pp511: migrationLabelSelectorTest is implemented in a similar way.

diptiranjanpx (approved):

PR #1359, opened 2023-04-11
pb-3773: Update restore CR with restored resources & state

lalat-das:
- Added two field to restore CR one signifies number of already applied resource
- The other is the phase in which resource restoration process operating. It has three phases
	- preparing resource
	- deleting resource( only if replace is selected)
	- verifying the deleted resource
	- applied resource

**What type of PR is this?**
>bug
feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: We need to showcase of a progress bar for large resource restore. Hence we need some additional field to be updated in certain interval  in restore CR. Additionally a restore resource state is also added to signify in what state the resource process is operating.

**Does this PR change a user-facing CRD or CLI?**: No, it adds values to only status field

```release-note
Issue:
User Impact:
Resolution

```

Unit test:
Verified a multi namespace and a large resource based restore and  large resource based resource which has a volume as well as resources present in it.
image

image

### DIFF
--- a/pkg/apis/stork/v1alpha1/applicationrestore.go
+++ b/pkg/apis/stork/v1alpha1/applicationrestore.go
@@ -49,18 +49,29 @@ const (
 	ApplicationRestoreReplacePolicyRetain ApplicationRestoreReplacePolicyType = "Retain"
 )
 
+type ApplicationRestoreResourceStateType string
+
+const (
+	ApplicationRestoreResourcePreparing ApplicationRestoreResourceStateType = "Preparing"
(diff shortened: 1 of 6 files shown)


diptiranjanpx (approved):

lalat-das:
@siva-portworx  gave a comment to update the current restore CR's in memory copy instead of operating on new structure. this is to avoid unnecessary assignments just before the final CR update in RestoreResources() call.  Hence modified the same too

siva-portworx:
Force merging as QA is blocked for 2.4.2 release testing. Will keep track of the travis and CBT test.

PR #1361, opened 2023-04-13
PWX-29789: Implementation of crd plural map for dynamically getting crd's plurals

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
>improvement

For finding crd's plural for querying,  currently inflect ruleset was used.
Sometimes the crd's plural does not align with the library's rules and in that case every time a stork release is required with added exceptions. 
The current fix will find the existing crd's plurals and keep it in a global cache for further use. 

**Test**
Tested following 2 cases with migration for confluent kafka application.
1. App was installed first and stork gets installed later.
2. Stork was installed first and app gets installed later.

Details of test results have been updated in PWX-29789.

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -33,6 +33,7 @@ import (
 	"github.com/libopenstorage/stork/pkg/migration/controllers"
 	"github.com/libopenstorage/stork/pkg/monitor"
 	"github.com/libopenstorage/stork/pkg/objectcontroller"
+	"github.com/libopenstorage/stork/pkg/pluralmap"
 	"github.com/libopenstorage/stork/pkg/pvcwatcher"
 	"github.com/libopenstorage/stork/pkg/resourcecollector"
 	"github.com/libopenstorage/stork/pkg/rule"
@@ -475,6 +476,12 @@ func runStork(mgr manager.Manager, ctx context.Context, d volume.Driver, recorde
 	if adminNamespace == "" {
 		adminNamespace = c.String("migration-admin-namespace")
 	}
+
+	// Setting up the pluralmap. It has the right plural for a crd kind installed in the cluster.
+	if err := pluralmap.CreateCRDPlurals(); err != nil {
+		log.Fatalf("failed to setup crd plural map: %v", err)
+	}
+	log.Infof("crd plural map has been intialized")
 
 	monitor := &monitor.Monitor{
 		Driver:      d,
--- a/pkg/applicationmanager/controllers/applicationregistration.go
+++ b/pkg/applicationmanager/controllers/applicationregistration.go
@@ -8,6 +8,7 @@ import (
 
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/appregistration"
+	"github.com/libopenstorage/stork/pkg/pluralmap"
 	"github.com/portworx/sched-ops/k8s/apiextensions"
 	"github.com/portworx/sched-ops/k8s/stork"
 	"github.com/portworx/sched-ops/task"
@@ -170,12 +171,24 @@ func RegisterDefaultCRDs() error {
 			if err := registerCRDV1(*crd); err != nil {
 				logrus.WithError(err).Error("unable to create appreg for v1 crd")
 			}
+			if len(crd.Spec.Names.Kind) > 0 {
+				if _, ok := pluralmap.Instance().GetCRDKindToPluralMap()[strings.ToLower(crd.Spec.Names.Kind)]; !ok {
+					pluralmap.Instance().SetPluralForCRDKind(crd.Spec.Names.Kind, crd.Spec.Names.Plural)
+					logrus.Infof("Adding new crd to plural map %s/%s", crd.Spec.Names.Kind, crd.Spec.Names.Plural)
+				}
+			}
 		} else if crd, ok := object.(*apiextensionsv1beta1.CustomResourceDefinition); ok {
 			if _, ok := skipCrds[crd.Spec.Group]; ok {
 				return
 			}
 			if err := registerCRD(*crd); err != nil {
 				logrus.WithError(err).Error("unable to create appreg for v1beta1 crd")
+			}
+			if len(crd.Spec.Names.Kind) > 0 {
+				if _, ok := pluralmap.Instance().GetCRDKindToPluralMap()[strings.ToLower(crd.Spec.Names.Kind)]; !ok {
+					pluralmap.Instance().SetPluralForCRDKind(crd.Spec.Names.Kind, crd.Spec.Names.Plural)
+					logrus.Infof("Adding new crd to plural map %s/%s", crd.Spec.Names.Kind, crd.Spec.Names.Plural)
+				}
 			}
 
 		} else {
--- a/pkg/pluralmap/pluralmap.go
+++ b/pkg/pluralmap/pluralmap.go
@@ -1,0 +1,94 @@
+package pluralmap
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/portworx/sched-ops/k8s/apiextensions"
+	"github.com/sirupsen/logrus"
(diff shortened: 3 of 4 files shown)


cmd/stork/stork.go
```diff
@@ -476,6 +477,12 @@ func runStork(mgr manager.Manager, ctx context.Context, d volume.Driver, recorde
 		adminNamespace = c.String("migration-admin-namespace")
 	}
 
+	// Setting up the generic cache. Currently it has the map to determine the right plural for a crd kind installed in the cluster.
+	if err := genericcache.CreateGenericCache(); err != nil {
```
pp511, line 481: I feel instead of CreateGenericCache() It should be CreateCrdCache() as this cache is very specific to crds

cmd/stork/stork.go
```diff
@@ -476,6 +477,12 @@ func runStork(mgr manager.Manager, ctx context.Context, d volume.Driver, recorde
...
+		log.Fatalf("failed to setup generic cache: %v", err)
+	}
+	log.Infof("generic cache has been intialized")
```
pp511, line 484: nit.. crd/<any other name> has been intialized

cmd/stork/stork.go
pp511: Or if we name the package crdcache instead of genericcache, we can name the function CreateCache()

pkg/applicationmanager/controllers/applicationregistration.go
```diff
@@ -170,13 +171,25 @@ func RegisterDefaultCRDs() error {
 			if err := registerCRDV1(*crd); err != nil {
 				logrus.WithError(err).Error("unable to create appreg for v1 crd")
 			}
+			if len(crd.Spec.Names.Kind) > 0 {
+				if _, ok := genericcache.Instance().GetCRDkindToPluralmap()[strings.ToLower(crd.Spec.Names.Kind)]; !ok {
+					genericcache.Instance().SetCRDkindToPluralmap(crd.Spec.Names.Kind, crd.Spec.Names.Plural)
```
pp511, line 176: nit..CRDkindToPluralmap-> Kind..Map. Let's try to follow camel case for all functions and variables.
SetCRDkindToPluralmap, crdkindToPluralmap

pp511 (commented):
some nitpicking that I need you to address. overall looks good.

cmd/stork/stork.go
diptiranjanpx: My intention in naming this as genericCache is that if in future we will require some more things need to be cached, we can make it as part of genericCache structure and provide the necessary methods to the interface. I agree that it is solving only crdPlurals for now.

pp511: Ok. Let's keep package name genericCache. But we can name the func CreateCache()

cmd/stork/stork.go
diptiranjanpx: Kept it as generic cache since we agreed upon the package name.

cmd/stork/stork.go
diptiranjanpx: Done

pkg/genericcache/genericcache.go
```diff
@@ -0,0 +1,89 @@
...
+type GenericCache interface {
+	// GetCRDKindToPluralMap returns the crdKindToPluralMap from generic cache singleton.
+	GetCRDKindToPluralMap() map[string]string
+
+	// SetCRDKindToPluralMap will set a new CRD's plural in the map
+	SetCRDKindToPluralMap(kind, plural string)
+}
+
+var (
+	genericCacheLock sync.Mutex
```
adityadani, line 28: move these into the `genericCache` struct instead of global variable.

pkg/genericcache/genericcache.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	genericCacheMap = &genericCache{}
+	genericCacheMap.crdKindToPluralMap = getCRDKindToGroupMap()
+	logrus.Debugf("Current crd plural map: %v", genericCacheMap.crdKindToPluralMap)
+
+	return nil
+}
+
+func (g *genericCache) GetCRDKindToPluralMap() map[string]string {
+	return genericCacheMap.crdKindToPluralMap
```
adityadani, line 81: This will lead to `concurrent map read write` error.

I don't think we should return this map which is also governed by a lock to the caller. You might get a watch update and try to add an entry into this map while someone is trying to do a get w/o a lock. 

Make a copy of the map under a lock and return the copy.

pkg/genericcache/genericcache.go
```diff
@@ -0,0 +1,89 @@
...
+	if genericCacheMap == nil {
+		CreateCache()
+	}
+	genericCacheLock.Lock()
+	defer genericCacheLock.Unlock()
+	return genericCacheMap
+}
+
+func getCRDKindToGroupMap() map[string]string {
+	kindToGroupMap := make(map[string]string)
```
adityadani, line 42: nit: kindToPluralMap ?

pkg/genericcache/genericcache.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			logrus.Warnf("unable to list v1beta1 crds: %v", err)
+			return kindToGroupMap
+		}
+		for _, crd := range crdv1beta1List.Items {
+			if len(crd.Spec.Names.Kind) > 0 {
+				kindToGroupMap[strings.ToLower(crd.Spec.Names.Kind)] = crd.Spec.Names.Plural
+			}
+		}
+	} else {
```
adityadani, line 60: This else is not needed?

pkg/genericcache/genericcache.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// GenericCache is a generic structure which will implement mostly non-k8s resources related information.
+// Individual members get and set need to be provided by the interface.
+type GenericCache interface {
+	// GetCRDKindToPluralMap returns the crdKindToPluralMap from generic cache singleton.
+	GetCRDKindToPluralMap() map[string]string
+
+	// SetCRDKindToPluralMap will set a new CRD's plural in the map
+	SetCRDKindToPluralMap(kind, plural string)
```
adityadani, line 24: Lets rename this to `SetPluralForCRDKind`. The associated Get function returns an actual map, but this function doesn't take a map as an input.

pkg/genericcache/genericcache.go
diptiranjanpx: As discussed kept this for accessing the global variable and also introduced a lock inside struct to avoid accessing the map at same time.

pkg/genericcache/genericcache.go
diptiranjanpx: Agreed. Fixed it.

pkg/genericcache/genericcache.go
diptiranjanpx: Done

pkg/genericcache/genericcache.go
diptiranjanpx: Done

pkg/genericcache/genericcache.go
diptiranjanpx: Done

pkg/pluralmap/pluralmap.go
```diff
@@ -0,0 +1,94 @@
...
+		crdKindToPluralMapCopy[k] = v
+	}
+	return crdKindToPluralMapCopy
+}
+
+func (p *pluralMap) SetPluralForCRDKind(kind, plural string) {
+	p.crdPluralLock.Lock()
+	defer p.crdPluralLock.Unlock()
+
+	crdPluralMap.crdKindToPluralMap[strings.ToLower(kind)] = plural
```
adityadani, line 93: use `p.crdKindToPluralMap` instead

adityadani (approved):
lgtm after comment is addressed

pkg/pluralmap/pluralmap.go
diptiranjanpx: Right, fixed it.

PR #1367, opened 2023-04-17
pb-3789: Add limit value for ListOption while getting resources

lalat-das:
- In clutsers having large count of resources the resource List API in k8s takes a lot of time and that make etcd server to timeout client call. Hence a limit is added to ListOption so that only certain number of resources will be fetched with a single call
- Currently the limit is set to 500 resources at a time.

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: With large resource count GetResource API in resource collector package used to fail. The reason getting  large amount o resource detail via k8s API timesout. Hence we have added the limit value so that only 500 resources will be fetched every List  API call.

```release-note
Issue:
User Impact:
Resolution

**Unit Test**
Currently preparing the setup and will update the PR with unit test results ... 

```

**Does this change need to be cherry-picked to a release branch?**: 23.3.1

Unit Test:
1. Taken a 500 configmap having 60,000 data in each configmap( as it is simulated in ticket)
2.  with 500 limit it failed hence kept it to 200 by changing the config-map entry in stork-controller-config 
3. it passed with right count of resources.
4. Performed a restore it succeed by applying the resources to a new namespace.

backup of the above case:
Screenshot from 2023-04-19 02-16-19

Restore with retain option
image

Multi namespace backup:
image

Additional FC execution will be done post merging the PR to master.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -1451,6 +1451,21 @@ func (a *ApplicationBackupController) backupResources(
 
 	// Don't modify resources if mentioned explicitly in specs
 	resourceCollectorOpts := resourcecollector.Options{}
+	resourceCollectorOpts.ResourceCountLimit = k8sutils.DefaultResourceCountLimit
+	// Read configMap for any user provided value. this will be used in to List call of getResource eventually.
+	resourceCountLimitString, err := k8sutils.GetConfigValue(k8sutils.StorkControllerConfigMapName, metav1.NamespaceSystem, k8sutils.ResourceCountLimitKeyName)
+	if err != nil {
+		logrus.Warnf("error in reading %v cm for the key %v, switching to default value passed to GetResource API: %v",
+			k8sutils.StorkControllerConfigMapName, k8sutils.ResourceCountLimitKeyName, err)
+	} else {
+		if len(resourceCountLimitString) != 0 {
+			resourceCollectorOpts.ResourceCountLimit, err = strconv.ParseInt(resourceCountLimitString, 10, 64)
+			if err != nil {
+				logrus.Warnf("error in conversion of resourceCountLimit: %v", err)
+				resourceCollectorOpts.ResourceCountLimit = k8sutils.DefaultResourceCountLimit
+			}
+		}
+	}
 	if backup.Spec.SkipServiceUpdate {
 		resourceCollectorOpts.SkipServices = true
 	}
@@ -1580,6 +1595,7 @@ func (a *ApplicationBackupController) backupResources(
 			if err != nil {
 				log.ApplicationBackupLog(backup).Errorf("failed to read config map %v's key %v, setting default value of 1MB", k8sutils.StorkControllerConfigMapName,
 					k8sutils.LargeResourceSizeLimitName)
+				largeResourceSizeLimit = k8sutils.LargeResourceSizeLimitDefault
 			}
 		}
 
--- a/pkg/resourcecollector/resourcecollector.go
+++ b/pkg/resourcecollector/resourcecollector.go
@@ -92,6 +92,11 @@ type Options struct {
 	// resource collector to perform transformations on certain k8s resources.
 	// TODO: temporary change required to handle project related transformations
 	RancherProjectMappings map[string]string
+	// limit ensures the k8s APIs will use this value in ListOption while fetching
+	// the resources from cluster. In case of Large number of resources with substantial
+	// content in them, the etcd Server times-out, hence limit will be used to reduce
+	// the number of resources fetched in a single call.
+	ResourceCountLimit int64
 }
 
 // Objects Collection of objects
@@ -306,9 +311,7 @@ func (r *ResourceCollector) GetResourcesForType(
 		default:
 			selectors = labels.Set(labelSelectors).String()
 		}
-		objectsList, err := dynamicClient.List(context.TODO(), metav1.ListOptions{
-			LabelSelector: selectors,
-		})
+		objectsList, err := gatherResourceInChunks(dynamicClient, opts, selectors)
 		if err != nil {
 			return nil, nil, fmt.Errorf("error listing objects for %v: %v", gvr, err)
 		}
@@ -448,9 +451,7 @@ func (r *ResourceCollector) GetResources(
 				default:
 					selectors = labels.Set(labelSelectors).String()
 				}
-				objectsList, err := dynamicClient.List(context.TODO(), metav1.ListOptions{
-					LabelSelector: selectors,
-				})
+				objectsList, err := gatherResourceInChunks(dynamicClient, opts, selectors)
 				if err != nil {
 					if apierrors.IsForbidden(err) {
 						continue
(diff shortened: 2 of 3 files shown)


pkg/resourcecollector/resourcecollector.go
```diff
@@ -306,9 +312,38 @@ func (r *ResourceCollector) GetResourcesForType(
...
+		var err error
+		// Read configMap for any user provided resource count limit to be provided to List call.
+		resourceCountLimit := opts.ResourceCountLimit
+		resourceCountLimitString, err := k8sutils.GetConfigValue(k8sutils.StorkControllerConfigMapName, metav1.NamespaceSystem, k8sutils.ResourceCountLimitKeyName)
+		if err != nil {
+			logrus.Infof("error in reading %v cm for the key %v, switching to default value passed to GetResource API",
+				k8sutils.StorkControllerConfigMapName, k8sutils.ResourceCountLimitKeyName)
+		} else {
+			if len(resourceCountLimitString) != 0 {
+				resourceCountLimit, err = strconv.ParseInt(resourceCountLimitString, 10, 64)
```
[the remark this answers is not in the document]
lalat-das: good catch.. Done the re-assignment.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -448,9 +483,38 @@ func (r *ResourceCollector) GetResources(
 				default:
 					selectors = labels.Set(labelSelectors).String()
 				}
-				objectsList, err := dynamicClient.List(context.TODO(), metav1.ListOptions{
+				var err error
```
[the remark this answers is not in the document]
lalat-das: Done.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -522,6 +524,52 @@ func (r *ResourceCollector) GetResources(
...
+// primarily helps in avoiding server timeout error when the number of resource or size
+// of it large.
+func gatherResourceInChunks(dynamicClient dynamic.ResourceInterface, opts Options, selectors string) (*unstructured.UnstructuredList, error) {
+	var err error
+	fn := "gatherResourceInChunks"
+	// Read configMap for any user provided resource count limit to be provided to List call.
+	resourceCountLimit := opts.ResourceCountLimit
+	resourceCountLimitString, err := k8sutils.GetConfigValue(k8sutils.StorkControllerConfigMapName, metav1.NamespaceSystem, k8sutils.ResourceCountLimitKeyName)
+	if err != nil {
+		logrus.Infof("error in reading %v cm for the key %v, switching to default value passed to GetResource API",
```
siva-portworx, line 538: Please include the err in the debug statement. It is better to change Infof to Warnf or Debugf

pkg/resourcecollector/resourcecollector.go
```diff
@@ -522,6 +524,52 @@ func (r *ResourceCollector) GetResources(
...
+// gatherResourceInChunks() collects all the resources present per ns or resource types.
+// It does it in batch of 500 unless specified by user via a config-map. This function
+// primarily helps in avoiding server timeout error when the number of resource or size
+// of it large.
+func gatherResourceInChunks(dynamicClient dynamic.ResourceInterface, opts Options, selectors string) (*unstructured.UnstructuredList, error) {
+	var err error
+	fn := "gatherResourceInChunks"
+	// Read configMap for any user provided resource count limit to be provided to List call.
+	resourceCountLimit := opts.ResourceCountLimit
+	resourceCountLimitString, err := k8sutils.GetConfigValue(k8sutils.StorkControllerConfigMapName, metav1.NamespaceSystem, k8sutils.ResourceCountLimitKeyName)
```
siva-portworx, line 536: I think reading the ConfigMap should be outside of the resource collector. This might not be applicable to other module, who callout.

pkg/resourcecollector/resourcecollector.go
lalat-das: Done.

pkg/resourcecollector/resourcecollector.go
lalat-das: Sure.. else we will overwrite the value provided by UI. Specifically when they also invoke this independently and expects to use environment variable provided value. Will change the code.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1451,6 +1451,21 @@ func (a *ApplicationBackupController) backupResources(
 
 	// Don't modify resources if mentioned explicitly in specs
 	resourceCollectorOpts := resourcecollector.Options{}
+	resourceCollectorOpts.ResourceCountLimit = k8sutils.DefaultResourceCountLimit
+	// Read configMap for any user provided value. this will be used in to List call of getResource eventually.
+	resourceCountLimitString, err := k8sutils.GetConfigValue(k8sutils.StorkControllerConfigMapName, metav1.NamespaceSystem, k8sutils.ResourceCountLimitKeyName)
```
prashanthpx, line 1456: nit: resourceCountLimitString -> resourceCountLimit, string is not required

pkg/resourcecollector/resourcecollector.go
```diff
@@ -522,6 +523,37 @@ func (r *ResourceCollector) GetResources(
...
+	objectsList := &unstructured.UnstructuredList{}
+	var temp *unstructured.UnstructuredList
+	listOps := metav1.ListOptions{
+		LabelSelector: selectors,
+		Limit:         opts.ResourceCountLimit,
+	}
+	for {
+		temp, err = dynamicClient.List(context.TODO(), listOps)
+		if err != nil {
+			logrus.Infof("%v: failed to list resources", fn)
```
prashanthpx, line 542: nit: should be errorf()

pkg/resourcecollector/resourcecollector.go
```diff
@@ -522,6 +523,37 @@ func (r *ResourceCollector) GetResources(
...
+			return nil, err
+		}
+		if len(temp.GetContinue()) != 0 {
```
prashanthpx, line 545: There could be a potential problem, when we are fetching the last chunk as i understand GetContinue() would be zero as there is nothing more to fetch, in such a case last chunk is missed. Is my understanding correct?

lalat-das: yes  I just figured that out while doing unit testing .. hence added that check now..

pkg/resourcecollector/resourcecollector.go
lalat-das: I made it as an Warnf and anyway the outer function shows the final error anyway

pkg/resourcecollector/resourcecollector.go
px-kesavan: By looking at the code I think, it's already handled and we don't need an additional append. As `temp.GetContinue()` is called before the append. So for the last chunk, the `temp.GetContinue()` will be zero and its appended to the list items and on the next iteration it just avoids the if clause

px-kesavan: Sorry for the noise, i just misunderstood :)

siva-portworx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
lalat-das: that function returns the string , I need to do a  typecast . Moreover I have seen in stork in many places it is done the same way .. Hence keeping it the same way for now

PR #1371, opened 2023-04-19
Testrail autoupdate

Rohit-PX:
**What type of PR is this?**
Integration test.

Autoupdating test results to testrail.

no

no

no


Rohit-PX:
Example of auto-update: https://portworx.testrail.net/index.php?/runs/view/5931&group_by=cases:section_id&group_order=asc

test/integration_test/common_test.go
```diff
@@ -48,6 +48,9 @@ import (
 	_ "github.com/portworx/torpedo/drivers/volume/generic_csi"
 	_ "github.com/portworx/torpedo/drivers/volume/linstor"
 	_ "github.com/portworx/torpedo/drivers/volume/portworx"
+	"github.com/portworx/torpedo/pkg/log"
+	"github.com/portworx/torpedo/pkg/testrailuttils"
+	testrailutils "github.com/portworx/torpedo/pkg/testrailuttils"
```
pp511, line 53: We are importing testrailuttils twice.

test/integration_test/common_test.go
```diff
@@ -117,6 +120,24 @@ const (
...
+	//testRailPasswordFlag        = "testrail-pwd"
+	//testrailMilestoneFlag       = "testrail-milestone"
+
+	testrailRunNameVar         = "TESTRAIL_RUN_NAME"
+	testrailRunIDVar           = "TESTRAIL_RUN_ID"
+	testrailJenkinsBuildURLVar = "TESTRAIL_JENKINS_BUILD_URL"
+	testrailHostVar            = "TESTRAIL_HOST"
+	testrailUserNameVar        = "TESTRAIL_USERNAME"
+	testrailPasswordVar        = "TESTRAIL_PASSWORD"
+	testrailMilestoneVar       = "TESTRAIL_MILESTONE"
```
pp511, line 140: nit.. I think we should remove "Var" from all these variables. It's redundant.

test/integration_test/common_test.go
```diff
@@ -117,6 +120,24 @@ const (
...
+
+	testResultPass = "Pass"
+	testResultFail = "Fail"
+	//testrailRunNameFlag         = "testrail-run-name"
+	//testrailRunIDFlag           = "testrail-run-id"
+	//testrailJenkinsBuildURLFlag = "testrail-jenkins-build-url"
+	//testRailHostFlag            = "testrail-host"
+	//testRailUserNameFlag        = "testrail-usrname"
+	//testRailPasswordFlag        = "testrail-pwd"
+	//testrailMilestoneFlag       = "testrail-milestone"
```
pp511, line 132: Can we remove this?

test/integration_test/common_test.go
```diff
@@ -140,6 +161,11 @@ var storkVersionCheck bool
 var cloudDeletionValidate bool
 var isInternalLBAws bool
 var pxNamespace string
+var storkVersion string
+var testrailHostname string
+var testrailUsername string
```
pp511, line 166: nit.HostName, UserName

test/integration_test/common_test.go
```diff
@@ -1062,6 +1093,94 @@ func IsEks() bool {
...
+		logrus.Infof("Testrail Host: %s", testrailHostname)
+	}
+	testrailUsername = os.Getenv(testrailUserNameVar)
+	if err == nil {
+		logrus.Infof("Testrail Host: %s", testrailUsername)
+	}
+	testrailPassword = os.Getenv(testrailPasswordVar)
+	if err == nil {
+		logrus.Infof("Testrail run name: %s", testrailPassword)
+	}
```
pp511, line 1126: We are checking err == nil but not assigning anything to err. We can check against the actual variables if they are empty.

Suggestion. We can do it this way if you want https://github.com/portworx/pds-agent/blob/master/cmd/server/config.go#:~:text=err%20%3A%3D%20y.Parse(args)
Another repo where this method is used to read env variable https://github.com/portworx/px-object-controller/blob/b579b19efa544385b8136fedd4c6d1a7f9d2ee0d/cmd/px-object-controller/main.go

test/integration_test/stork-test-pod.yaml
```diff
@@ -96,6 +96,20 @@ spec:
...
+    - name: TESTRAIL_RUN_NAME
+      value: testrail_run_name
+    - name: TESTRAIL_RUN_ID
+      value: testrail_run_id
+    - name: TESTRAIL_JENKINS_BUILD_URL
+      value: testrail_jenkins_build_url
+    - name: TESTRAIL_HOST
+      value: testrail_host
+    - name: TESTRAIL_USERNAME
+      value: testrail_uame
```
pp511, line 108: nit.. why uame? should be uname. or even better user_name

test/integration_test/extender_test.go
```diff
@@ -326,6 +386,9 @@ func antihyperconvergenceTestPreferRemoteOnlyTest(t *testing.T) {
 		err = core.Instance().UnCordonNode(schedNode.Name, defaultWaitTimeout, defaultWaitInterval)
 		require.NoError(t, err, "Error uncordorning k8s node for stork test pod")
 	}
+
+	// If we are here then the test has passed
+	testResult = testResultPass
```
pp511, line 391: Why not the log message after test end here?

pp511 (commented):
Mostly looks ok. Some nits that you need to address.

test/integration_test/stork-test-pod.yaml
Rohit-PX: This is specifically named because it interferes with `sed` command we use in the `test-deploy.sh` if we have it as `user_name`

test/integration_test/common_test.go
Rohit-PX: This has been named so, because we want to indentify that these are the host and username for testrail specifically.

test/integration_test/common_test.go
Rohit-PX: Fixed.

test/integration_test/common_test.go
Rohit-PX: Removed.

test/integration_test/common_test.go
Rohit-PX: Makes sense, will check for the values of the actual variables. The reluctance to use the `yag` way is because we should not fail here for the testrail env variables.

test/integration_test/extender_test.go
Rohit-PX: Fixed.

pp511 (approved):
lgmt

pp511:
Approved. Let's create another PR to add more test cases.

PR #1373, opened 2023-04-21
PWX-30291: Removing sts and deployments from stork webhook

pp511:
**What type of PR is this?**
>bug

Remove Stork as the scheduler from Statefulsets and Deployments.

Currently we are not able to set Stork as the scheduler for Statefulsets because we rely on the PVC to be owned by Portworx. STS uses volumeClaimTemplates where PVC is not created when the webhook admission request comes to stork and schedulerName is not updated. Later when webhook admission request arrives for Pods, Stork does update the Schedulername and it works.

I believe there is no reason to update Stork as the scheduler for StatefulSets and Deployments when we have the webhook enabled for Pods. This PR is removing these webhook for Deployments and Statefulsets.

Please note that any annotation set on the Deployment or STS is not passed onto the Pods. To be able to specify stork.libopenstorage.org/disable-admission-controller that disables webhook for the specific STS /Deployment, we need to update it on  spec.templace.metadata.annotations. 

no

```release-note
Issue: Webhook errors observed when trying to mutate statefulsets to update Stork as the scheduler.
User Impact: Customers can not understand this error and gets the impression that Stork webhook is not working
Resolution: Removing wtbhookd for statefulsets and deployments as Stork already has a webhook for Pods that manages setting Stork as the default scheduler.

```

yes 23.5

### DIFF
--- a/pkg/webhookadmission/webhook.go
+++ b/pkg/webhookadmission/webhook.go
@@ -97,7 +97,6 @@ func (c *Controller) processMutateRequest(w http.ResponseWriter, req *http.Reque
 		errMsg := fmt.Errorf("kind=%s not supported", arReq.Kind.Kind)
 		log.Errorf("Failed to serve admission review request: %v", err)
 		c.Recorder.Event(webhookConfig, v1.EventTypeWarning, "invalid admission review request", errMsg.Error())
-		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


diptiranjanpx (approved):

pkg/webhookadmission/webhook.go
```diff
@@ -94,82 +93,46 @@ func (c *Controller) processMutateRequest(w http.ResponseWriter, req *http.Reque
...
-		var pod v1.Pod
-		if err := json.Unmarshal(arReq.Object.Raw, &pod); err != nil {
-			log.Errorf("Could not unmarshal admission review object: %v", err)
-			c.Recorder.Event(webhookConfig, v1.EventTypeWarning, "could not unmarshal ar object", err.Error())
-			http.Error(w, "Decode error", http.StatusBadRequest)
+	if arReq.Kind.Kind != "Pod" {
+		errMsg := fmt.Errorf("kind=%s not supported", arReq.Kind.Kind)
+		log.Errorf("Failed to serve admission review request: %v", err)
+		c.Recorder.Event(webhookConfig, v1.EventTypeWarning, "invalid admission review request", errMsg.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
adityadani, line 100: Should we just log the event on the requesting object and not return an error here? Why block a request if we cannot process it.

adityadani (approved):
looks good - make sure to test upgrades

PR #1377, opened 2023-04-27
pb-3448: Support for restore with NFS backuplocation

siva-portworx:
**What type of PR is this?**
>feature
NFS backuplocation support.

```
 pb-3448: Support for restore with NFS backuplocation

            - In restoreVolume cleaned up the cases to have only nfs and non nfs for
              calling driver.startRestore
            - changed getRestoreStorageClasses and downloadObject to get the
              storageclass.json in the case of NFS backuplocation directly from mount NFS pvc
            - Removed some of the conversion function, which is not required.
                      - convertResourceVolInfoToAppBkpVolInfo
                      - convertResourceVolInfoToAppBkpVolInfo
```

No

N.A
N.A

Testing:
Basic testing of Native CSI, Direct KDMP and PX backup/restore was done with NFS backuplocation.

### DIFF
--- a/drivers/volume/aws/aws.go
+++ b/drivers/volume/aws/aws.go
@@ -122,11 +122,6 @@ func (a *aws) OwnsPVCForBackup(
 		// If user has forced the backupType in config map, default to generic always
 		return false
 	}
-	// For AWS volume and backuplocation type is NFS, we will not own.
-	// It will default to kdmp
-	if blType == storkapi.BackupLocationNFS {
-		return false
-	}
 	return a.OwnsPVC(coreOps, pvc)
 }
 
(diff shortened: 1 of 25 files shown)


drivers/volume/aws/aws.go
```diff
@@ -122,11 +122,6 @@ func (a *aws) OwnsPVCForBackup(
 		// If user has forced the backupType in config map, default to generic always
 		return false
 	}
-	// For AWS volume and backuplocation type is NFS, we will not own.
-	// It will default to kdmp
-	if blType == storkapi.BackupLocationNFS {
```
lalat-das, line 127: We may need to remove this argument blType from  function call altogether.. I guess it is no more used anywhere now ?

lalat-das (approved):

siva-portworx:
Closing the PR as it was created from incorrect branch.

PR #1382, opened 2023-04-27
PWX-30280: application test for nearsync

dgoel-px:
**What type of PR is this?**
integration-test

The PR adds application tests with postgres, cassandra and elasticsearch for nearsync

No

No

No


test/integration_test/nearsync_application_test.go
```diff
@@ -0,0 +1,126 @@
...
+	stressAppKey         string
+	instanceId           string
+	repl                 int
+	nearsyncReplStrategy string
+}
+
+func TestApplicationsWithNearSync(t *testing.T) {
+
+	// create secret on source and destination
+	funcCreateSecret := func() {
```
Rohit-PX, line 29: nit: just call it `createSecret()`

test/integration_test/nearsync_application_test.go
```diff
@@ -0,0 +1,126 @@
...
+func helperTestAppWithNearSync(t *testing.T, config configTestAppWithNearSync) {
+	instanceIDs := []string{config.instanceId}
+	storageClass := "px-sc"
+	migrationName := "failover-migration"
+	actionName := "failover-action"
+	namespaces := getNamespaces(instanceIDs, config.appKey)
+	cleanup(t, namespaces[0], storageClass)
+
+	// create storage class
+	_, err := createStorageClass(StorageClass{
```
Rohit-PX, line 72: get the storage class object here and use it in the rest of the test, that way we have the actual object to work with.

test/integration_test/nearsync_application_test.go
dgoel-px: Because its called twice, once on source and once on destination

test/integration_test/nearsync_application_test.go
```diff
@@ -0,0 +1,126 @@
...
+		startFailover(t, actionName, namespaces)
+		validateFailover(t, actionName, namespaces, preMigrationCtxs, true)
+	})
+}
+
+func updateClusterDomain(t *testing.T, activate bool) {
+	storageNode := node.GetStorageDriverNodes()[0]
+	var cmd string
+	if activate {
+		cmd = "cluster domains activate --name dc1"
```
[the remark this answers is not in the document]
dgoel-px: Yes this change was on my radar. Going to fix it

test/integration_test/nearsync_application_test.go
dgoel-px: I don't use the storage class object anywhere in this test. It used automatically by the application specs.

Rohit-PX (approved):

PR #1387, opened 2023-05-01
PB-3827 - Fix loop in cleanupSnapshotsForRestore to avoid assignment of the last entry to all keys

px-kesavan:
**What type of PR is this?**
>bug

**What this PR does / why we need it**: Fixes the for loop in cleanupSnapshotsForRestore to avoid assignment of the last entry to all keys

```release-note
Issue:
User Impact:
Resolution

```

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -653,7 +653,8 @@ func (c *csi) cleanupSnapshotsForRestore(
 			if err != nil {
 				return err
 			}
-			for _, vs := range vsList.Items {
+			for _, val := range vsList.Items {
+				vs := val
 				vsMap[vs.Name] = &vs
 			}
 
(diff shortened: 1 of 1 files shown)


drivers/volume/csi/csi.go
```diff
@@ -684,7 +684,8 @@ func (c *csi) cleanupSnapshotsForRestore(
 		if err != nil {
 			return err
 		}
-		for _, vs := range vsList.Items {
+		for _, val := range vsList.Items {
+			vs := val
```
siva-portworx, line 688: We need to fix it in c.v1SnapshotRequired. case as well.

px-kesavan: fixed

siva-portworx (approved):

PR #1388, opened 2023-05-02
PWX-30706: Use same cmdexecutor image tag as stork

pp511:
**What type of PR is this?**
>bug

Use updated cmdExecutorImage.https://portworx.atlassian.net/browse/PWX-30706

no

```release-note
Issue: Older image of Cmdexecutor was being referenced from Stork  
User Impact: Any new improvements in cmdexecuter image were not being used in Stork.
Resolution: Stork is now updated to use the same tag of cmdexecutor

```

yes 23.5

### DIFF
--- a/pkg/rule/rule.go
+++ b/pkg/rule/rule.go
@@ -40,7 +40,8 @@ const (
 	// environment variable for command executor image registry and image registry secret.
 	cmdExecutorImageRegistryEnvVar       = "CMD-EXECUTOR-IMAGE-REGISTRY"
 	cmdExecutorImageRegistrySecretEnvVar = "CMD-EXECUTOR-IMAGE-REGISTRY-SECRET"
-	defaultCmdExecutorImage              = "cmdexecutor:0.1"
+	defaultCmdExecutorImage              = "cmdexecutor"
+	defaultCmdExecutorTag                = "0.1"
 	// annotation key value for command executor image registry and image registry secret.
 	cmdExecutorImageOverrideKey          = "stork.libopenstorage.org/cmdexecutor-image"
 	cmdExecutorImageOverrideSecretKey    = "stork.libopenstorage.org/cmdexecutor-image-secret"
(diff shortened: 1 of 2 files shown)


pkg/rule/rule.go
```diff
@@ -343,22 +344,22 @@ func executeCommandAction(
...
+			cmdExecutorImage = registry + "/" + defaultCmdExecutorImage + ":" + imageTag
 		} else {
-			cmdExecutorImage = defaultCmdExecutorImage
+			cmdExecutorImage = defaultCmdExecutorImage + ":" + defaultCmdExecutorTag
 		}
 		cmdExecutorImageSecret = registrySecret
 	} else {
 		// if env is set get it from env variable.
-		cmdExecutorImage = os.Getenv(cmdExecutorImageRegistryEnvVar) + "/" + defaultCmdExecutorImage
+		cmdExecutorImage = os.Getenv(cmdExecutorImageRegistryEnvVar) + "/" + defaultCmdExecutorImage + ":" + defaultCmdExecutorTag
```
pp511, line 362: Here 0.1 version of cmdexecutor is still being pulled.  I am assuming when cmdExecutorImageRegistryEnvVar is used then annotation cmdExecutorImageOverrideKey is used to override the image. If needed a new environment variable to override cmdexecutor image can be introduced but I am not sure how it will be set.

pkg/rule/rule.go
```diff
@@ -343,22 +344,22 @@ func executeCommandAction(
 			return err
 		}
 		// If env is not set get the values from stork deployment spec.
-		registry, registrySecret, err := k8sutils.GetImageRegistryFromDeployment(
+		registry, registrySecret, imageTag, err := k8sutils.GetImageInfoFromDeployment(
```
diptiranjanpx, line 347: If we will try with any custom stork image tag, then the cmdexecutor tag need to be present also in that repo.

pp511: @diptiranjanpx If your usecase needs cmdexecutor yes. Else let me know if there is a better wat.

adityadani: Since creating custom images is not standard procedure, we can expect the cmd executor image to be also built with the same image tag, or we can ask them to use the `cmdExecutorImageOverrideKey ` annotation

pkg/rule/rule.go
adityadani: I think we can stick with just the annotations for now. We should push for not using any of these env variables or annotations. If a registry is being used then the same registry will be used for both stork and cmdexecutor and we should just push for using the same tags.

adityadani (approved):

PR #1389, opened 2023-05-02
adding name of resource that were not restored on applicationrestore.go

camartinez04:
**What type of PR is this?**

improvement

On pkg/applicationmanager/controllers/applicationrestore.go line 1608 we need to check which resource were not successfully restored.

no

```restore.Status.Reason = fmt.Sprintf("Volumes were restored successfully. Some existing resources were not replaced, resource: %s", resource.Name)```

no

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -1605,7 +1605,7 @@ func (a *ApplicationRestoreController) restoreResources(
 	for _, resource := range restore.Status.Resources {
 		if resource.Status != storkapi.ApplicationRestoreStatusSuccessful {
 			restore.Status.Status = storkapi.ApplicationRestoreStatusPartialSuccess
-			restore.Status.Reason = "Volumes were restored successfully. Some existing resources were not replaced"
+			restore.Status.Reason = fmt.Sprintf("Volumes were restored successfully. Some existing resources were not replaced, resource: %s", resource.Name)
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/applicationmanager/controllers/applicationrestore.go [hunk 1]
```diff
@@ -1605,7 +1605,7 @@ func (a *ApplicationRestoreController) restoreResources(
 	for _, resource := range restore.Status.Resources {
 		if resource.Status != storkapi.ApplicationRestoreStatusSuccessful {
 			restore.Status.Status = storkapi.ApplicationRestoreStatusPartialSuccess
-			restore.Status.Reason = "Volumes were restored successfully. Some existing resources were not replaced"
+			restore.Status.Reason = fmt.Sprintf("Volumes were restored successfully. Some existing resources were not replaced, resource: %s", resource.Name)
```
siva-portworx, line 1608: This is the overall status of the restore. So we are avoiding giving details of the specific resource. We do have resource level status, where the retained status is maintained. If we add the resource name here, it will confusing that only because of this resource alone, it is partial success.

camartinez04: @siva-portworx I was thinking on that, it would require to change our Struct, changing the field restore.Status.reason to a slice of Strings, filling that list with the non-restored objects (with a previous check, if all of them failed, just inform that) and non stopping the for loop at the first occurrence as we currently do. All that sounds like lot of effort. I see, probably a message of where the user can find more details would be useful.

PR #1390, opened 2023-05-03
PWX-30712: Initializing shared informer cache for all stork Pods

pp511:
**What type of PR is this?**
>bug

Shared informer cache should be initialized for all Stork Pods.
We need to register applciation controller CRDs explicitly before initializing Cache.

no

yes 23.4

**Test**:
Migration Jenkins Job: https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.4-dev/job/23-4-dev-migration-k8s-1-24-0/

### DIFF
--- a/cmd/stork/stork.go
+++ b/cmd/stork/stork.go
@@ -356,11 +356,10 @@ func run(c *cli.Context) {
 	}
 	// Registering application registration CRDs explicitly to use
 	// shared informer cache for caching application-controller CRs
-	if c.Bool("application-controller") {
-		if err := applicationmanager.CreateCRD(); err != nil {
-			log.Fatalf("Error creating CRDs for application manager: %v", err)
-		}
+	if err := applicationmanager.CreateCRD(); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/stork/stork.go
```diff
@@ -354,6 +354,20 @@ func run(c *cli.Context) {
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
 		log.Fatalf("Setup scheme failed for stork resources: %v", err)
 	}
+	// Registering application registration CRDs explicitly to use
+	// shared informer cache for caching application-controller CRs
+	if c.Bool("application-controller") {
```
adityadani, line 359: Lets create the CRDs regardless of whether application controller is enabled or not, since the cache's APIs are not guarded by this flag.

adityadani (approved):

cmd/stork/stork.go
pp511: Agree. Updated. Will merge and cherry-pick after successful CBT run.

PR #1391, opened 2023-05-03
PWX-30712: Initializing shared informer cache for all stork Pods (#1390)

pp511:
**What type of PR is this?**
>bug

Shared informer cache should be initialized for all Stork Pods.
We need to register applciation controller CRDs explicitly before initializing Cache.

no

yes 23.4


pp511:
Missed running CBT for this on 23.4. Created a dummy branch to make sure CBTs pass on 23.4 branch.  All tests passing https://github.com/libopenstorage/stork/pull/1392

PR #1393, opened 2023-05-03
pb-3831: In csi driver CancelRestore, cleaning up dangling PVC, if restore is not succeed and retain as well

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3831: In csi driver CancelRestore, cleaning up dangling PVC, if restore is not succeed and retain as well
```

no

```release-note
Issue: In CSI, in-place restore, the retained PVC are getting deleted after restore.
User Impact: After in-place CSI restore, the PVC stork in Terminating state and app is not usable.
Resolution: With the fix, the in-place CSI restore, the existing PVC will not deleted and remain unchanged.

```

Need decide.

Tested on NFS setup.

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -1683,7 +1683,10 @@ func (c *csi) CancelRestore(restore *storkapi.ApplicationRestore) error {
 		if vrInfo.DriverName != storkvolume.CSIDriverName {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


px-kesavan (approved):

diptiranjanpx (approved):

lalat-das (approved):

siva-portworx:
All test passed in the previous cycle. Rebasing of branch triggered all the test again. So force merging it.

PR #1395, opened 2023-05-04
PWX-29422: Providing option to specify namespaces selector in migrati…

pp511:
…on spec.

**What type of PR is this?**
>feature

https://portworx.atlassian.net/browse/PWX-29422
Allow users to specify namespace selectors in migration spec.

yes, 

Yes
* You can now provide namespace labels in MigrationSchedule spec to specify namespaces for migration.

yes 23.5.0

**Notes**:
* I have added one function getMigrationNamespaces to get all unique namespaces for migration and referencing this list throughout our migration flow.
* 
 ```
MigrationSchedule ->
spec:
  template:
    spec:
        namespaceSelectors:
          ns: migrate

Namespace ->
kind: Namespace
metadata:
  labels:
    kubernetes.io/metadata.name: mysql
    ns: migrate
```
* Triggered  migration Jenkins job https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.4-dev/job/23-4-dev-migration-k8s-1-24-0/463/ to make sure there are no regressions

### DIFF
--- a/pkg/apis/stork/v1alpha1/migration.go
+++ b/pkg/apis/stork/v1alpha1/migration.go
@@ -16,6 +16,7 @@ type MigrationSpec struct {
 	ClusterPair                  string            `json:"clusterPair"`
 	AdminClusterPair             string            `json:"adminClusterPair"`
 	Namespaces                   []string          `json:"namespaces"`
+	NamespaceSelectors           map[string]string `json:"namespaceSelectors"`
 	IncludeResources             *bool             `json:"includeResources"`
 	IncludeVolumes               *bool             `json:"includeVolumes"`
 	StartApplications            *bool             `json:"startApplications"`
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -548,11 +548,9 @@ func (m *MigrationController) getMigrationNamespaces(ctx context.Context, migrat
 	uniqueNamespaces := make(map[string]bool)
 
 	for _, ns := range migration.Spec.Namespaces {
-		if _, ok := uniqueNamespaces[ns]; !ok {
-			migrationNamespaces = append(migrationNamespaces, ns)
-		}
 		uniqueNamespaces[ns] = true
 	}
+
 	config, err := rest.InClusterConfig()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: pkg/migration/controllers/migration.go [hunk 1]
```diff
@@ -535,8 +543,44 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
...
+		label := key + "=" + val
+		namespaces, err := clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{LabelSelector: label})
+		if err != nil {
+			return nil, err
+		}
+		for _, namespace := range namespaces.Items {
+			if _, ok := uniqueNamespaces[namespace.GetName()]; !ok {
+				migrationNamespaces = append(migrationNamespaces, namespace.GetName())
+			}
+			uniqueNamespaces[namespace.GetName()] = true
```
diptiranjanpx, line 575: Another approach could have been to only populate `uniqueNamespaces` map with namespaces and at the end create the list of keys and return that. The current approach is fine too.

diptiranjanpx (approved):

test/integration_test/migration_test.go
```diff
@@ -456,6 +457,19 @@ func migrationLabelExcludeSelectorTest(t *testing.T) {
...
+	triggerMigrationTest(
+		t,
+		"ns-selector-test",
+		"cassandra",
+		[]string{"mysql-1-pvc"},
+		"namespace-selector-migration",
+		true,
+		true,
+		true,
+	)
```
adityadani, line 470: Lets add another app in a different namespace and ensure that that app is not migrated and cassandra gets migrated

pkg/migration/controllers/migration.go
pp511: Updated as per suggestion.

test/integration_test/migration_test.go
pp511: I have modified the test to run two apps in different namespace and only migrate cassandra for which namespace label matches.

pkg/apis/stork/v1alpha1/migration.go
```diff
@@ -16,6 +16,7 @@ type MigrationSpec struct {
 	ClusterPair                  string            `json:"clusterPair"`
 	AdminClusterPair             string            `json:"adminClusterPair"`
 	Namespaces                   []string          `json:"namespaces"`
+	NamespaceSelectors           map[string]string `json:"namespaceSelectors"`
```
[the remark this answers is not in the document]
adityadani: You are right. My bad

adityadani (approved):

PR #1408, opened 2023-05-18
PWX-31149: Use the correct node object in stork monitor.

adityadani:
**What type of PR is this?**
>bug

Stork would delete a pod on a node which was detected Offline but which is turned Online during the exponential backoff wait. A recent change introduced this regression in 23.5

No

No - This is a regression in 23.5-dev

Yes - 23.5

### DIFF
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -235,7 +235,7 @@ func (m *Monitor) cleanupDriverNodePods(node *volume.NodeInfo) {
 		if err != nil {
 			return false, nil
 		}
-		if node.Status == volume.NodeOffline || node.Status == volume.NodeDegraded {
+		if n.Status == volume.NodeOffline || n.Status == volume.NodeDegraded {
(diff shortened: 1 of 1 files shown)


pureneelesh (approved):

pkg/monitor/monitor.go
```diff
@@ -235,7 +235,7 @@ func (m *Monitor) cleanupDriverNodePods(node *volume.NodeInfo) {
 		if err != nil {
 			return false, nil
 		}
-		if node.Status == volume.NodeOffline || node.Status == volume.NodeDegraded {
+		if n.Status == volume.NodeOffline || n.Status == volume.NodeDegraded {
```
harsh-px, line 238: Can we check why existing UTs isn't hitting this line?

adityadani: There is no UT or an integration test exercising a node which goes Offline but comes back Online within the monitor timeout. The current UT framework just fakes data, so I would go for an integration test to exercise this scenario. cc @Rohit-PX

PR #1411, opened 2023-05-22
pb-3891: updating the resourcount of restore CR in updateResourceStatus.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3891: updating the resourcount of restore CR in updateResourceStatus.
```

no

no

no
Currently the PV and PVC counts of CSI and kdmp resources are not getting updated in the final count.


lalat-das (approved):
lgtm

PR #1428, opened 2023-06-06
pb-3823: Make the pre-exec cmd-executor POD session alive

lalat-das:
- Certain APP like MYSQL unlocks the tables if the session in which the DB is locked, gets over.
- We have defered the killing of pre-exec POD till all the volume snapshot is taken thereby ensuring DB is not in inconsistent state during snapshot process

>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: It ensures the session for the MYSQL kind of APP to linger so that DB don't get unlocked in the middle of volumesnapshot process.

**Does this PR change a user-facing CRD or CLI?**: NO

**Is a release note needed?**: NO

```release-note
Issue:
User Impact:
Resolution

```

image


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -122,6 +122,7 @@ type ApplicationBackupController struct {
 	recorder             record.EventRecorder
 	resourceCollector    resourcecollector.ResourceCollector
 	backupAdminNamespace string
+	terminationChannels  []chan bool
```
adityadani, line 125: The controller handles different ApplicationBackup CRs at the same time. You need to maintain a separate terminationChannels slice for each applicationBackup CR. 
```
terminationChannelsMap map[string][]chan bool // key is the ApplicationBackup.UID
```

adityadani (changes requested):

pkg/applicationmanager/controllers/applicationbackup.go
lalat-das: Good catch @adityadani , will chnage it to a map with value as this slice..

lalat-das:
Closing the PR as 2.5.0-nfs-final is merged to master stork branch hence this PR showing unwanted commits now.. Created a fresh one against master with same bug pb-3823

pkg/applicationmanager/controllers/applicationbackup.go
lalat-das: @adityadani  I have addressed this comment in the latest PR that is raised against master.

PR #1433, opened 2023-06-10
PB-3919 : fix to handle project annotations/labels for the px-backup created namespace

px-kesavan:

**What this PR does / why we need it**: fix to handle project annotations/labels for the px-backup created namespace.
Add all labels/annotations from Namespace.json except project labels/annotations when not found in the namespace which is not created by px-backup.
With retain policy, project labels/annotations should not be applied to the namespace with no projects. Applies to a scenario where project association is removed by the user after taking the backup with project on a namespace


px-kesavan:
Closing this PR , will be tracked in https://github.com/libopenstorage/stork/pull/1443

PR #1436, opened 2023-06-13
pb-3823: Make the pre-exec cmd-executor POD session alive during snapshot process

lalat-das:
- Certain APPs like MYSQL unlocks the tables if the session in which the DB is locked, is ended.
- We have deferred the killing of pre-exec POD till all the volume snapshot is taken thereby ensuring DB is not in inconsistent state during snapshot process

**What type of PR is this?** BUG
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: Because pre-exec pod which waits for volume backup  to get over gets killed untimely it leads to data inconsistsency.  Though it has executed the necessary commands to lock & stop the IO but POD eviction unlocks the mysql App DB which triggers the inconsistency. Hence fixed the untimely killing of pre-exec pod. 

**Does this PR change a user-facing CRD or CLI?**: NO

```release-note
Issue: mysql app can show inconsistent data after being restored. 
User Impact: All mysql backup may fail to run mysql application after restore operation due to data inconsistency
Resolution: Fixed the data inconsistency by holding the table lock for a required interval while backup is in progress.

```

**Does this change need to be cherry-picked to a release branch?**: 23.6

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -122,6 +122,7 @@ type ApplicationBackupController struct {
 	recorder             record.EventRecorder
 	resourceCollector    resourcecollector.ResourceCollector
 	backupAdminNamespace string
+	terminationChannels  map[string][]chan bool
 	reconcileTime        time.Duration
 }
 
@@ -138,6 +139,7 @@ func (a *ApplicationBackupController) Init(mgr manager.Manager, backupAdminNames
 		return err
 	}
 	a.reconcileTime = time.Duration(syncTime) * time.Second
+	a.terminationChannels = make(map[string][]chan bool)
 	return controllers.RegisterTo(mgr, "application-backup-controller", a, &stork_api.ApplicationBackup{})
 }
 
@@ -308,7 +310,6 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		return nil
 	}
 
-	var terminationChannels []chan bool
 	var err error
 
 	if a.setDefaults(backup) {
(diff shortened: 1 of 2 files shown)


lalat-das:
Currently working on fixing portworx volume scenario where in snapshot is asynchronous, hence waiting for it get over before killing the pre-exec POD or execution of post-exec rules.

lalat-das:
> Currently working on fixing portworx volume scenario where in snapshot is asynchronous, hence waiting for it get over before killing the pre-exec POD or execution of post-exec rules.

Done with unite testing and coding for portworx too

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -705,37 +701,64 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+							volInfos, err := driver.GetBackupStatus(backup)
+							if err != nil {
+								return fmt.Errorf("error getting backup status for driver %v: %v", driverName, err)
+							}
+							for _, volInfo := range volInfos {
+								if volInfo.BackupID == "" {
+									log.ApplicationBackupLog(backup).Tracef("Snapshot of volume %v is not done yet, need to loop till snapshot is done...", volInfo.PersistentVolumeClaim)
+									snapshotNotCompleted = true
+								}
+							}
```
siva-portworx, line 720: Can we add a logic to add the CR update to avoid px-backup 30 mint timeout as it is a infinite for loop?

siva-portworx (approved):

prashanthpx (approved):

PR #1444, opened 2023-06-20
PWX-30577: Changes for getting px-auth-token for getting clusterpair token call.

diptiranjanpx:
**What type of PR is this?**
improvement

1. Getting px-auth-token for auth enabled clusters.
2. Removing backuplocations creation in destination as this will be done as part of bidirectional clusterpair with storkctl.

**Test**
Test results have been updated in the ticket.

### DIFF
--- a/pkg/storkctl/clusterpair.go
+++ b/pkg/storkctl/clusterpair.go
@@ -33,6 +33,9 @@ const (
 	gcloudPath             = "./google-cloud-sdk/bin/gcloud"
 	gcloudBinaryName       = "gcloud"
 	skipResourceAnnotation = "stork.libopenstorage.org/skip-resource"
+	pxAdminTokenSecret     = "px-admin-token"
+	secretNamespace        = "openstorage.io/auth-secret-namespace"
+	secretName             = "openstorage.io/auth-secret-name"
 )
 
 var (
@@ -247,6 +250,8 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 	var azureAccountName, azureAccountKey string
 	var googleProjectID, googleJSONKey string
 	var syncDR bool
+	var pxAuthTokenSrc, pxAuthSecretNamespaceSrc string
+	var pxAuthTokenDest, pxAuthSecretNamespaceDest string
 
 	createClusterPairCommand := &cobra.Command{
 		Use:   clusterPairSubcommand,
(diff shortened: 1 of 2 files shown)


pkg/storkctl/clusterpair.go
```diff
@@ -740,11 +827,47 @@ func getPXToken(config, pxEndpoint string) (string, error) {
 	return token, nil
 }
 
+func getPXAuthToken(configFile string) (string, string, error) {
+	var authToken, authSecretNamespace string
+	// Create cluster-pair on source cluster
```
adityadani, line 832: remove comment

pkg/storkctl/clusterpair.go
```diff
@@ -267,10 +274,14 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 				return
 			}
 
-			if len(sIP) > 0 || len(sPort) > 0 || len(dIP) > 0 || len(dPort) > 0 {
+			if len(sIP) > 0 || len(sPort) > 0 || len(dIP) > 0 || len(dPort) > 0 || len(srcToken) > 0 || len(destToken) > 0 {
```
adityadani, line 277: Lets keep the option for the user to specify `srcToken` & `destToken` if for some reason the automatic token fetch fails. You can still mark it as deprecated in the CLI

pkg/storkctl/clusterpair.go
```diff
@@ -740,11 +827,47 @@ func getPXToken(config, pxEndpoint string) (string, error) {
...
+			}
+
+			if secretData, ok = authTokenSecret.Data["auth-token"]; !ok {
+				return authToken, authSecretNamespace, fmt.Errorf("invalid secret key data")
+			}
+			return string(secretData), secret.Namespace, nil
+		}
+	}
+
+	// If PX_AUTH_TOKEN secret is not found, return empty token as it is not secured cluster.
```
adityadani, line 862: nit: if token secret is not found ...

adityadani (approved):

pkg/storkctl/clusterpair.go
diptiranjanpx: Done

pkg/storkctl/clusterpair.go
diptiranjanpx: Done

pkg/storkctl/clusterpair.go
diptiranjanpx: srcToken and destToken is deprecated now (it will show as deprecated but values will be honoured but in --help these options will not show up.

diptiranjanpx: As discussed , removing `MarkDeprecated` for `srcToken` and `destToken`. These will be optional params.

PR #1445, opened 2023-06-23
pb-3966: Updating the restore PVC size only when the volumesnapshot size is greater than PVC size

siva-portworx:
**What type of PR is this?**
>bug

```
pb-3966: Updating the restore PVC size only when the volumesnapshot size is greater than PVC size
```

No

```release-note
Issue: CSI restore fails for some of the storage provider, when the volumesnapshot size is less the PVC size
User Impact: User will not able to do CSI restore with some storage provider.
Resolution: Fixed such the restore PVC size will be always equal to source volume size or greater then source volume size, if the volumesnapshot size is greater then the source volume size

```

23.6 branch.

Tested:
1) Tested kdmp+localsnapshot backup and restore on QA IBM setup.


siva-portworx:
Force merging the PR as QA need to start testing.

PR #1454, opened 2023-07-17
pb-4000: Increased the value of shortRetryTimeout to 5mints.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-4000: Increased the value of shortRetryTimeout to 5mints.

            - On some of the platform, the deletion of the job pod that we
              create to temporarily to bound the "waitForFirstConsumer" volumebind mode
              PVC takes time.
            - So increased the timeout value for the waiting for the bound
              to be deleted to 5 minutes from 30 sec.
```

No

```release-note
Issue: In some cases, the bound pod deletion takes time, So current timeout value of 30 sec is not enough and backup fails.
User Impact: Some time the backup of the volume whose volume bind mode with "WaitForFirstConsumer" fails with timeout error.
Resolution: Increased the timeout value to 5 minutes, So that deletion of the bound pod created to bind the volume with "WaitForFirstConsumer" will not fail with timeout error.

```

23.6.1 branch.

Testing:
Validated the fix on the azure setup, where we were seeing timeout failure.

### DIFF
--- a/pkg/snapshotter/snapshotter_csi.go
+++ b/pkg/snapshotter/snapshotter_csi.go
@@ -61,9 +61,9 @@ const (
 	// will incorporate in their names
 	pvcNameLenLimitForJob = 48
 	// shortRetryTimeout gets used for retry timeout
-	shortRetryTimeout = 30 * time.Second
+	shortRetryTimeout = 300 * time.Second
 	// shortRetryTimeout gets used for retry timeout interval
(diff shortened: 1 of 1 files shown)


lalat-das (approved):
lgtm

pallav-px (approved):

siva-portworx:
Only code coverage failed. So force pushing it.

PR #1460, opened 2023-07-25
pb-4093 Deleting the volumesnapshotcontent in the case of failed native CSI backup.

siva-portworx:
**What type of PR is this?**
>bug

```
     pb-4093 Deleting the volumesnapshotcontent in the case of failed native
    CSI backup

            - Adding the vsc name in the map in CleanupBackupResources with
              out any status check ( success/failure/in-progress)
            - In SnapshotStatus() also removed the check for ready state to
              add the volumesnapshotcontent name in the snapshotInfo. If the status
              volumesnapshotcontent is present, we will add the VSC name in the snapshotInfo.
```

no

```release-note
Issue: The volumesnapshotcontent are not getting deleted when the native CSI backup fails with the timeout error.
User Impact: The volumesnapshotcontent will get accumulated in the case of native CSI backup failures
Resolution: Taking care of deleting the volumesnapshotcontent in the case of failure as well.

```

Yes. 
23.7 branch and 23.7.1 branch.

**Testing:
Vaidated the fix on IBM setup with timeout error.**

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -1906,14 +1906,12 @@ func (c *csi) CleanupBackupResources(backup *storkapi.ApplicationBackup) error {
 				continue
 			}
 			vsMap.(map[string]*kSnapshotv1.VolumeSnapshot)[vInfo.BackupID] = snapshot
-			if snapshotInfo.Status == snapshotter.StatusReady {
-				snapshotContent, ok := snapshotInfo.Content.(*kSnapshotv1.VolumeSnapshotContent)
-				if !ok {
-					logrus.Warnf("failed to map volumesnapshotcontent object")
-					continue
-				}
-				vsContentMap.(map[string]*kSnapshotv1.VolumeSnapshotContent)[vInfo.BackupID] = snapshotContent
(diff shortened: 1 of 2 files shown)


px-kesavan (approved):

diptiranjanpx (commented):
Is similar change required in px-backup delete path also?

siva-portworx:
> Is similar change required in px-backup delete path also?

In the case of delete path, if the snapshot itself fails, we will not have snapshot.json. Then we will not apply any VS or VSC.
In the case of success, it work as it is.

diptiranjanpx:
> > Is similar change required in px-backup delete path also?
> 
> In the case of delete path, if the snapshot itself fails, we will not have snapshot.json. Then we will not apply any VS or VSC. In the case of success, it work as it is.

In case the backup was successful, but during delete the volumesnapshotcontent did not become ready, are we going to delete the applied VS and VSC CRs even though they are in failed condition? If yes, then it looks fine.

diptiranjanpx (approved):

siva-portworx:
Merge the PR as Functional test is blocking. Will track, it if we see any failure.

PR #1461, opened 2023-07-25
PWX-32496: Modifying input param "google-json-key" to "google-json-key-file" for creating clusterpair with google objectstore.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

Modifying "google-json-key" to "google-json-key-file" to have better understanding for the input parameter that it requires a file path for input.

### DIFF
--- a/pkg/storkctl/clusterpair.go
+++ b/pkg/storkctl/clusterpair.go
@@ -412,7 +412,7 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 					return
 				}
 				if len(googleJSONKey) == 0 {
-					util.CheckErr(getMissingParameterError("google-json-key", "Json key file missing for Google"))
+					util.CheckErr(getMissingParameterError("google-key-file-path", "Json key file missing for Google"))
 					return
 				}
 				// Read the jsonkey file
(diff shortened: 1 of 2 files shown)


pkg/storkctl/clusterpair.go
```diff
@@ -623,7 +623,7 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 	createClusterPairCommand.Flags().StringVar(&azureAccountKey, "azure-account-key", "", "Account key for Azure")
 	// Google
 	createClusterPairCommand.Flags().StringVar(&googleProjectID, "google-project-id", "", "Project ID for Google")
-	createClusterPairCommand.Flags().StringVar(&googleJSONKey, "google-json-key", "", "Json key for Google")
+	createClusterPairCommand.Flags().StringVar(&googleJSONKey, "google-json-key-file", "", "Json key for Google")
```
adityadani, line 626: We can be more explicit and say 
--google-key-file-path

adityadani (approved):

pkg/storkctl/clusterpair.go
diptiranjanpx: Done

PR #1464, opened 2023-07-26
pb-4106: added nil ptr check while accessing snapshot.Status in SnapshotStatus() api.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-4106: added nil ptr check while accessing snapshot.Status in SnapshotStatus() api.
```

No

No, as it is regression from previous commit.

Yes to 2.7 and 2.7.1

Testing:
Validate a successful Native CSI backup with the fix.
Screenshot 2023-07-26 at 6 21 26 PM


diptiranjanpx (approved):

lalat-das (approved):

siva-portworx:
Merging it as Functional test is blocked. Will track it, if fails.

PR #1468, opened 2023-08-08
PWX-31012: Provide the PVC namespace and name labels while restoring Portworx backup.

adityadani:
**What type of PR is this?**
>improvement

- The name and namespace labels are provided as part of the Portworx Cloudsnap Restore request.
- The name of the PVC stays the same as it was when the backup was taken. However the namespace can change if the user decides to restore in a different namespace.
- These labels ensure Portworx driver knows which namespace the volume is getting restored to.

No

Yes
```release-note
Portworx driver will ensure the correct pvc and namespace labels are applied on the Portworx volume while restoring from an ApplicationBackup.

```

Yes - 23.8


diptiranjanpx (approved):
LGTM

pp511 (approved):
lgtm

PR #1469, opened 2023-08-09
Added test for operator migration with CRs 

gejain:
**What type of PR is this?**
>integration-test

Added test for operator migration with CRs 
https://portworx.atlassian.net/browse/PTX-19446

Local run result -
--- PASS: TestOperatorMig (716.23s)
    --- PASS: TestOperatorMig/testoperatormig (712.90s)
PASS


gejain:
We need to do more changes here, as per reviews from Rohit, will be working on same.

PR #1485, opened 2023-08-23
remove test results mount from stork integration test container

Rohit-PX:
**What type of PR is this?**
integration-test

* OCP tests are running when running stork-integration test pod as the test tries to create the `testresults` directory but
does not have adequate permissions
* The solution is to run the test pod with privileged security context
* But since we don't use the testresults directory any more (jenkins is now configured to not store that artifact) we can remove this from the stork-integration test pod spec and still make it work for OCP
* This way we can also avoid having to run the stork-integration test pod with privileged mode 

no

no

no


diptiranjanpx (commented):
Can we mount the /testresults directory in the spec so that the directory is already available and no need to create it ? it can be an alternative. If at all we are not using the result, then it looks fine to remove.

PR #1489, opened 2023-08-29
PWX-33267: use gcc 10.5.0 container to build px_statfs.so

pureneelesh:
**What type of PR is this?**

bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

We should be able to load px_statfs.so on RHEL8.5 (OCP 4.12) which uses glibc 2.28. Therefore, we need to pin down the version of gcc used to build px_statfs.so.

Switched the px_statfs.so build container to gcc:10.5.0 from golang:1.19.10 which uses gcc 12.

No

```release-note
Issue: 
When using Stork versions 23.7.0 and 23.7.1, KubeVirt pods go into CrashLoopBackoff state. 
The following errors appear in the pod logs:

/usr/bin/virt-launcher-monitor: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /etc/px_statfs.so)
/usr/bin/virt-launcher-monitor: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /etc/px_statfs.so)

User Impact:
KubeVirt VMs are unusable.

Resolution:
The problem is fixed in Stork version 23.7.2. However, there may be an additional post-upgrade step required depending on which version you are upgrading from.
- When upgrading from any release older than 23.7.0, no additional action is required after the upgrade.
- When upgrading from 23.7.0 or 23.7.1, the following additional step is needed.

Perform the following steps for each virt-launcher pod that is in a CrashLoopBackOff state:
1. Stop KubeVirt VM.
2. Delete a configMap called `px-statfs` in the same namespace as the virt-launcher pod.
    > kubectl -n <vm-namespace> delete configmap px-statfs
3. Start KubVirt VM.
```

23.7.2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,12 @@ STORK_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORK_IMAGE):$(DOCKER_HUB_STORK_TAG)
 CMD_EXECUTOR_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_CMD_EXECUTOR_IMAGE):$(DOCKER_HUB_CMD_EXECUTOR_TAG)
 STORK_TEST_IMG=$(DOCKER_HUB_REPO)/$(DOCKER_HUB_STORK_TEST_IMAGE):$(DOCKER_HUB_STORK_TEST_TAG)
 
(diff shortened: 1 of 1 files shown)


pp511 (approved):
lgtm.
Please make sure to add release notes too.

PR #1493, opened 2023-08-31
 PWX-33267: update px-statfs configmap if needed

pureneelesh:
**What type of PR is this?**
>bug
>feature
improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

When a new virt-launcher pod is starting, update the existing px-statfs configMap, if needed.

no

no

no

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -179,7 +179,9 @@ const (
 	pureBlockParam   = "pure_block"
 
 	statfsSOName               = "px_statfs.so"
+	stasfsSOSumName            = "px_statfs.so.sha256"
 	statfsSOPathInStork        = "/" + statfsSOName
+	statfsSOSumPathInStork     = "/" + stasfsSOSumName
 	statfsSOPathInVirtLauncher = "/etc/" + statfsSOName
 	statfsConfigMapName        = "px-statfs"
 	statfsVolName              = "px-statfs"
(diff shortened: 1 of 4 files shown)


test/integration_test/px_statfs_webhook_test.go
```diff
@@ -153,3 +167,106 @@ func runCommandInSv4TestContainer(t *testing.T, pod *corev1.Pod, cmd []string) s
...
+		updateFn(cm)
+		cm, err = core.Instance().UpdateConfigMap(cm)
+		require.NoError(t, err)
+
+		err = core.Instance().DeletePod(pod.Name, pod.Namespace, false)
+		require.NoError(t, err)
+
+		require.Eventuallyf(t, func() bool {
+			return configMapMatches(pod.Namespace, origData, origBinaryData)
+		}, 3*time.Minute, 10*time.Second, "ConfigMap %s/%s did not get updated", pod.Namespace, statfsConfigMapName)
```
pp511, line 240: Help me understand this. From what I understand, origData and origBinaryData is referring to older data in the cm.  updateFn should update it and we should have updatedData and updatedBinaryData against which we should compare in  configMapMatches. What am I missing?

pp511 (commented):
Implementation looks good. 
One comment and one clarification.

drivers/volume/portworx/portworx.go
```diff
@@ -4314,17 +4342,22 @@ func (p *portworx) createStatfsConfigMap(cmNamespace string) error {
...
-	} else if err == nil {
-		logrus.Infof("Created configmap %v/%v", cmNamespace, statfsConfigMapName)
+func (p *portworx) statfsConfigMapNeedsUpdate(existing *v1.ConfigMap, statfsSOContents []byte) bool {
+	existingBytes := existing.BinaryData[statfsSOName]
+	if len(existingBytes) != len(statfsSOContents) {
+		return true
 	}
-	return nil
+	for i := range existingBytes {
+		if existingBytes[i] != statfsSOContents[i] {
```
adityadani, line 4356: Should we calculate an md5sum and compare that instead of the byte content? The downside is we will end up updating the config map the first time for everyone who upgrades from a version which does not have an md5sum.

test/integration_test/px_statfs_webhook_test.go
```diff
@@ -54,9 +61,16 @@ func TestExtenderWebhookStatfs(t *testing.T) {
 		require.NoError(t, err, "Error waiting for app to get to running state")
 	}
 
+	foundVirtLauncher := false
 	for _, appCtx := range ctxs {
 		validateStatfs(t, appCtx)
+
+		if !foundVirtLauncher && isVirtLauncherContext(appCtx) {
```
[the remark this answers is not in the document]
pureneelesh: > !foundVirtLauncher is not required in the if condition

I am using that condition to run the configmap test only for one of the contexts since there is no need to repeat it for the multiple virt-launcher contexts.

test/integration_test/px_statfs_webhook_test.go
pureneelesh: Adding the following comment to that function to clarify this. Let me know if this makes sense:
```
// This test verifies that the webhook updates the px-statfs configMap, if its contents do not match
// the expected values.
// Test steps:
//
// - Start with a "good" px-statfs configmap that the webhook has already created.
// - Make some dummy changes to the px-statfs configmap
// - Trigger the webhook by deleting one of the virt-launcher-sim pod (new pod should start)
// - Wait until the webhook has restored the original configMap
```

test/integration_test/px_statfs_webhook_test.go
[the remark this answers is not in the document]
pp511: Got it.

test/integration_test/px_statfs_webhook_test.go
pp511: Makes sense.

drivers/volume/portworx/portworx.go
pp511: I think it's a good idea. 
We can also use hashstructure library as used here https://github.com/libopenstorage/stork/blob/b6e03704b305d2c601dd2fb7d420398ba9b98103/pkg/migration/controllers/migration.go#L1305C2-L1305C2 
Downside will only come in picture if we store the hash of existing CM and not when do both conversions on the fly. Even then it should be ok.

pureneelesh: Added sha256. In hindsight, this is a bit of an overkill since the file size is only 20k. But might help in future if the file size grows.

adityadani (approved):

adityadani (approved):

PR #1505, opened 2023-09-15
Made sse in backuplocation CR to be public by changing it to SSE

siva-portworx:
**What type of PR is this?**
>bug

```
Made sse in backuplocation CR to be public by changing it to SSE
```

no

NA ( will add once the full support is added)

23.8.0


prashanthpx (approved):

siva-portworx:
Code coverage is failing, So force merging it.

PR #1507, opened 2023-09-17
PWX-33787: Ignore backuplocation not found error during clusterpair deletion.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
bug

Deleting a namespace can make backuplocation to delete first and it should not block the clusterpair deletion from px.

**Test**
➜  stork git:(23.7.3) ✗ kubectl delete ns test1
namespace "test1" deleted
➜  stork git:(23.7.3) ✗ kubectl get clusterpair -A
No resources found
➜  stork git:(23.7.3) ✗ kubectl get backuplocations -ntest1
No resources found in test1 namespace.
➜  stork git:(23.7.3) ✗ kubectl get secret -ntest1
No resources found in test1 namespace.

[root@dipti5-k8s-pxc-8d21-0 ~]# docker run --rm -e ETCDCTL_API=3 sens/etcdctl --endpoints http://*******:9019 get --prefix=true pwx/dipti5/cluster/pair
[root@dipti5-k8s-pxc-8d21-0 ~]#

Complete result has been updated in PWX-33787,

### DIFF
--- a/pkg/migration/controllers/clusterpair.go
+++ b/pkg/migration/controllers/clusterpair.go
@@ -233,13 +233,11 @@ func (c *ClusterPairController) cleanup(clusterPair *stork_api.ClusterPair) erro
 	// Delete the backuplocation and secret associated with clusterpair as part of the delete
 	if backuplocationName, ok := clusterPair.Spec.Options["backuplocation"]; ok {
 		bl, err := storkops.Instance().GetBackupLocation(backuplocationName, clusterPair.Namespace)
-		if err != nil {
-			if !errors.IsNotFound(err) {
-				logrus.Errorf("fetching backuplocation %s in ns %s failed: %v", backuplocationName, clusterPair.Namespace, err)
-			}
-			return nil
+		if err != nil && !errors.IsNotFound(err) {
+			logrus.Errorf("fetching backuplocation %s in ns %s failed: %v", backuplocationName, clusterPair.Namespace, err)
+			return err
 		}
-		if bl.Annotations[storkCreatedAnnotation] == "true" {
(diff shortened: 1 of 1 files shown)


pp511 (commented):
Along the same line of thought..Is it not possible that Secret still exists and doesn't get cleaned up?

diptiranjanpx:
> Along the same line of thought..Is it not possible that Secret still exists and doesn't get cleaned up?

Earlier issue was , there was a return statement if backuplocation is not present which was causing clusterpair not to be deleted. The fix is around that to ignore that error and continue to delete the clusterpair from px.

Coming back to if secret can still lie around, unless manually the backuplocation is deleted it should not happen.
1. If backuplocation gets manually deleted, we will not find any ref to the secret, so it will still exists as part of clusterpair deletion flow.
2. If ns gets deleted, secret may not get deleted as part of clusterpair deletion flow but as part of ns deletion, it will get deleted anyway.
3. In a regular clusterpair deletion workflow, backuplocation deletion is after secret deletion so as long as all exists, deletion happens in order
      a. secret
      b. backuplocation
      c. clusterpair entry in px.

pp511 (approved):
If backuplocation gets manually deleted, we will not find any ref to the secret, so it will still exists as part of clusterpair deletion flow.
> If I understand correctly this is still an issue we can address. Can we file a ticket to track this? We will discuss about the actual implementation when we get to it.

diptiranjanpx:
> If backuplocation gets manually deleted, we will not find any ref to the secret, so it will still exists as part of clusterpair deletion flow.
> 
> > If I understand correctly this is still an issue we can address. Can we file a ticket to track this? We will discuss about the actual implementation when we get to it.

done

PR #1508, opened 2023-09-18
update stork integration test container to Go 1.17.0

Rohit-PX:
integration-test

https://portworx.atlassian.net/browse/PTX-20175

No

No

No

### DIFF
--- a/test/integration_test/Dockerfile
+++ b/test/integration_test/Dockerfile
@@ -1,8 +1,8 @@
-FROM golang:1.16.1
+FROM golang:1.17.0
 
 # Install dependancies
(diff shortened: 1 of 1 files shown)


pp511:
We should move the golang 1.19.10 which is the same container Stork image is using.

Also nother side note. I just noticed that we are using older versions of google sdk and aws authenticator in the test binaries. I think we should migrate them all to the latest versions and keep them in sync with stork container. We can do it as a separate task.

pp511 (approved):
Changes look ok. 
I took the liberty of updating PR description to show 1.17.0 instead of 1.19.10.

PR #1523, opened 2023-10-07
PB-3046 :: Restrict portworx ns backup

vikasit12:
**What type of PR is this?**

>bug

```
- Restrict backup of namespace where portworx is installed in case of all namespaces i.e. *
- Restrict backup of namespace where portworx is installed in case of label-selector
- Allow backup of namespace where portworx is installed in case API is specifically passing it i.e. namsespace=kube-system in API call
```

no

```release-note
Issue: #PB-3046
User Impact: User will not be able to backup namespace where portworx is installed if tried with label-selector or all namespace
Resolution: If user wants to backup namespace where portworx is installed the one needs to pass it specifically as namsespace=kube-system in API calls
```

yes, 23.9

**Testing Screenshot**
image
image
image

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -221,9 +221,10 @@ func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.
 	if err != nil {
 		return fmt.Errorf("error updating with all namespaces for wildcard: %v", err)
 	}
+	pxNs, _ := utils.GetPortworxNamespace()
 	namespacesToBackup := make([]string, 0)
 	for _, ns := range namespaces.Items {
-		if ns.Name != "kube-system" {
+		if ns.Name != "kube-system" && ns.Name != pxNs {
 			namespacesToBackup = append(namespacesToBackup, ns.Name)
 		}
 	}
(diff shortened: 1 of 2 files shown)


siva-portworx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -221,9 +221,10 @@ func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.
 	if err != nil {
 		return fmt.Errorf("error updating with all namespaces for wildcard: %v", err)
 	}
+	pxNs, _ := utils.GetPortworxNamespace()
```
px-kesavan, line 224: We already evaluated the pxNs in the handle method , can we pass the pxNs to this method or persist in ApplicationBackupController?. So no need for reevaluating the pxNs , cause listing all services in a cluster is an expensive operation

siva-portworx: Both label and all namespace can not be set together?

vikasit12: No both cannot be set together

vikasit12: > We already evaluated the pxNs in the handle method , can we pass the pxNs to this method or persist in ApplicationBackupController?. So no need for reevaluating the pxNs , cause listing all services in a cluster is an expensive operation

Yes this would have been the ideal solution but there are possibilities that a customer a person is having stork but not portworx initially and later he might have it. In that scenario is it wise to store it?

vikasit12: > Both label and all namespace can not be set together?

No both cannot be set together.

vikasit12: > @vikasit12 I think, @px-kesavan is referring to avoid the call, for each reconcile cycle. In the case namespace label, we are calling functions to fetch the namespace everytime in the handle. That means it will be called every reconciler cyle. You can avoid it by checking if namespace label is present and namespace list is empty, then fetch the namespace.

Makes sense and changes have been made accordingly

px-kesavan (approved):

PR #1530, opened 2023-10-12
PB-4598: Fix for Stork pods crash issue if the RT CR using the a transform of type slice and operation add

sgajawada-px:
- Use SetNestedStringSlice instead of SetNestedField in case of SliceResourceType to avoid stork pod crash due to a panic

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

To fix the stork pod crash issue when ever a rt cr created with slice type with add operation

no

no

```release-note
Issue:
User Impact:
Resolution

```

 Yes because it is a bug in the latest release stork release.

issue: https://portworx.atlassian.net/browse/PB-4598
unit test result:
Screenshot from 2023-10-12 18-54-59

### DIFF
--- a/pkg/resourcecollector/resourcetransformation.go
+++ b/pkg/resourcecollector/resourcetransformation.go
@@ -68,7 +68,8 @@ func TransformResources(
 							return err
 						}
 					} else if path.Type == stork_api.SliceResourceType {
-						err := unstructured.SetNestedField(content, value, strings.Split(path.Path, ".")...)
+						updateSlice := value.([]string)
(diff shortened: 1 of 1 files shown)


lalat-das:
@sgajawada-px We need to cherry-pick this change to 23.9, Since it can crash stork pod repeatedly because of resourcetransformation CR and keep crashing even after pod bounce, till the CR is deleted. @adityadani releasing this bug as part of 23.9 should suffice right ?

lalat-das (approved):
lgtm

diptiranjanpx (approved):

PR #1536, opened 2023-10-16
PB-4069 :: Restrict initial-namespaces backup

vikasit12:
- Restrict backup of initial-namespaces in case of all namespaces i.e. *
- Restrict backup of initial-namespaces in case of label-selector
- Allow backup of initial-namespaces in case API is specifically passing it i.e. namsespace=kube-public in API call initial-namespaces are https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#initial-namespaces

**What type of PR is this?**
>improvement

```
- Restrict backup of initial-namespaces in case of all namespaces i.e. *
- Restrict backup of initial-namespaces in case of label-selector
- Allow backup of initial-namespaces in case API is specifically passing it i.e. namsespace=kube-public in API call 

initial-namespaces are https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#initial-namespaces
```

no

```release-note
Issue: #PB-4609
User Impact: User will not be able to backup initial namespaces if tried with label-selector or all namespace
Resolution: If user wants to backup initial namespace then one needs to pass it specifically as namsespace=kube-public in API calls

```

23.9

**Testing Screenshot ::**
image
image

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -222,9 +222,16 @@ func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.
 		return fmt.Errorf("error updating with all namespaces for wildcard: %v", err)
 	}
 	pxNs, _ := utils.GetPortworxNamespace()
+	// Create a map to store the namespaces to be ignored for fast lookup
+	ignoreNamespaces := map[string]bool{
+		pxNs:              true,
+		"kube-system":     true,
+		"kube-public":     true,
+		"kube-node-lease": true,
+	}
 	namespacesToBackup := make([]string, 0)
 	for _, ns := range namespaces.Items {
-		if ns.Name != "kube-system" && ns.Name != pxNs {
+		if _, found := ignoreNamespaces[ns.Name]; !found {
 			namespacesToBackup = append(namespacesToBackup, ns.Name)
 		}
 	}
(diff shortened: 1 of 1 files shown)


siva-portworx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -222,9 +222,16 @@ func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.
 		return fmt.Errorf("error updating with all namespaces for wildcard: %v", err)
 	}
 	pxNs, _ := utils.GetPortworxNamespace()
+	// Create a map to store the namespaces to be ignored for fast lookup
+	ignoreNamespaces := map[string]bool{
```
prashanthpx, line 226: Can we this move to pkg/utils rather than defining in both the places?

prashanthpx (approved):
LGTM after dressing the review comment

px-kesavan (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -222,9 +222,16 @@ func (a *ApplicationBackupController) updateWithAllNamespaces(backup *stork_api.
...
+		pxNs:              true,
+		"kube-system":     true,
+		"kube-public":     true,
+		"kube-node-lease": true,
```
diptiranjanpx, line 230: Are there any other openshift specific namespaces those need to be ignored?

vikasit12: Right now not probably in future releases. Currently we are ignoring only the initial namespaces and portworx namespace.

PR #1539, opened 2023-10-18
PWX-34395: Existing pod attachment info not required for scheduling d…

pp511:
…uring live migration

**What type of PR is this?**
>improvement

* Existing Pod node will never be a part of input set of nodes due to Pod anti affinities set by kubevirt. In such a case isPodUsingLocallyAttachedVolume is not necessarily required. It does simplify the scheduling logic. 
```
kind: Pod
...
  labels:
    kubevirt.io/created-by: 30716366-8726-4024-9d8d-5500ba8ceb87
...
spec:
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            kubevirt.io/created-by: 30716366-8726-4024-9d8d-5500ba8ceb87
        topologyKey: kubernetes.io/hostname

```
* UTs are updated to give higher score to the nodes which have volume attached irrespective of the pod placement on them. In an actual scenario, these nodes will not be a part of the scheduling node request input.

yes

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -1006,7 +1006,6 @@ func (e *Extender) updateVirtLauncherPodPrioritizeScores(
 		}
 	} else {
 		driverNodes = volume.RemoveDuplicateOfflineNodes(driverNodes)
-
 		for _, dnode := range driverNodes {
 			for _, knode := range args.Nodes.Items {
 				// Initialize with score to with score for remote node
@@ -1018,11 +1017,7 @@ func (e *Extender) updateVirtLauncherPodPrioritizeScores(
 						if dataNode == dnode.StorageID {
 							// Current node has the volume replica
 							score = replicaNodeScore
-							nodeHasAttachedVolume := e.nodeHasAttachedVolume(dnode, volInfo)
-							if nodeHasAttachedVolume {
-								// There is no LiveMigration in progress
-								// When a new pod gets scheduling request
-								// assign a higher score for local volume attachment
+							if e.nodeHasAttachedVolume(dnode, volInfo) {
 								score = int64(2 * nodePriorityScore)
 							}
 						}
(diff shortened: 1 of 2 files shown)


pkg/extender/extender.go
```diff
@@ -1053,21 +1019,11 @@ func (e *Extender) updateVirtLauncherPodPrioritizeScores(
...
-								if nodeHasAttachedVolume {
-									// There is no LiveMigration in progress
-									// When a new pod gets scheduling request
-									// assign a higher score for local volume attachment
-									score = int64(2 * nodePriorityScore)
-								}
+							if nodeHasAttachedVolume {
+								// There is no LiveMigration in progress
+								// When a new pod gets scheduling request
+								// assign a higher score for local volume attachment
```
pureneelesh, line 1025: nit: this comment does not make sense to me. If live migration is NOT in progress, there will not be any volume attachment yet.

pkg/extender/extender.go
```diff
@@ -1053,21 +1019,11 @@ func (e *Extender) updateVirtLauncherPodPrioritizeScores(
...
-									score = int64(2 * nodePriorityScore)
-								}
-							} else {
-								if nodeHasAttachedVolume {
-									// There is no LiveMigration in progress
-									// When a new pod gets scheduling request
-									// assign a higher score for local volume attachment
-									score = int64(2 * nodePriorityScore)
-								}
+							if nodeHasAttachedVolume {
```
pureneelesh, line 1022: nit: no need for the variable anymore. We can do `if e.nodeHasAttachedVolume(dnode, volInfo)` since it is called only once.

pureneelesh (approved):

pkg/extender/extender.go
pp511: This is a comment from the previous commit where it made sense based on the particular conditional statement it was inside. Now the corresponding change is gone and it's pointless here. Thanks for pointing. Removing this.

pkg/extender/extender.go
pp511: Ack.

PR #1540, opened 2023-10-19
Changed backup tests sequence

gejain:
**What type of PR is this?**
>integration-test

Changed backup tests sequence

Passed Job - https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.8-dev/job/backup-k8s-1-25-0/1129/consoleText


Rohit-PX (approved):

pp511 (approved):
Looks ok. But wanted to understand any particular reason this was done?

gejain:
> Looks ok. But wanted to understand any particular reason this was done?

Some of the below cases are causing context change and they are running in sequence. So keeping this with simple backup restore cases will keep things as it is. But, I will open another ticket to analyse the specific case which is causing failure of backup restore cases if kept on last in sequence. As of now this looks an immidiate fix.

PR #1541, opened 2023-10-20
PWX-32199: Improvements to storkctl cli create migration schedule

olavangad-px:
**What type of PR is this?**
>improvement
>unit-test

[PWX-32199](https://portworx.atlassian.net/browse/PWX-32199)
We are updating the storkctl create migrationschedule to add support for --interval , --auto-suspend and --annotations flags.

Yes, storkctl create migrationschedules cli command is updated with new flags

```[root@ip-10-13-195-219 ~]# storkctl create migrationschedules -h
Create a migration schedule

Usage:
  storkctl create migrationschedules [flags]

Aliases:
  migrationschedules, migrationschedule

Flags:
      --annotations stringToString    Add required annotations to the resource in comma-separated key value pairs. key1=value1,key2=value2,...  (default [])
  -c, --cluster-pair string           Specify the name of the ClusterPair in the same namespace to be used for the migration
      --disable-auto-suspend          Prevent automatic suspension of DR migration schedules on the source cluster in case of a disaster
      --exclude-resources             If present, Kubernetes resources will not be migrated
      --exclude-volumes               If present, the underlying Portworx volumes will not be migrated. This is the only allowed and default behaviour in sync-dr use cases.
  -h, --help                          help for migrationschedules
  -i, --interval int                  Specify the time interval, in minutes, at which Stork should trigger migrations (default 30)
      --namespaces strings            Specify the comma-separated list of namespaces to be included in the migration
      --post-exec-rule string         Specify the name of the rule to be executed after every migration is triggered
      --pre-exec-rule string          Specify the name of the rule to be executed before every migration is triggered
  -s, --schedule-policy-name string   Name of the schedule policy to use. If you want to create a new interval policy, use the --interval flag instead (default "default-migration-policy")
  -a, --start-applications            If present, the applications will be scaled up on the target cluster after a successful migration
      --suspend                       Flag to denote whether schedule should be suspended on creation
```

Not sure, will update

Yes, 23.10.0

Testing details:
Have added unit tests covering all the branches.
Have manually created migration schedules using the command and verified migrations are scheduled as expected in my local cluster.
Will be taking up integration tests for these changes as the next task.

[PWX-32199]: https://portworx.atlassian.net/browse/PWX-32199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/storkctl/migrationschedule.go
+++ b/pkg/storkctl/migrationschedule.go
@@ -107,6 +107,8 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 					return
 				}
 			} else {
+				//we reach here if user has not set the interval flag value.
+				//So we take either the user input for schedulePolicyName or the default schedulePolicyName
 				logrus.Infof("schedulePolicy name is %s", schedulePolicyName)
 				_, err := storkops.Instance().GetSchedulePolicy(schedulePolicyName)
 				if err != nil {
@@ -152,12 +154,10 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 	createMigrationScheduleCommand.Flags().BoolVarP(&startApplications, "startApplications", "a", false, "Start applications on the destination cluster after migration")
 	createMigrationScheduleCommand.Flags().StringVarP(&preExecRule, "preExecRule", "", "", "Rule to run before executing migration")
 	createMigrationScheduleCommand.Flags().StringVarP(&postExecRule, "postExecRule", "", "", "Rule to run after executing migration")
-	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. "+
-		"If you want to create a new interval policy, use interval flag instead.")
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. If you want to create a new interval policy, use the interval flag instead")
 	createMigrationScheduleCommand.Flags().BoolVar(&suspend, "suspend", false, "Flag to denote whether schedule should be suspended on creation")
-	createMigrationScheduleCommand.Flags().BoolVar(&autoSuspend, "auto-suspend", true, "If set to true, in the event of a disaster, Stork will automatically suspend the DR migration schedules on your source cluster, "+
-		"allowing you to migrate your application to an active Kubernetes cluster. ")
-	createMigrationScheduleCommand.Flags().IntVarP(&intervalMinutes, "interval", "i", 30, "Specify the periodic interval, in minutes, after which Stork should trigger the operation.")
+	createMigrationScheduleCommand.Flags().BoolVar(&autoSuspend, "auto-suspend", true, "In case of a disaster, Stork will automatically suspend DR migration schedules on your source cluster, enabling migration to an active Kubernetes cluster")
+	createMigrationScheduleCommand.Flags().IntVarP(&intervalMinutes, "interval", "i", 30, "Specify the periodic interval, in minutes, after which Stork should trigger the migration")
 	createMigrationScheduleCommand.Flags().StringToStringVar(&annotations, "annotations", map[string]string{}, "Add required annotations to the resource in the format key1=value1,key2=value2,... ")
 	return createMigrationScheduleCommand
 }
--- a/pkg/storkctl/migrationschedule.go
+++ b/pkg/storkctl/migrationschedule.go
@@ -19,6 +19,13 @@ var migrationScheduleSubcommand = "migrationschedules"
 var migrationScheduleAliases = []string{"migrationschedule"}
 
 func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
+
+	const (
+		intervalFlag           = "interval"
+		schedulePolicyNameFlag = "schedule-policy-name"
+		excludeVolumesFlag     = "exclude-volumes"
+	)
+
 	var migrationScheduleName string
 	var clusterPair string
 	var namespaceList []string
@@ -60,17 +67,18 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 				util.CheckErr(fmt.Errorf("unable to find the cluster pair in the given namespace"))
 				return
 			}
-			isSyncDr := false
-			//clusterPairObj.Spec.Options map is empty for syncDr and non-empty for asyncDR
+			isStorageOptionsProvided := true
+			//clusterPairObj.Spec.Options map is empty for syncDr and should be non-empty for volume migration in asyncDR
 			if len(clusterPairObj.Spec.Options) == 0 {
-				isSyncDr = true
+				isStorageOptionsProvided = false
 			}
 
-			// Default value of includeVolumes for asyncDr is true and for sync DR is false
-			if isSyncDr {
-				//covers the unlikely scenario where user sets the --exclude-volumes=false value in the command for a sync-dr migration use case
-				if c.Flags().Changed("exclude-volumes") && includeVolumes {
-					util.CheckErr(fmt.Errorf("--exclude-volumes can only be set to true in case of a sync-dr usecase as there is a single stretched cluster from storage perspective"))
+			// Default value of includeVolumes for asyncDr with StorageOptions provided is true and for sync DR is false
+			if !isStorageOptionsProvided {
+				//covers the scenario where user sets the --exclude-volumes=false value in the command for a use case where clusterPair's storage-status is Not Provided
+				//this is true if the cluster pair is created for a sync-dr use-case or if clusterPair was created without specifying storage options
+				if c.Flags().Changed(excludeVolumesFlag) && includeVolumes {
+					util.CheckErr(fmt.Errorf("--exclude-volumes can only be set to true if it is a sync-dr use case or storage options are not provided in the cluster pair"))
 					return
 				}
 				includeVolumes = false
@@ -82,17 +90,17 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 			}
 
 			//There are 4 possible cases for schedulePolicyName and interval flags
-			//user provides both schedulePolicyName and interval value -> we throw an error saying you can only fill one of the values
-			//user provides interval value only -> we will create a new schedule policy and use that in the migrationSchedule
-			//user provides schedulePolicyName only -> we will check if such a schedule policy exists and if yes use that schedulePolicy.
-			//user doesn't provide schedulePolicyName nor interval value -> we go ahead with the "default-migration-policy"
+			//1. user provides both schedulePolicyName and interval value -> we throw an error saying you can only fill one of the values
+			//2. user provides interval value only -> we will create a new schedule policy and use that in the migrationSchedule
+			//3. user provides schedulePolicyName only -> we will check if such a schedule policy exists and if yes use that schedulePolicy.
+			//4. user doesn't provide schedulePolicyName nor interval value -> we go ahead with the "default-migration-policy"
 
-			if c.Flags().Changed("schedule-policy-name") && c.Flags().Changed("interval") {
+			if c.Flags().Changed(schedulePolicyNameFlag) && c.Flags().Changed(intervalFlag) {
 				util.CheckErr(fmt.Errorf("must provide only one of schedule-policy-name or interval values"))
 				return
 			}
 
-			if c.Flags().Changed("interval") {
+			if c.Flags().Changed(intervalFlag) {
 				var policyItem storkv1.SchedulePolicyItem
 				var intervalPolicy storkv1.IntervalPolicy
 				intervalPolicy.IntervalMinutes = intervalMinutes
@@ -115,7 +123,7 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 					return
 				}
 			} else {
-				// case 3 and 4 are covered in this branch
+				// case #3 and #4 are covered in this branch
 				_, err := storkops.Instance().GetSchedulePolicy(schedulePolicyName)
 				if err != nil {
 					util.CheckErr(fmt.Errorf("unable to get schedulepolicy %v: %v", schedulePolicyName, err))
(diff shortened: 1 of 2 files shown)


pkg/storkctl/migrationschedule.go
```diff
@@ -93,9 +152,13 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. "+
+		"If you want to create a new interval policy, use interval flag instead.")
 	createMigrationScheduleCommand.Flags().BoolVar(&suspend, "suspend", false, "Flag to denote whether schedule should be suspended on creation")
-
+	createMigrationScheduleCommand.Flags().BoolVar(&autoSuspend, "auto-suspend", true, "If set to true, in the event of a disaster, Stork will automatically suspend the DR migration schedules on your source cluster, "+
+		"allowing you to migrate your application to an active Kubernetes cluster. ")
+	createMigrationScheduleCommand.Flags().IntVarP(&intervalMinutes, "interval", "i", 30, "Specify the periodic interval, in minutes, after which Stork should trigger the operation.")
+	createMigrationScheduleCommand.Flags().StringToStringVar(&annotations, "annotations", map[string]string{}, "Add required annotations to the resource in the format key1=value1,key2=value2,... ")
```
olavangad-px, line 161: If user has to use annotations flag then they will have to input something like this ->
`--annotations openstorage.io/auth-secret-namespace=val1,openstorage.io/auth-secret-name=val2` which seems a bit inconvenient
If in the annotations need to take only these 2 above values , then we could instead add 2 separate flags for them
Something like `--authSecretNamespace val1 --authSecretName val2`

@pp511 @adityadani

pp511: We should  comma separated annotations to make it future proof.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +48,77 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+				}
+				policyItem.Interval = &intervalPolicy
+				//name of the schedule policy created will be same as the migration schedule name provided as an input.
+				schedulePolicyName = migrationScheduleName
+				var schedulePolicy storkv1.SchedulePolicy
+				schedulePolicy.Name = schedulePolicyName
+				schedulePolicy.Policy = policyItem
+				_, err = storkops.Instance().CreateSchedulePolicy(&schedulePolicy)
+				if err != nil {
+					util.CheckErr(fmt.Errorf("could not create a schedule policy with specified interval. %v", err))
```
pp511, line 106: nit..Could

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +48,77 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+					return
+				}
+			} else {
+				//we reach here if user has not set the interval flag value.
+				//So we take either the user input for schedulePolicyName or the default schedulePolicyName
+				logrus.Infof("schedulePolicy name is %s", schedulePolicyName)
```
pp511, line 112: I didn't understand the comment here.. "or the default schedulePolicyName". Can you explain?

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +48,77 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+				_, err := storkops.Instance().GetSchedulePolicy(schedulePolicyName)
+				if err != nil {
+					util.CheckErr(fmt.Errorf("unable to get schedulepolicy %v: %v", schedulePolicyName, err))
```
pp511, line 115: nit..Unable

pp511 (commented):
Some concerns for you to address.
Will review the test migration_test.go in second pass after you have addressed current concerns.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +48,77 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
 			if len(namespaceList) == 0 {
 				util.CheckErr(fmt.Errorf("need to provide atleast one namespace to migrate"))
 				return
 			}
-			if len(schedulePolicyName) == 0 {
-				util.CheckErr(fmt.Errorf("need to provide schedulePolicyName"))
+
+			//Allow the user to only add one of the flags schedulePolicyName or interval
+			if c.Flags().Changed("schedulePolicyName") && c.Flags().Changed("interval") {
+				util.CheckErr(fmt.Errorf("must provide only one of schedulePolicyName or interval values"))
```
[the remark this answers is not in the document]
olavangad-px: Addressing reason all the error messages not starting with capitalized letters.
In one of the previous PRs I have already tried making it capitalized however it leads to travis build failure due to the staticcheck.
```
pkg/storkctl/schedulepolicy.go:119:20: error strings should not be capitalized (ST1005)
Makefile:85: recipe for target 'staticcheck' failed
make: *** [staticcheck] Error 1
```

pkg/storkctl/migrationschedule.go
olavangad-px: There are 4 cases in all
1. user doesn't provide schedulePolicyName nor interval value -> we should go ahead with the "default-migration-policy"
2. user provides both schedulePolicyName and interval value -> we throw an error saying you can only fill one of the values
3. user provides interval value only -> we will create a new schedule policy and use that in the migrationschedule
4. user provides schedulePolicyName only -> we will check if such a schedule policy exists and if yes use that schedulePolicy.

So in my code case 1 and 4 are being covered in this branch

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +48,77 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+			//clusterPairObj.Spec.Options map is empty for syncDr and non-empty for asyncDR
+			if len(clusterPairObj.Spec.Options) == 0 {
+				isSyncDr = true
+			}
+
+			// Default value of includeVolumes for syncDr is false and for async DR is true
+			if isSyncDr {
+				if c.Flags().Changed("includeVolumes") && includeVolumes {
+					util.CheckErr(fmt.Errorf("IncludeVolumes can only be set to false because the volumes are already present " +
+						"on the destination cluster as we deal with a single Portworx cluster in sync-dr usecases"))
```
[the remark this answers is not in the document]
olavangad-px: will address

pkg/storkctl/migrationschedule.go
```diff
@@ -93,9 +154,11 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
 	createMigrationScheduleCommand.Flags().StringVarP(&postExecRule, "postExecRule", "", "", "Rule to run after executing migration")
-	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use")
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. If you want to create a new interval policy, use the interval flag instead")
 	createMigrationScheduleCommand.Flags().BoolVar(&suspend, "suspend", false, "Flag to denote whether schedule should be suspended on creation")
-
+	createMigrationScheduleCommand.Flags().BoolVar(&autoSuspend, "autoSuspend", true, "In case of a disaster, Stork will automatically suspend DR migration schedules on your source cluster, enabling migration to an active Kubernetes cluster")
+	createMigrationScheduleCommand.Flags().IntVarP(&intervalMinutes, "interval", "i", 30, "Specify the periodic interval, in minutes, after which Stork should trigger the migration")
```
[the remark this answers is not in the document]
olavangad-px: Makes sense. Will address

pkg/storkctl/migrationschedule.go
```diff
@@ -93,9 +154,11 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
 	createMigrationScheduleCommand.Flags().StringVarP(&preExecRule, "preExecRule", "", "", "Rule to run before executing migration")
 	createMigrationScheduleCommand.Flags().StringVarP(&postExecRule, "postExecRule", "", "", "Rule to run after executing migration")
-	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use")
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. If you want to create a new interval policy, use the interval flag instead")
 	createMigrationScheduleCommand.Flags().BoolVar(&suspend, "suspend", false, "Flag to denote whether schedule should be suspended on creation")
-
+	createMigrationScheduleCommand.Flags().BoolVar(&autoSuspend, "autoSuspend", true, "In case of a disaster, Stork will automatically suspend DR migration schedules on your source cluster, enabling migration to an active Kubernetes cluster")
```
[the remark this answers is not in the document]
olavangad-px: will address

pkg/storkctl/migrationschedule.go
```diff
@@ -93,9 +154,11 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 	createMigrationScheduleCommand.Flags().BoolVarP(&startApplications, "startApplications", "a", false, "Start applications on the destination cluster after migration")
 	createMigrationScheduleCommand.Flags().StringVarP(&preExecRule, "preExecRule", "", "", "Rule to run before executing migration")
 	createMigrationScheduleCommand.Flags().StringVarP(&postExecRule, "postExecRule", "", "", "Rule to run after executing migration")
-	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use")
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedulePolicyName", "s", "default-migration-policy", "Name of the schedule policy to use. If you want to create a new interval policy, use the interval flag instead")
```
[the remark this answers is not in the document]
olavangad-px: okay, will take a look at all the flag descriptions and modify them if needed

pkg/storkctl/migrationschedule.go
olavangad-px: Have added better comments to make this code block self-explanatory.

pkg/storkctl/migrationschedule.go
```diff
@@ -86,16 +152,18 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+	createMigrationScheduleCommand.Flags().BoolVarP(&excludeResources, "exclude-resources", "", false, "If present, Kubernetes resources will not be migrated")
+	createMigrationScheduleCommand.Flags().BoolVarP(&excludeVolumes, "exclude-volumes", "", false, "If present, the underlying Portworx volumes will not be migrated. This is the only allowed and default behaviour in sync-dr use cases.")
+	createMigrationScheduleCommand.Flags().BoolVarP(&startApplications, "start-applications", "a", false, "If present, the applications will be scaled up on the target cluster after a successful migration")
+	createMigrationScheduleCommand.Flags().StringVarP(&preExecRule, "pre-exec-rule", "", "", "Specify the name of the rule to be executed before every migration is triggered")
+	createMigrationScheduleCommand.Flags().StringVarP(&postExecRule, "post-exec-rule", "", "", "Specify the name of the rule to be executed after every migration is triggered")
+	createMigrationScheduleCommand.Flags().StringVarP(&schedulePolicyName, "schedule-policy-name", "s", "default-migration-policy", "Name of the schedule policy to use. If you want to create a new interval policy, use the interval flag instead")
```
pp511, line 162: nit.. --interval flag

pp511 (commented):
LGTM. Some nits that you can address/respond to. Will approve once you have pushed the revision if you decide to act of those.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +52,78 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+			}
+			isSyncDr := false
+			//clusterPairObj.Spec.Options map is empty for syncDr and non-empty for asyncDR
+			if len(clusterPairObj.Spec.Options) == 0 {
+				isSyncDr = true
+			}
+
+			// Default value of includeVolumes for syncDr is false and for async DR is true
+			if isSyncDr {
+				if c.Flags().Changed("exclude-volumes") && includeVolumes {
```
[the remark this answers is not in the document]
olavangad-px: c.Flags().Changed("exclude-volumes") tells if user has explicitly set the value of the flag.
Just to cover all our bases, this condition covers the unlikely scenario where a user sets the --exclude-volumes=false value in the command for a sync-dr migration use case.

If the above scenario doesn't occur in the next line, we set the default value of includeVolumes to false when using a syncDr clusterpair.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +52,78 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+				_, err = storkops.Instance().CreateSchedulePolicy(&schedulePolicy)
+				if err != nil {
+					util.CheckErr(fmt.Errorf("could not create a schedule policy with specified interval. %v", err))
+					return
+				}
+			} else {
+				// case 3 and 4 are covered in this branch
+				_, err := storkops.Instance().GetSchedulePolicy(schedulePolicyName)
+				if err != nil {
+					util.CheckErr(fmt.Errorf("unable to get schedulepolicy %v: %v", schedulePolicyName, err))
```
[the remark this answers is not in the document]
olavangad-px: Makes sense. Will update the error messages accordingly

pkg/storkctl/migrationschedule.go
olavangad-px: will address

pkg/storkctl/migrationschedule_test.go
```diff
@@ -254,7 +254,7 @@ func TestCreateMigrationScheduleSyncDrExcludeVolumesFalse(t *testing.T) {
 	createClusterPair(t, clusterPair, "namespace1", "sync-dr")
 	cmdArgs := []string{"create", "migrationschedules", "-i", "15", "-c", clusterPair,
 		"--namespaces", namespace, "--annotations", "key1=value1", name, "-n", namespace, "--exclude-volumes=" + strconv.FormatBool(false)}
-	expected := "error: the flag exclude-volumes can only be set to true in case of a sync-dr usecase as there is a single stretched cluster from storage perspective"
+	expected := "error: the --exclude-volumes flags can only be set to true in case of a sync-dr usecase as there is a single stretched cluster from storage perspective"
```
pp511, line 257: nit..flag

pp511: Actually. We can just write..
error: --exclude-volumes can only be set
Just a suggestion if you like.

pkg/storkctl/migrationschedule.go
```diff
@@ -66,10 +66,11 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 				isSyncDr = true
 			}
 
-			// Default value of includeVolumes for syncDr is false and for async DR is true
+			// Default value of includeVolumes for asyncDr is true and for sync DR is false
 			if isSyncDr {
+				//covers the unlikely scenario where user sets the --exclude-volumes=false value in the command for a sync-dr migration use case
 				if c.Flags().Changed("exclude-volumes") && includeVolumes {
-					util.CheckErr(fmt.Errorf("the flag exclude-volumes can only be set to true in case of a sync-dr usecase as there is a single stretched cluster from storage perspective"))
+					util.CheckErr(fmt.Errorf("the --exclude-volumes flag can only be set to true in case of a sync-dr usecase as there is a single stretched cluster from storage perspective"))
```
pp511, line 73: nit..maybe just 
--exclude-volumes can only be set..

pp511 (approved):
LGTM.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +52,79 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+			//Identify whether it is a Sync-Dr or Async-Dr use-case from the ClusterPair spec
+			migrationScheduleNs := cmdFactory.GetNamespace()
+			clusterPairObj, err := storkops.Instance().GetClusterPair(clusterPair, migrationScheduleNs)
+			if err != nil {
+				util.CheckErr(fmt.Errorf("unable to find the cluster pair in the given namespace"))
+				return
+			}
+			isSyncDr := false
+			//clusterPairObj.Spec.Options map is empty for syncDr and non-empty for asyncDR
+			if len(clusterPairObj.Spec.Options) == 0 {
```
diptiranjanpx, line 65: We should check for the storage and scheduler status of the clusterpair before deciding if it is syncDR. It can be a miss also from the customer.

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +52,79 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
-			if len(schedulePolicyName) == 0 {
-				util.CheckErr(fmt.Errorf("need to provide schedulePolicyName"))
+
+			//There are 4 possible cases for schedulePolicyName and interval flags
+			//user provides both schedulePolicyName and interval value -> we throw an error saying you can only fill one of the values
+			//user provides interval value only -> we will create a new schedule policy and use that in the migrationSchedule
+			//user provides schedulePolicyName only -> we will check if such a schedule policy exists and if yes use that schedulePolicy.
+			//user doesn't provide schedulePolicyName nor interval value -> we go ahead with the "default-migration-policy"
+
+			if c.Flags().Changed("schedule-policy-name") && c.Flags().Changed("interval") {
```
diptiranjanpx, line 90: nit: make "interval" a constant

pkg/storkctl/migrationschedule.go
```diff
@@ -43,22 +52,79 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
...
+
 			if len(namespaceList) == 0 {
 				util.CheckErr(fmt.Errorf("need to provide atleast one namespace to migrate"))
 				return
 			}
-			if len(schedulePolicyName) == 0 {
-				util.CheckErr(fmt.Errorf("need to provide schedulePolicyName"))
+
+			//There are 4 possible cases for schedulePolicyName and interval flags
+			//user provides both schedulePolicyName and interval value -> we throw an error saying you can only fill one of the values
```
diptiranjanpx, line 85: Number these scenarios starting with 1 as these are getting referred in later comments.

pkg/storkctl/migrationschedule.go
olavangad-px: After offline discussion with @diptiranjanpx have modified the error message a bit.
includeVolumes = false is the default and only allowed value in 2 scenarios ->
1. it is a sync-dr usecase, hence no need for volume migration as it is a single stretched cluster from storage perspective.
2. while creating the clusterpair, storage options were not provided, hence storage status is NotProvided and volume migration is not possible.

diptiranjanpx (approved):

pkg/storkctl/migrationschedule.go
olavangad-px: Addressed in the revision

pkg/storkctl/migrationschedule.go
olavangad-px: Addressed in the revision

PR #1546, opened 2023-10-26
PWX-33934: remove tight loop around volumeBackup status check

lalat-das:
- Removed the frequent BackupId check while a volume backup is in progress It was in ang called in a tight for loop, hence made it every reconciler call interval based.

bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: When  portworx status check doesn't retiurn any error yet the backupId for a volume Info is nil then the backup logic will attempt to query the status frequently assuming that the volume  snapshot is still in progress. This status check was happening in a tight for loop, which was making portworx overwhelm with requests. Hence rate limited it by calling it every reconciler interval in such scenario. 

**Is a release note needed?**: No , since the actual root cause of Key not found need to be the actual release notable item.

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: 23.9 

Unit Test :
Tried a multi volume backup and restore as below 
Pasted Graphic 1

Pasted Graphic 2

Pasted Graphic

Simulated the error scenario by stubbing in the code and observed the status is being querried every 5 second which is the reconciler's  sync timer value. The log's timing can be visualized as below

Pasted Graphic 3

Pasted Graphic 4

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -660,7 +660,6 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 		namespacedName.Namespace = backup.Namespace
 		namespacedName.Name = backup.Name
 		if len(backup.Status.Volumes) != pvcCount {
-
 			for driverName, pvcs := range pvcMappings {
 				var driver volume.Driver
 				driver, err = volume.Get(driverName)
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -730,39 +729,19 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
-								if volInfo.BackupID == "" {
-									log.ApplicationBackupLog(backup).Tracef("Snapshot of volume %v is not done yet, need to loop till snapshot is done...", volInfo.PersistentVolumeClaim)
-									snapshotNotCompleted = true
-								}
-							}
-							if !snapshotNotCompleted {
-								break
+						volumeInfos, err = driver.GetBackupStatus(backup)
+						if err != nil {
+							log.ApplicationBackupLog(backup).Errorf("error getting backup status for driver %v: %v", driverName, err)
```
pp511, line 734: We do not need to log this error here as we are returning the same error and the caller is logging it.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -730,39 +729,19 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+							return fmt.Errorf("error getting backup status for driver %v: %v", driverName, err)
+						}
+						for _, volInfo := range volumeInfos {
+							if volInfo.BackupID == "" {
+								log.ApplicationBackupLog(backup).Tracef("Snapshot of volume [%v] hasn't completed yet, let's retry snapshot", volInfo.PersistentVolumeClaim)
```
pp511, line 739: I think, this should be Warnf(or atleast Infof) as we are not able to find backup status of a backup that we created. We shouldn't let this log go unnoticed.

lalat-das: sure will alter it

pp511 (approved):
lgtm

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -730,39 +729,18 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+							if volInfo.BackupID == "" {
+								log.ApplicationBackupLog(backup).Infof("Snapshot of volume [%v] hasn't completed yet, let's retry snapshot", volInfo.PersistentVolumeClaim)
+								// Some volume snapshot is not completed yet
+								// hence we will retry checking the status in the next reconciler iteration
+								// *stork_api.ApplicationBackupVolumeInfo.Status is not being checked here
+								// since backpID confirms if the snapshot is done or not already
+								return nil
 							}
 						}
 					}
```
siva-portworx, line 746: With the current logic, we are starting backup for only one batch of PVC in a reconciler cycle. Should the checking of the snapshot ID be outside the for loop, where startbackup and before the code that starts Postexec rule.
With the current, we are unnecessarily serialising the backup start, when we have more number of PVC. Please check it out.

lalat-das: I am bit confused here.. 
1. In 1st for loop we are iterating over all volumes eligible to be backed up... 
2. In 2nd for loop we are invoking startBackup()  in batches  till we  completely call this for all the PVCs  in batches
3. And in this 2nd for loop, per iteration I have added another for loop to run through all the PVCs in that batch to check if snapshot is done or not.  

- failure scenario In the 2nd for loop itself If any pvc is not backed up yet then return and let the reconciler kick in again to start from the PVC wherein we have left, by collecting all the PVCs freshly which is not backed up yet.
- success scenario - And in case that batch is completed gracefully we will go for 2nd batch again till we finish all  pvcs. If Px backend respond success for every batch of pvcs then in a single reconciler call itself we will finsh backup of all PVCs. 

I am unable to conceive without a failure how the  for-loop introduced has made each reconciler loop only takes care of one batch of volume.  This will happen only if in a particular batch volume backup hasn't completed and we need to return from reconciler. And in case we identified backup is not completed at backend it make sense to come back again in reconciler other wise we will start executing post-exec.

lalat-das: @siva-portworx  I see your point here. every time a batch is picked to be backed-up and first backup is not set  we will return from reconciler, then by that time that batch may have completed and the same thing can repeat for 2nd batch that way every reconciler  loop we will consider one one batch only in worst case scenario.

So as you suggested let me put this check out side of for loops so that the check happen after we issue backup for all the batches then a reconciler based wait makes it more faster..

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -726,44 +725,31 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+					return err
+				}
+				if driverName == volume.PortworxDriverName {
+					volumeInfos, err := driver.GetBackupStatus(backup)
+					if err != nil {
+						return fmt.Errorf("error getting backup status: %v", err)
+					}
+					for _, volInfo := range volumeInfos {
+						if volInfo.BackupID == "" {
+							log.ApplicationBackupLog(backup).Infof("Snapshot of volume [%v] hasn't completed yet, let's retry snapshot", volInfo.PersistentVolumeClaim)
```
siva-portworx, line 747: let's retry snapshot"  is confusing, here we are not going to retry the snapshot, we will retry checking of status again. Please accordingly.

siva-portworx (approved):

PR #1547, opened 2023-10-26
PWX-34428: Add integration tests for DataExport with rsync.

adityadani:
**What type of PR is this?**
integration-test

- The integration test comprises of 4 tests:
  - Fail if destination PVC does not exist.
  - Fail if source PVC is in use.
  - Fail if destination PVC size is not same or greater than source PVC.
  - Successful data transfer between PVCs and the usage on the destination PVC increases.

No

No

Yes - 23.10.0 - so that this test runs for the current release branch


pp511 (approved):
Looks good.

diptiranjanpx (approved):

PR #1548, opened 2023-10-27
skip waiting for importer pods, if template PVCs already exist

Rohit-PX:
**What type of PR is this?**
integration-test

If PVC templates exist, don't wait for importer pods to be triggered,

no

no

23.10

### DIFF
--- a/test/integration_test/kubevirt_test.go
+++ b/test/integration_test/kubevirt_test.go
@@ -5,7 +5,6 @@ package integrationtest
 
 import (
 	"fmt"
-	"strings"
 	"testing"
 	"time"
 
@@ -15,7 +14,6 @@ import (
 	"github.com/portworx/torpedo/drivers/scheduler"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
-	corev1 "k8s.io/api/core/v1"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 )
 
@@ -29,7 +27,9 @@ const (
 	importerPodCompletionTimeout = 20 * time.Minute
 	importerPodRetryInterval     = 10 * time.Second
 
-	kubevirtTemplateNamespace = "openshift-virtualization-os-images"
+	kubevirtTemplateNamespace             = "openshift-virtualization-os-images"
+	kubevirtCDIStorageConditionAnnotation = "cdi.kubevirt.io/storage.condition.running.reason"
+	kubevirtCDIStoragePodPhaseAnnotation  = "cdi.kubevirt.io/storage.pod.phase"
 )
 
 func TestKubevirt(t *testing.T) {
@@ -37,7 +37,7 @@ func TestKubevirt(t *testing.T) {
 	err := setMockTime(nil)
 	require.NoError(t, err, "Error resetting mock time")
 
-	err = createTemplatePVC()
+	err = createTemplatePVC(t)
 	require.NoError(t, err, "Error creating template")
 
 	t.Run("kubevirtDeployFedoraVMWithClonePVC", kubevirtDeployFedoraVMWithClonePVC)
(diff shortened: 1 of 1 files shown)


test/integration_test/kubevirt_test.go
```diff
@@ -98,14 +98,33 @@ func validateVM(t *testing.T, virtualMachine kubevirtv1.VirtualMachine, vmName s
...
+		return fmt.Errorf("failed to get PVCs from %s namespace post templates", kubevirtTemplateNamespace)
+	}
+
+	// If number of PVCs is the same in the template namespace that means no new PVCs are created
+	// therefore no importer pods will be run, hence exit the function
+	if len(pvcsPreTemplate.Items) == len(pvcsPostTemplate.Items) {
+		logrus.Infof("No new PVCs were created for templates. Won't wait for importer pods since they will not be started.")
+		return nil
+	}
+
```
pureneelesh, line 127: We seem to ignore errors from the importer pod. If the state is not "running", we assume that it completed successfully. But what if the importer pod encountered an error?

So, I am wondering if we can just ignore the importer pod and check the pvc annotations. I observed that the PVC annotations change as below:
```
$ diff pvc.before pvc.after
5,7c5,7
<     cdi.kubevirt.io/storage.condition.running: "true"
<     cdi.kubevirt.io/storage.condition.running.message: ""
<     cdi.kubevirt.io/storage.condition.running.reason: Pod is running
---
>     cdi.kubevirt.io/storage.condition.running: "false"
>     cdi.kubevirt.io/storage.condition.running.message: Import Complete
>     cdi.kubevirt.io/storage.condition.running.reason: Completed
10c10
<     cdi.kubevirt.io/storage.pod.phase: Running
---
>     cdi.kubevirt.io/storage.pod.phase: Succeeded
``` 
We can look at the `cdi.kubevirt.io/storage.pod.phase` on the PVC. Also, we already hardcode the namespace in this code. Can we also just get the PVC by name and look at the annotations?

test/integration_test/kubevirt_test.go
```diff
@@ -97,54 +99,64 @@ func validateVM(t *testing.T, virtualMachine kubevirtv1.VirtualMachine, vmName s
...
+	pvcTemplates, err := core.Instance().GetPersistentVolumeClaims(kubevirtTemplateNamespace, nil)
+	if err != nil {
+		logrus.Errorf("failed to get PVCs from %s namespace post templates", kubevirtTemplateNamespace)
+		return false
+	}
+
+	// Loop through all PVCs and check for annotations that signify existing downloaded templates
+	for _, pvc := range pvcTemplates.Items {
+		storageCondition, _ := pvc.ObjectMeta.Annotations[kubevirtCdiStorageConditionAnnotation]
+		if storageCondition != "Completed" {
```
pureneelesh, line 150: nit: you can just do
```
if pvc.ObjectMeta.Annotations[kubevirtCdiStorageConditionAnnotation] != "Completed" {
```

pureneelesh, line 150: There is no need to duplicate these checks in this function. We can just call schedulerDriver.Schedule and return. In fact, we can just get rid of the whole function and move `schedulerDriver.Schedule` call to the createTemplatePVC function?

test/integration_test/kubevirt_test.go
```diff
@@ -29,15 +27,17 @@ const (
 	importerPodCompletionTimeout = 20 * time.Minute
 	importerPodRetryInterval     = 10 * time.Second
 
-	kubevirtTemplateNamespace = "openshift-virtualization-os-images"
+	kubevirtTemplateNamespace             = "openshift-virtualization-os-images"
+	kubevirtCdiStorageConditionAnnotation = "cdi.kubevirt.io/storage.condition.running.reason"
```
pureneelesh, line 31: nit: Cdi should be CDI since it is an acronym

pureneelesh (approved):

pp511 (approved):
lgtm

PR #1550, opened 2023-10-30
pb-4683: Added fix to pass the destination namespace in locator to PX, if ns mapping is passed.

siva-portworx:
**What type of PR is this?**
>bug

```
    pb-4683: Added fix to pass the destination namespace in locator to PX,
    if ns mapping is passed.

            - Added checks such the destination namespace value will be
              passed in the locator namespacelabel map, if it is present.
            - Similarly added fix for the storageclass map, where the values
              will be passed only the sc mapping provided.
```

No

```release-note
Issue: Restore of RWX portworx volume to a different namespace fails.
User Impact: User will not be able to restore the backup with RWX portworx volume to a different namespace in different destination cluster.
Resolution: Fixed the issue by passing the correct destination namespace to the portworx, So that restore to different namespace will succeed.

```

23.9

**Testing:**
**Tested below three cases:**
**Case 1:**
1) Custom restore with default storageclass and modified namespace.
2) In this case the storage class mapping will be nil and namespace mapping will have source/dest namespace.
3) Restore completed successfully and VM pod came up fine.
```
[root@ip-10-13-154-105 ~]# kubectl  get pods -n ttt1
NAME                           READY   STATUS    RESTARTS   AGE
virt-launcher-cirrosvm-jb4zn   1/1     Running   0          55m
[root@ip-10-13-154-105 ~]
```
Screenshot 2023-10-30 at 10 58 39 AM

**Case 2:**
1) Custom restore with modified storageclass and modified namespace.
2) In this case the storage class mapping will have source/dest SC and namespace mapping will have source/dest namespace.
3) Restore completed successfully and VM pod came up fine.
```
[root@ip-10-13-154-105 ~]# kubectl  get pods -n ttt2
NAME                           READY   STATUS    RESTARTS   AGE
virt-launcher-cirrosvm-kp7lx   1/1     Running   0          51m
[root@ip-10-13-154-105 ~]#
```
Screenshot 2023-10-30 at 10 59 04 AM

**Case 3:**
1) Default restore, where no SC and NS was selected.
2) Restote completed successfully and VM pod came up fine.
```
[root@ip-10-13-154-105 ~]# kubectl  get pods -n source
NAME                           READY   STATUS    RESTARTS   AGE
virt-launcher-cirrosvm-79kmr   1/1     Running   0          47m
[root@ip-10-13-154-105 ~]#
```
Screenshot 2023-10-30 at 10 59 58 AM

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -3711,7 +3711,7 @@ func (p *portworx) getCloudBackupRestoreSpec(
 	}
 	var restoreSpec *api.RestoreVolumeSpec
 	destStorageClass := storageClassMapping[sourceStorageClass]
-	if len(destStorageClass) == 0 {
+	if len(destStorageClass) == 0 && len(destinationNamespace) == 0 {
 		restoreSpec = &api.RestoreVolumeSpec{
 			IoProfileBkupSrc: true, // setting this for backward compatibility
 		}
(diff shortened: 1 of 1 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -3711,7 +3711,7 @@ func (p *portworx) getCloudBackupRestoreSpec(
 	}
 	var restoreSpec *api.RestoreVolumeSpec
 	destStorageClass := storageClassMapping[sourceStorageClass]
-	if len(destStorageClass) == 0 {
+	if len(destStorageClass) == 0 && len(destinationNamespace) == 0 {
```
prashanthpx, line 3714: To understand, why we need this check len(destinationNamespace) == 0 ?

drivers/volume/portworx/portworx.go
```diff
@@ -3720,31 +3720,35 @@ func (p *portworx) getCloudBackupRestoreSpec(
...
+		// Special handling for io_profile
+		// The default volumeSpec.IoProfile is set to sequential
+		// Check in storage class if an io profile is set. If not set use
+		// the io profile from the source
+		if _, ok := sc.Parameters[api.SpecIoProfile]; !ok {
+			// io profile not specified in storage class
+			restoreSpec.IoProfileBkupSrc = true
+		}
+	}
+	if len(destinationNamespace) != 0 {
```
prashanthpx, line 3746: Sholdn't we pass this irrespective of destinationNamespace always?

prashanthpx (approved):

drivers/volume/portworx/portworx.go
siva-portworx: Basically, we want do not want to set the locator labels, if the namespace mapping is not set.

drivers/volume/portworx/portworx.go
siva-portworx: If the namespace is not changing, we do not want to pass any details in the locator label. So added this check.

px-kesavan (approved):

PR #1552, opened 2023-10-31
PB-4632: Skip Default Kubernetes Endpoint

pallav-px:
**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: We need in case of all namespace backup we backup default k8s endpoint which will be already there in any cluster so stork will skip that resource and mark it as partial restore. So to avoid partial restore we skip the default endpoint

**Does this PR change a user-facing CRD or CLI?**: no

**Is a release note needed?**: yes
We are excluding the default kubernetes endpoint in default namespace to be backed up because wherever we will be doing the restore the k8s endpoint will always be there and will result in partial restore as this resource will be skipped because the resource already exist. To avoid this scenario we are not backing this specific object.

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: yes. 23.9

Screenshot 2023-11-03 at 2 40 10 PM

Screenshot 2023-11-03 at 2 39 55 PM

### DIFF
--- a/pkg/resourcecollector/endpoint.go
+++ b/pkg/resourcecollector/endpoint.go
@@ -18,6 +18,10 @@ func (r *ResourceCollector) endpointsToBeCollected(
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(object.UnstructuredContent(), &endpoint); err != nil {
 		return false, fmt.Errorf("error converting to endpoint: %v", err)
 	}
+	// Don't migrate the kubernetes endpoint
+	if endpoint.GetName() == "kubernetes" && endpoint.GetNamespace() == v1.NamespaceDefault {
+		return false, nil
+	}
 
 	if endpoint.Annotations != nil {
 		// collect endpoint which has include-resources annotation applied


lalat-das (approved):

lalat-das:
@pallav-px Please put your unit test screenshots, couldn't find it PB ticket also

pkg/resourcecollector/endpoint.go
```diff
@@ -19,6 +19,11 @@ func (r *ResourceCollector) endpointsToBeCollected(
 		return false, fmt.Errorf("error converting to endpoint: %v", err)
 	}
 
+	// Don't migrate the kubernetes endpoint
+	if endpoint.GetName() == "kubernetes" {
```
siva-portworx, line 23: Should we also add the namespace check for "default" namespace. The idea is only to prevent default namespace's "kubernetes" endpoint.

siva-portworx: Please add release note in the PR.

PR #1554, opened 2023-11-02
PWX-31058: Supporting exclude resource types in migration.

diptiranjanpx:
**What type of PR is this?**
improvement

Supporting `excludeResourceTypes` with migration to exclude a particular kind of resources

```release-note
Issue: A particular kind of k8s resource could not be excluded with migration.
User Impact: Using labels for excluding selectors does not work in certain cases where the resource is controlled by some operator which resets the user set labels.
Resolution: With the supported excludeResourceTypes, now a certain kind of resources can be excluded from migration.

```

```
23.11.0
```

**Test**
Tests with following has been done and have been updated in the ticket.
1. Excluding only a particular resource kind like Deployment or PersistentVolumeClaim.
2. Excluding multiple resource kinds 

Existing Integration test with the fix will be updated later.

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -2826,7 +2826,14 @@ func getRelatedCRDListWRTGroupAndCategories(client *apiextensionsclient.Clientse
 	return filteredCRDList
 }
 
-func (m *MigrationController) getResources(namespaces []string, migration *stork_api.Migration, labelSelectors map[string]string, excludeSelectors map[string]string, resourceCollectorOpts resourcecollector.Options, remote bool) ([]runtime.Unstructured, []v1.PersistentVolumeClaim, error) {
+func (m *MigrationController) getResources(
+	namespaces []string,
+	migration *stork_api.Migration,
+	labelSelectors map[string]string,
+	excludeSelectors map[string]string,
+	resourceCollectorOpts resourcecollector.Options,
+	remote bool,
+) ([]runtime.Unstructured, []v1.PersistentVolumeClaim, error) {
 
 	var objects []runtime.Unstructured
 	var pvcs []v1.PersistentVolumeClaim
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -625,15 +625,16 @@ func (m *MigrationController) purgeMigratedResources(
 		excludeSelectors = migration.Spec.ExcludeSelectors
 	}
 	excludeSelectors[StashCRLabel] = "true"
-	destObjects, _, err := rc.GetResources(
+
+	destObjects, _, err := m.getResources(
 		migrationNamespaces,
+		migration,
 		migration.Spec.Selectors,
 		excludeSelectors,
-		nil,
-		migration.Spec.IncludeOptionalResourceTypes,
-		false,
 		resourceCollectorOpts,
+		true,
 	)
+
 	if err != nil {
 		m.recorder.Event(migration,
 			v1.EventTypeWarning,
@@ -642,14 +643,13 @@ func (m *MigrationController) purgeMigratedResources(
 		log.MigrationLog(migration).Errorf("Error getting resources: %v", err)
 		return err
 	}
-	srcObjects, _, err := m.resourceCollector.GetResources(
+	srcObjects, _, err := m.getResources(
 		migrationNamespaces,
+		migration,
 		migration.Spec.Selectors,
 		migration.Spec.ExcludeSelectors,
-		nil,
-		migration.Spec.IncludeOptionalResourceTypes,
-		false,
 		resourceCollectorOpts,
+		false,
 	)
 	if err != nil {
 		m.recorder.Event(migration,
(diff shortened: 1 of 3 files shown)


pkg/resourcecollector/resourcecollector.go
```diff
@@ -379,6 +402,106 @@ func (r *ResourceCollector) GetResourcesForType(
 	return objects, pvcsWithOwnerReference, nil
 }
 
+func (r *ResourceCollector) GetResourcesExcludingTypes(
```
siva-portworx, line 405: Should we consider overloading the GetResourcesForType with both the include and exclude parameter?

diptiranjanpx: Trying to overload `GetResourcesForType` with excludeType would have been very different code, as the function has already knowledge of for which resource type we have to collect the resources.

Rather I found the excludeType could be easily supported with `GetResources` and again did not want to change the definition of `GetResources` as it is being vendored into different projects. Anyway the common part has been carved into a different function.

pkg/resourcecollector/resourcecollector.go
```diff
@@ -179,6 +179,29 @@ func resourceToBeCollected(resource metav1.APIResource, grp schema.GroupVersion,
 	return GetSupportedK8SResources(resource.Kind, optionalResourceTypes)
 }
 
+func resourceToBeCollectedWithExcludeTypes(resource metav1.APIResource, grp schema.GroupVersion, crdKinds []metav1.GroupVersionKind, optionalResourceTypes []string, exlcudeTypes []string) bool {
```
pp511, line 182: We could reuse the function resourceToBeCollected. The only thing we have to check extra is 
```
	for _, excludeType := range exlcudeTypes {
		if strings.EqualFold(resource.Kind, excludeType) {
			return false
		}
	}
```
but for rest we could just call that function. Isn't it?

pkg/resourcecollector/resourcecollector.go
```diff
@@ -379,6 +402,106 @@ func (r *ResourceCollector) GetResourcesForType(
...
+	resourceMap := make(map[types.UID]bool)
+	var crdResources []metav1.GroupVersionKind
+	var crdList *stork_api.ApplicationRegistrationList
+	if !reflect.ValueOf(storkcache.Instance()).IsNil() {
+		crdList, err = storkcache.Instance().ListApplicationRegistrations()
+	} else {
+		crdList, err = r.storkOps.ListApplicationRegistrations()
+	}
+	if err != nil {
+		logrus.Warnf("Unable to get registered crds, err %v", err)
```
pp511, line 430: We don't error out if we fail to get registred RDs?

pkg/resourcecollector/resourcecollector.go
```diff
@@ -379,6 +402,106 @@ func (r *ResourceCollector) GetResourcesForType(
...
+	}
+
+	for _, group := range r.discoveryHelper.Resources() {
+		groupVersion, err := schema.ParseGroupVersion(group.GroupVersion)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		for _, resource := range group.APIResources {
+			if !resourceToBeCollectedWithExcludeTypes(resource, groupVersion, crdResources, optionalResourceTypes, excludeResourceTypes) {
```
pp511, line 455: nit..split it in multiple lines

pkg/resourcecollector/resourcecollector.go
```diff
@@ -379,6 +402,106 @@ func (r *ResourceCollector) GetResourcesForType(
...
+	}
+
+	// Creating a list of PVCs with owner reference before calling prepareResourcesForCollection
+	// prepareResourcesForCollection can update the PVC metadata which is required when updating
+	// owner references on the destination PVC objects
+	var pvcsWithOwnerReference []v1.PersistentVolumeClaim
+	var pvc v1.PersistentVolumeClaim
+	for _, o := range pvcObjectsWithOwnerRef {
+		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(o.UnstructuredContent(), &pvc); err != nil {
+			logrus.Warnf("unable to cast pvcs with owner reference: %v", err)
```
pp511, line 492: nit..Unable

pkg/resourcecollector/resourcecollector.go
pp511: I am also trying to understand what was the issue with using GetResources and just adding `if !resourceToBeCollectedWithExcludeTypes(resource, groupVersion, crdResources, optionalResourceTypes, excludeResourceTypes)` instead of creating a new function?
Can you point me to what else is it that we need to do different for exclude type in this function?

pp511: Ok I read your comment about vendoring later..I guess it's ok then...

pkg/migration/controllers/migration.go
```diff
@@ -2816,3 +2816,55 @@ func getRelatedCRDListWRTGroupAndCategories(client *apiextensionsclient.Clientse
 
 	return filteredCRDList
 }
+
+func (m *MigrationController) getResources(namespaces []string, migration *stork_api.Migration, labelSelectors map[string]string, excludeSelectors map[string]string, resourceCollectorOpts resourcecollector.Options, remote bool) ([]runtime.Unstructured, []v1.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
diptiranjanpx: Done

pkg/resourcecollector/resourcecollector.go
```diff
@@ -179,6 +179,29 @@ func resourceToBeCollected(resource metav1.APIResource, grp schema.GroupVersion,
...
+	for _, excludeType := range exlcudeTypes {
+		if strings.EqualFold(resource.Kind, excludeType) {
+			return false
+		}
+	}
+
+	// Include all namespaced CRDs
+	for _, res := range crdKinds {
+		if res.Kind == resource.Kind &&
+			res.Group == grp.Group && res.Version == grp.Version && resource.Namespaced {
```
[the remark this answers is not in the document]
diptiranjanpx: changed in resourceToBeCollected

siva-portworx (approved):

pp511 (approved):
lgtm

PR #1556, opened 2023-11-07
PWX-34866: Fix error message when namespaces and namespaceSelectors a…

pp511:
…re invalid.

**What type of PR is this?**
When invalid namespaces and namespaceSelectors are specified by mistake, the error during migration is following  "namespaces for migration cannot be empty". This is sreturned from following https://github.com/libopenstorage/stork/blob/df7b1d2405dd816f71491f7fc92d85bbc7d11a95/drivers/volume/portworx/portworx.go#L2591
This code is triggered from MigrationStageVolumes. If extracted migrationNamespaces are emptry we should fail early. Also this message should be more descriptive. Existing error message gives an impression that we need "namespaces" to be non empty which isn't true as one can specify namespaceSelectors too.

**Testing notes**:
* Didn't do any testing yet. Creating clusters to verify the error message. Will post here once done.

no

no

yes, 23.10.0

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -334,6 +334,15 @@ func (m *MigrationController) handle(ctx context.Context, migration *stork_api.M
 			err.Error())
 		return nil
 	}
+	if len(migrationNamespaces) == 0 {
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

olavangad-px (approved):

Rohit-PX (approved):
LGTM

PR #1560, opened 2023-11-09
PWX-34118: Skip service accounts, roles and rolebindings as part of pruning of resources with ownerreferences.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

Some of the operators like postgres were not able to reconcile on the CR and create the pods if service account, roles and rolebindings were already created.

```release-note
Issue: Postgres operator throws error of serviceaccount, role and rolebindings already exist after migration.
User Impact: Unable to scale up postgres app installed via openshift operator hub after migration.
Resolution: Postgres pods came up fine after skipping the resources serviceaccount, role and rolebindings as part of resource migration if they have ownerreference set.

```

```
23.11.0
```

**Test**
Tested with crunchydata postgres operator 5.4.3 and details have been updated in the ticket.

### DIFF
--- a/pkg/resourcecollector/resourcecollector.go
+++ b/pkg/resourcecollector/resourcecollector.go
@@ -755,6 +755,9 @@ func (r *ResourceCollector) pruneOwnedResources(
 							objectType.GetKind() != "StatefulSet" &&
 							objectType.GetKind() != "ReplicaSet" &&
 							objectType.GetKind() != "DeploymentConfig" &&
+							objectType.GetKind() != "ServiceAccount" &&
+							objectType.GetKind() != "Role" &&
+							objectType.GetKind() != "RoleBinding" &&
 							objectType.GetKind() != "Service" {
 							continue
 						}


pp511 (approved):
looks ok

pkg/resourcecollector/resourcecollector.go
```diff
@@ -755,6 +755,9 @@ func (r *ResourceCollector) pruneOwnedResources(
 							objectType.GetKind() != "StatefulSet" &&
 							objectType.GetKind() != "ReplicaSet" &&
 							objectType.GetKind() != "DeploymentConfig" &&
+							objectType.GetKind() != "ServiceAccount" &&
+							objectType.GetKind() != "Role" &&
+							objectType.GetKind() != "RoleBinding" &&
```
pp511, line 760: Why do all these checks have to be in the for loop? We should check it outside of  `for _, owner := range owners `loop. Am I missing anything?

pp511 (changes requested):

pkg/resourcecollector/resourcecollector.go
diptiranjanpx: There is a check for `if a pod is an owner` is there before we are checking for if object is not one among these types (deploy, sts, service etc...). If we take the decision of collecting the object before the for loop, there can be scenario the owner might be a pod itself and the real intention was to skip it.

Anyway we will get rid of all these checks and make it generic for all resources not to be collected  or something similar. So don't want to alter the logic now.

pp511 (approved):

siva-portworx (approved):

PR #1562, opened 2023-11-13
[PWX-34708] Support multi-DR for normal application resources and CRs with suspend options

olavangad-px:
**What type of PR is this?**
>bug

[PWX-34708](https://portworx.atlassian.net/browse/PWX-34708)
To carry forward the appropriate replicas value in case of multi-DR
For now we have only fixed the bug for normal application resources and CRs with suspend options defined.
Changes for CRs with stashStrategy enabled will be taken up separately.

No

Yes, TBD

Yes, 23.11.0

**Testing Details**
Have manually tested my changes in a setup of 3 clusters with ASYNC-ASYNC relationship. (Metro or Async should not matter from code point of view)
Have tested that correct annotation is carried forward for following cases -> 

1. Deployment with replicas set to 2, is migrated from C1 to C2. And then from C2 to C3.
2. MongoDBCommunity resource with spec.members set to 3, is migrated from C1 to C2. And then from C2 to C3. (suspend path spec.members with suspend.Type int)
3. Kubevirt virtualmachine with runStrategy set to Always , is migrated from C1 to C2. And then from C2 to C3. (suspend path spec.runStrategy with suspend.Type string and suspend.Value Halted)

Have verified that created resources have expected annotation values in both C2 and C3.

[PWX-34708]: https://portworx.atlassian.net/browse/PWX-34708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -1470,9 +1470,6 @@ func (m *MigrationController) prepareApplicationResource(
 			}
 		}
 	}
-	if *migration.Spec.StartApplications {
-		return nil
-	}
 
 	// Reset the replicas to 0 and store the current replicas in an annotation
 	replicas, found, err := unstructured.NestedInt64(content, "spec", "replicas")
@@ -1483,6 +1480,38 @@ func (m *MigrationController) prepareApplicationResource(
 		replicas = 1
 	}
 
+	annotations, found, err := unstructured.NestedStringMap(content, "metadata", "annotations")
+	if err != nil {
+		return err
+	}
+	if !found {
+		annotations = make(map[string]string)
+	}
+	migrationReplicasAnnotationValue, replicaAnnotationIsPresent := annotations[StorkMigrationReplicasAnnotation]
+
+	if replicas == 0 && replicaAnnotationIsPresent {
+		//Only if the actual replica count is 0 and migrationReplicas annotation is more than 0, then carry forward the annotation.
+		migrationReplicas, err := strconv.ParseInt(migrationReplicasAnnotationValue, 10, 64)
+		if err != nil {
+			return err
+		}
+		if migrationReplicas > 0 {
+			//carry forward the migrationReplica value
+			replicas = migrationReplicas
+			//we reset the actual replica count as well in case the migration has startApplications enabled
+			err = unstructured.SetNestedField(content, replicas, "spec", "replicas")
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	//No need to scale down if StartApplications is set to True in the migration spec
+	if *migration.Spec.StartApplications {
+		return nil
+	}
+
+	//we set the actual replica count to 0 to scale down the resource on target cluster
 	err = unstructured.SetNestedField(content, int64(0), "spec", "replicas")
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migration.go
```diff
@@ -1547,18 +1567,55 @@ func (m *MigrationController) prepareCRDClusterResource(
 					return fmt.Errorf("unable to find suspend path, err: %v", err)
 				}
 				disableVersion = int64(0)
+				if curr == 0 && suspendAnnotationIsPresent {
+					//suspendAnnotation has value set as {currVal + "," + suspend.Value}
+					//we need to extract only the currVal from it
+					annotationValue := strings.Split(SuspendAnnotationValue, ",")[0]
```
diptiranjanpx, line 1573: Is it really required to do the split and again modify the `curr` value and later again add it to the annotation?

If curr == 0 &&  suspendAnnotationIsPresent 
       We don't have to set anything, we can take the object with the annotation itself.

pkg/migration/controllers/migration.go
```diff
@@ -1547,18 +1567,55 @@ func (m *MigrationController) prepareCRDClusterResource(
...
+						}
+					}
+				}
 				currVal = curr
 			} else {
 				return fmt.Errorf("invalid type %v to suspend cr", suspend.Type)
 			}
 
+			//No need to scale down if StartApplications is set to True in the migration spec
+			if *migration.Spec.StartApplications {
```
diptiranjanpx, line 1614: Why is this required, can you please explain?

pp511 (commented):
Did take one pass at it. Looks fine so far.
Will take a second look at it again tomorrow and  will dry-run with a concrete example to make sure everything is ok.

pkg/migration/controllers/migration.go
olavangad-px: 1. The condition I am checking as of now is curr == 0 && suspendAnnotation is > 0. So I need to extract the annotation value for that.
2. Additionally I am also updating the suspend path value with the annotationValue in line #1583
This is done for the scenario when migration.spec.startApplications is true.

Let's consider the case where resources were migrated from C1 to C2. Now the resource in C2 is scaled down so actual replica value will be 0 while annotation will have the replicaCount from C1 (let's say it is 3)
Now if I migrate the resource from C2 to C3 with startApplications as true, I would not be scaling down the resource and migrating it as is. For this case I should have modified the actual replica value in the resource as well. This is what is done in #1583

So that in #1614 when we return nil because startApplications is true and we do not need to scale down, the resource applied in C3 will have replicaCount set to 3.

pkg/migration/controllers/migration.go
olavangad-px: In the existing implementation this check is added at the beginning of the functions (#1519 in previous code) since we do not need to scale down the resource if we are going start the applications in the target cluster.

Since as part of this bug, we may want a different value of replica compared to the existing count (which could be 0 and the real useful value is stored in the annotation), I have moved the check to a different place.

I have explained how we get the appropriate value for this in the above comment with an example.

pkg/migration/controllers/migration.go
diptiranjanpx: Got it thanks, this is required for `startApplications : true`. I did not think about that.

diptiranjanpx (approved):

pp511 (approved):
Lgtm

PR #1567, opened 2023-11-16
pb-4887: Added fix in handling the already exists error while creating

siva-portworx:
**What type of PR is this?**
>bug

```
    pb-4887: Added fix in handling the already exists error while creating
    volumesnapshotclass during restore.

            - If we have csi-snapshot-webhook webhook, the creation of
              volumesnapshotclass fails with different eror message for already exists error.
            - So add a get call and then calling create volumesnapshot call,
              if Get failed with NotFound error.
```

no

```release-note
Issue: csi based restore fails on some of the setup that had csi-snapshot-webhook admin webhook as we get different error for alreadyexist error, while creating the volumesnapshotclass resource.
User Impact: CSI based restore was failing on the setup that has csi-snapshot-webhook admin webhook
Resolution: Added a get call before create call, such that create call will happen only the get failed with "NotFound" error.

```

23.9

**Testing:**
1) Validated the Native CSI restore and kdmp localsnapshot restore, which was failing with out fix.

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -1409,22 +1409,37 @@ func (c *csi) restoreVolumeSnapshotClass(vsClass interface{}) (interface{}, erro
 	if c.v1SnapshotRequired {
 		vsClass.(*kSnapshotv1.VolumeSnapshotClass).ResourceVersion = ""
 		vsClass.(*kSnapshotv1.VolumeSnapshotClass).UID = ""
-		newVSClass, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Create(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass), metav1.CreateOptions{})
+		_, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Get(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass).Name, metav1.GetOptions{})
 		if err != nil {
-			if k8s_errors.IsAlreadyExists(err) {
-				return vsClass, nil
+			// did not find vs class, create one
+			if k8s_errors.IsNotFound(err) {
+				newVSClass, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Create(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass), metav1.CreateOptions{})
+				if err != nil {
+					if !k8s_errors.IsAlreadyExists(err) {
+						return nil, err
+					}
+				}
 			}
-			return nil, err
+		} else {
+			return vsClass, nil
 		}
+
 	} else {
 		vsClass.(*kSnapshotv1beta1.VolumeSnapshotClass).ResourceVersion = ""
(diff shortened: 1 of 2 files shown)


px-kesavan (approved):
LGTM

prashanthpx (approved):

lalat-das (approved):

kshithijiyer-px (approved):
LGTM

drivers/volume/csi/csi.go
```diff
@@ -1409,22 +1409,37 @@ func (c *csi) restoreVolumeSnapshotClass(vsClass interface{}) (interface{}, erro
...
+				newVSClass, err = c.snapshotClient.SnapshotV1beta1().VolumeSnapshotClasses().Create(context.TODO(), vsClass.(*kSnapshotv1beta1.VolumeSnapshotClass), metav1.CreateOptions{})
+				if err != nil {
+					if !k8s_errors.IsAlreadyExists(err) {
+						return nil, err
+					}
+				}
 			}
-			return nil, err
+		} else {
+			return vsClass, nil
```
mkoppal-px, line 1442: Same here

drivers/volume/csi/csi.go
```diff
@@ -1409,22 +1409,37 @@ func (c *csi) restoreVolumeSnapshotClass(vsClass interface{}) (interface{}, erro
...
+				newVSClass, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Create(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass), metav1.CreateOptions{})
+				if err != nil {
+					if !k8s_errors.IsAlreadyExists(err) {
+						return nil, err
+					}
+				}
 			}
-			return nil, err
+		} else {
+			return vsClass, nil
```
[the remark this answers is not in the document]
siva-portworx: Yes it is needed. We need to return the copy of the existing vsclass, if get call passes.

drivers/volume/csi/csi.go
siva-portworx: Yes it is needed. We need to return the copy of the existing vsclass, if get call passes.

mkoppal-px (approved):
LGTM

drivers/volume/csi/csi.go
```diff
@@ -1409,22 +1409,37 @@ func (c *csi) restoreVolumeSnapshotClass(vsClass interface{}) (interface{}, erro
 	if c.v1SnapshotRequired {
 		vsClass.(*kSnapshotv1.VolumeSnapshotClass).ResourceVersion = ""
 		vsClass.(*kSnapshotv1.VolumeSnapshotClass).UID = ""
-		newVSClass, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Create(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass), metav1.CreateOptions{})
+		_, err = c.snapshotClient.SnapshotV1().VolumeSnapshotClasses().Get(context.TODO(), vsClass.(*kSnapshotv1.VolumeSnapshotClass).Name, metav1.GetOptions{})
```
diptiranjanpx, line 1412: why not return the vsClass object returned by this Get function?

PR #1569, opened 2023-11-20
[PWX-34991] Fetching rules from migration namespace instead of namespace to migrate

olavangad-px:
**What type of PR is this?**
>bug

As part of existing code, when users create migrationschedule and clusterpair in the admin namespace, they have to create preExec and postExec rules in both the admin Namespace as well as the namespaces to be migrated.
We fix this inconvenience by fetching the rules from adminNamespace only. 

No

Yes, TBD

Yes, 23.11.0

**Testing Details**
Have tested the changes in a local cluster and verified the bug as well as the fix.
Validated the fix by creating the migrationSchedule in the admin namespace and migrating default namespace. For this case the rules had to be created in the admin namespace.

### DIFF
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -907,11 +907,11 @@ func (m *MigrationController) runPreExecRule(migration *stork_api.Migration, mig
 			return nil, nil
 		}
 	}
-	terminationChannels := make([]chan bool, 0)
 	r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, migration.Namespace)
 	if err != nil {
 		return nil, err
 	}
+	terminationChannels := make([]chan bool, 0)
 	for _, ns := range migrationNamespaces {
 		ch, err := rule.ExecuteRule(r, rule.PreExecRule, migration, ns)
 		if err != nil {
--- a/pkg/migration/controllers/migration.go
+++ b/pkg/migration/controllers/migration.go
@@ -909,7 +909,7 @@ func (m *MigrationController) runPreExecRule(migration *stork_api.Migration, mig
 	}
 	terminationChannels := make([]chan bool, 0)
 	for _, ns := range migrationNamespaces {
-		r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, ns)
+		r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, migration.Namespace)
 		if err != nil {
 			for _, channel := range terminationChannels {
 				channel <- true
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

pkg/migration/controllers/migration.go
```diff
@@ -909,7 +909,7 @@ func (m *MigrationController) runPreExecRule(migration *stork_api.Migration, mig
 	}
 	terminationChannels := make([]chan bool, 0)
 	for _, ns := range migrationNamespaces {
-		r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, ns)
+		r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, migration.Namespace)
```
pp511, line 912: The rule will remain same in all iterations of the for loop. This call can be moved outside the for loop to avoid calling it multiple times in case of migration schedule in admin namespace.

pkg/migration/controllers/migration.go
```diff
@@ -933,7 +933,7 @@ func (m *MigrationController) runPreExecRule(migration *stork_api.Migration, mig
 
 func (m *MigrationController) runPostExecRule(migration *stork_api.Migration, migrationNamespaces []string) error {
 	for _, ns := range migrationNamespaces {
-		r, err := storkops.Instance().GetRule(migration.Spec.PostExecRule, ns)
+		r, err := storkops.Instance().GetRule(migration.Spec.PostExecRule, migration.Namespace)
```
pp511, line 936: Same concern as above.

pp511 (changes requested):

pkg/migration/controllers/migration.go
olavangad-px: Makes sense. Will address

pkg/migration/controllers/migration.go
olavangad-px: Makes sense. Will address

pkg/migration/controllers/migration.go
```diff
@@ -908,15 +908,14 @@ func (m *MigrationController) runPreExecRule(migration *stork_api.Migration, mig
...
-		r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, ns)
-		if err != nil {
-			for _, channel := range terminationChannels {
-				channel <- true
-			}
-			return nil, err
+	r, err := storkops.Instance().GetRule(migration.Spec.PreExecRule, migration.Namespace)
+	if err != nil {
+		for _, channel := range terminationChannels {
+			channel <- true
```
pp511, line 914: Overall I am having a little hard time understanding the implementation of terminationChannels. I do not think it is implemented correctly. But just from the point of view of this PR, terminationChannels is initialized right above this call and we will never be entering inside the for loop. 
We can simply return from this function if err is not nil just like we are doing in case of postExecRule.

olavangad-px: Ya, I also was not clear on that part but did not want to change the existing behaviour

pp511: Let's remove the for loop and just return err.

pp511 (approved):
lgtm

pkg/migration/controllers/migration.go
diptiranjanpx: In earlier code, both GetRule and ExecuteRule were inside the loop. It made sense to append the terminationchannels and looping over inside it. Now that GetRule is outside, we can simply return as @pp511 is pointing.

diptiranjanpx (approved):

PR #1570, opened 2023-11-20
[PWX-35097]Adding ExcludeResourceTypes param in storkctl create migrationschedule

olavangad-px:
**What type of PR is this?**
>improvement

Adding support for new migrationSchedule param `ExcludeResourceTypes` as part of storkctl create migrationschedule cli

Yes
```
[root@ip-10-13-192-45 configs]# storkctl create migrationschedule -h
Create a migration schedule

Usage:
  storkctl create migrationschedules [flags]

Aliases:
  migrationschedules, migrationschedule

Flags:
      --admin-cluster-pair string            Specify the name of the admin ClusterPair used to migrate cluster-scoped resources, if the ClusterPair is present in a non-admin namespace
      --annotations stringToString           Add required annotations to the resource in comma-separated key value pairs. key1=value1,key2=value2,...  (default [])
  -c, --cluster-pair string                  Specify the name of the ClusterPair in the same namespace to be used for the migration
      --disable-auto-suspend                 Prevent automatic suspension of DR migration schedules on the source cluster in case of a disaster
      --disable-skip-deleted-namespaces      If present, Stork will fail the migration when it encounters a namespace that is deleted but specified in the namespaces field. By default, Stork ignores deleted namespaces during migration
      --exclude-resource-types strings       Comma-separated list of the specific resource types which need to be excluded from migration, ex: Deployment,PersistentVolumeClaim
      --exclude-resources                    If present, Kubernetes resources will not be migrated
      --exclude-selectors stringToString     Resources with the provided labels will be excluded from the migration. All the labels provided in this option will be OR'ed (default [])
      --exclude-volumes                      If present, the underlying Portworx volumes will not be migrated. This is the only allowed and default behaviour in sync-dr use-cases or when storage options are not provided in the cluster pair
  -h, --help                                 help for migrationschedules
      --ignore-owner-references-check        If set, resources with ownerReferences will also be migrated, even if the corresponding owners are getting migrated
      --include-jobs                         Set this flag to ensure that K8s Job resources are migrated. By default, the Job resources are not migrated
      --include-network-policy-with-cidr     If set, the underlying network policies will be migrated even if a fixed CIDR is present on them
  -i, --interval int                         Specify the time interval, in minutes, at which Stork should trigger migrations (default 30)
      --namespace-selectors stringToString   Resources in the namespaces with the specified namespace labels will be migrated. All the labels provided in this option will be OR'ed (default [])
      --namespaces strings                   Specify the comma-separated list of namespaces to be included in the migration
      --post-exec-rule string                Specify the name of the rule to be executed after every migration is triggered
      --pre-exec-rule string                 Specify the name of the rule to be executed before every migration is triggered
      --purge-deleted-resources              Set this flag to automatically delete Kubernetes resources in the target cluster when they are removed from the source cluster
  -s, --schedule-policy-name string          Name of the schedule policy to use. If you want to create a new interval policy, use the --interval flag instead (default "default-migration-policy")
      --selectors stringToString             Only resources with the provided labels will be migrated. All the labels provided in this option will be OR'ed (default [])
      --skip-service-update                  If set, service objects will be skipped during migration
  -a, --start-applications                   If present, the applications will be scaled up on the target cluster after a successful migration
      --suspend                              Flag to denote whether schedule should be suspended on creation
      --transform-spec string                Specify the ResourceTransformation spec to be applied during migration
```

No

Yes, 23.11.0

**Testing Details**
Tested manually without excludeResourceTypes flag, invalid type in excludeResourceTypes and valid types.

### DIFF
--- a/pkg/storkctl/migrationschedule.go
+++ b/pkg/storkctl/migrationschedule.go
@@ -2,6 +2,7 @@ package storkctl
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
@@ -53,6 +54,7 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 	var includeJobs bool
 	var selectors map[string]string
 	var excludeSelectors map[string]string
+	var excludeResourceTypes []string
 
 	createMigrationScheduleCommand := &cobra.Command{
 		Use:     migrationScheduleSubcommand,
@@ -197,6 +199,16 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 				includeOptionalResourceTypes = append(includeOptionalResourceTypes, "Job")
 			}
 
+			//validate excludeResourceTypes
+			if len(excludeResourceTypes) != 0 {
+				resourceTypes := strings.Join(excludeResourceTypes, ",")
+				excludeResourceTypes, err = getResourceTypes(resourceTypes, ioStreams)
+				if err != nil {
+					util.CheckErr(err)
+					return
+				}
+			}
+
 			migrationSchedule := &storkv1.MigrationSchedule{
 				ObjectMeta: meta.ObjectMeta{Annotations: annotations},
 				Spec: storkv1.MigrationScheduleSpec{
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

pkg/storkctl/migrationschedule.go
```diff
@@ -197,6 +199,16 @@ func newCreateMigrationScheduleCommand(cmdFactory Factory, ioStreams genericclio
 				includeOptionalResourceTypes = append(includeOptionalResourceTypes, "Job")
 			}
 
+			//validate excludeResourceTypes
+			if len(excludeResourceTypes) != 0 {
+				resourceTypes := strings.Join(excludeResourceTypes, ",")
+				excludeResourceTypes, err = getResourceTypes(resourceTypes, ioStreams)
```
olavangad-px, line 205: I am reusing getResourceTypes function from applicationBackup code path.
Discussed with Dipti that this cannot be unit tested because of the aeclient.Discovery() being used in the code

pp511 (approved):

PR #1572, opened 2023-11-22
Pb 4872:Added code to pass the storageclass value as label in the dataexport CR.

siva-portworx:
**What type of PR is this?**
>improvement

```
  pb-4872: Added code to pass the storageclass value as label in the
    dataexport CR.
```

no

no. This change does not affect the functionality that customer use.
```release-note
Issue:
User Impact:
Resolution

```

23.9.1 branch


dbinnal-px (approved):

prashanthpx (approved):

siva-portworx:
Since this fix is blocking customer release, force merging it.

PR #1574, opened 2023-11-24
Pb 4996 post exec running forever

lalat-das:
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: The postexec rule is getting executed many times and backups are timing out for large sizzed volume based backups. 

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: 23.9

Unit test done: ( for application having 3 volume backup size of 33GB
1. parallel schedule backup triggered with rules set and verified pre and post exec ran exactly once for all the backups
2. Executed the above case with two parallel manual backup and a schedule backup
3. Verified with debug log that the termination channel map and executed flag map is removing the entry for all the required backup.
4. Added an action to rule and verified it is being executed as part of post exec rule. There is an issue observed, we are working on it. it is unrelated to the changes to this  PR.

The results are posted in PB ticket 
https://portworx.atlassian.net/browse/PB-4996


lalat-das:
This is raised from a wrong branch hence closing it.

PR #1577, opened 2023-11-27
Pb-4844: Changes in stork for it to enable debug level log messages for kopia operations

kgarg-px:
**What type of PR is this?**

>improvement

This PR introduces a change where we can access debug level logs for kopia operations. We need the this, so that in future whenever required, we could use it to view logs in debug level.

```release-note
Issue:
User Impact:
Resolution

```

Link to PX-Backup PR related to this particular PR is: https://github.com/portworx/px-backup/pull/1407
Link to KDMP PR related to this particular PR is: https://github.com/portworx/kdmp/pull/324

Attaching some screenshots for the reference:
Screenshot 2023-11-27 at 4 05 47 PM
Screenshot 2023-11-27 at 4 12 44 PM
Screenshot 2023-11-27 at 4 15 43 PM
Screenshot 2023-11-27 at 10 50 26 AM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -333,7 +333,7 @@ replace (
 	github.com/libopenstorage/autopilot-api => github.com/libopenstorage/autopilot-api v0.6.1-0.20210301232050-ca2633c6e114
 	github.com/libopenstorage/openstorage => github.com/libopenstorage/openstorage v0.0.0-20230511212757-41751b27d69f
 	github.com/libopenstorage/secrets => github.com/libopenstorage/secrets v0.0.0-20220413195519-57d1c446c5e9
-	github.com/portworx/kdmp => github.com/portworx/kdmp v0.4.1-0.20231130175942-e8a1e0c64972
+	github.com/portworx/kdmp => github.com/portworx/kdmp v0.4.1-0.20231218141125-9b389a7c1a98
 	github.com/portworx/sched-ops => github.com/portworx/sched-ops v1.20.4-rc1.0.20230930052008-186afa1e82c4
 	github.com/portworx/torpedo => github.com/portworx/torpedo v0.0.0-20231013194137-84e2957806f0
 	gopkg.in/fsnotify.v1 v1.4.7 => github.com/fsnotify/fsnotify v1.4.7
--- a/go.sum
+++ b/go.sum
@@ -2894,8 +2894,8 @@ github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 h1:Jamvg5psRI
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/polyfloyd/go-errorlint v1.0.5/go.mod h1:APVvOesVSAnne5SClsPxPdfvZTVDojXh1/G3qb5wjGI=
 github.com/portworx/dcos-secrets v0.0.0-20180616013705-8e8ec3f66611/go.mod h1:4hklRW/4DQpLqkcXcjtNprbH2tz/sJaNtqinfPWl/LA=
-github.com/portworx/kdmp v0.4.1-0.20231130175942-e8a1e0c64972 h1:9UBxkJyCz+UzsUMa+dAYhThmQ0T9nShQJePVQrDEKpk=
-github.com/portworx/kdmp v0.4.1-0.20231130175942-e8a1e0c64972/go.mod h1:mMZUDNoAfslBw6DxEXKN2KSEt14l/HIGtt1CzgOghp0=
+github.com/portworx/kdmp v0.4.1-0.20231218141125-9b389a7c1a98 h1:gdCYAMznWjlQF0eepXtRakdOYI98STjeX41BcF8z1sI=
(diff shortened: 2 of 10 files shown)


drivers/volume/kdmp/kdmp.go
```diff
@@ -371,6 +375,24 @@ func (k *kdmp) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.
...
 	return volumeInfos, nil
 }
+
+// Check whether to run kopia executor in debug mode
+func enableKopiaExecutorDebugMode() bool {
+	logrus.Info("enableKopiaExecutorDebugMode entering")
+	kdmpData, err := core.Instance().GetConfigMap(stork_driver.KdmpConfigmapName, stork_driver.KdmpConfigmapNamespace)
+	if err != nil {
+		logrus.Info("enableKopiaExecutorDebugMode unable to read the cm")
+		logrus.Tracef("error readig kdmp config map: %v", err)
```
vsundarraj-px, line 385: nit: Typo on readig

drivers/volume/kdmp/kdmp.go
```diff
@@ -371,6 +375,24 @@ func (k *kdmp) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.
 	}
 	return volumeInfos, nil
 }
+
+// Check whether to run kopia executor in debug mode
```
vsundarraj-px, line 379: If its check function, then the function name is missleading. enableKopiaExecutorDebugMode() means enabling it not checking.
Suggestion:  IsKopiaExecutorDebugModeEnabled()

drivers/volume/kdmp/kdmp.go
```diff
@@ -371,6 +375,24 @@ func (k *kdmp) GetBackupStatus(backup *storkapi.ApplicationBackup) ([]*storkapi.
...
+func enableKopiaExecutorDebugMode() bool {
+	logrus.Info("enableKopiaExecutorDebugMode entering")
```
vsundarraj-px, line 381: This is ok while developing but not for production, suggestion to remove this info message.

drivers/volume/kdmp/kdmp.go
```diff
@@ -765,6 +787,12 @@ func (k *kdmp) StartRestore(
 		dataExport.Annotations[utils.SkipResourceAnnotation] = "true"
 		dataExport.Annotations[utils.BackupObjectUIDKey] = backupUID
 		dataExport.Annotations[pvcUIDKey] = bkpvInfo.PersistentVolumeClaimUID
+		logrus.Info("enableKopiaExecutorDebugMode calling")
+		if enableKopiaExecutorDebugMode() {
+			logrus.Info("enableKopiaExecutorDebugMode able to find the debug mode present in kdmp-config cm")
```
vsundarraj-px, line 792: Redundant Info log. Its already getting printed in check function.. hence this should be removed to reduce logfile payload.

drivers/volume/kdmp/kdmp.go
```diff
@@ -765,6 +787,12 @@ func (k *kdmp) StartRestore(
...
+			dataExport.Annotations[utils.KopiaDebugModeEnabled] = "true"
+			logrus.Info("enableKopiaDebugMode annotation added in the CR")
```
vsundarraj-px, line 794: Same as above, if you are already checking checkFunction you can avoid this. Or you can remove the log.info in checkfunction() if you are printing here.. Too much of log messages will create high logfile payload.

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -361,6 +361,7 @@ func (a *ApplicationRestoreController) handle(ctx context.Context, restore *stor
 		}
 	}
 
+	logrus.Infof("StorageClassMapping restore spec storage class mapping is: %v", restore.Spec.StorageClassMapping)
```
vsundarraj-px, line 364: Why is this log added in this pR?  Merge conflicts?

vsundarraj-px (changes requested):

drivers/volume/kdmp/kdmp.go
kgarg-px: Going to change it to, logrus.warnf and thinking of adding cm name as well, would that be better? @vsundarraj-px

drivers/volume/kdmp/kdmp.go
kgarg-px: Yup, sorry about this, I forgot to remove log messages while raising the PR

pkg/applicationmanager/controllers/applicationrestore.go
kgarg-px: There was a case when while restoring, storage class was not being taken in destination cluster by default, had used this log to debug and find out the issue, I forgot to remove the all the log messages that I had used while developing. Will be removing these. Thanks for pointing them out!

vsundarraj-px (approved):
LGTM

vsundarraj-px:
You need resolve PR checks.

prashanthpx:
@kgarg-px Indeed we need not do this way as reconciler code in kdmp does read kdmp-config value (for eg check KopiaResourceRequirements) where we can set kdmp job resoruces on the fly.
We should be following the same with which, we won't need any stork change except vendoring the code to stork

PR #1580, opened 2023-11-28
change stash strategy test name to not include it in migration suite

Rohit-PX:
**What type of PR is this?**
integration-test

* Stash strategy migration test keep failing, until we debug this, renaming it so it does not get run as part of the migration suite
* We will still be running it in a separate job, that way it will be easier to debug.

no

no

no


pp511 (approved):
lgtm

PR #1581, opened 2023-11-29
Aetos integration

Rohit-PX:
**What type of PR is this?**
integration-test

* Add aetos integration to stork integration tests
* That way we can analyze the failures more easily using aetos
* It will also be simpler to track what automation has run and what exactly the tests are doing
* Example: https://aetos.pwx.purestorage.com/resultSet/testSetID/449066 

no

no

no

### DIFF
--- a/test/integration_test/extender_test.go
+++ b/test/integration_test/extender_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/portworx/sched-ops/k8s/storage"
 	"github.com/portworx/torpedo/drivers/node"
 	"github.com/portworx/torpedo/drivers/scheduler"
+	"github.com/portworx/torpedo/pkg/log"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	apps_api "k8s.io/api/apps/v1"
@@ -27,6 +28,7 @@ const (
 )
 
 func TestExtender(t *testing.T) {
+	dash.TestSetBegin(dash.TestSet)
 	err := setSourceKubeConfig()
 	require.NoError(t, err, "failed to set kubeconfig to source cluster: %v", err)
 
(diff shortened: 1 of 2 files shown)


test/integration_test/migration_test.go
```diff
@@ -134,6 +135,7 @@ func triggerMigration(
 	projectIDMappings string,
 	namespaceLabels map[string]string,
 ) ([]*scheduler.Context, *scheduler.Context) {
+	log.InfoD("Schedule mysql app")
```
pp511, line 138: It will not always be mysql app. We should be printing the appKey instead

test/integration_test/migration_test.go
```diff
@@ -144,6 +146,7 @@ func triggerMigration(
 
 	err = schedulerDriver.WaitForRunning(ctxs[0], defaultWaitTimeout, defaultWaitInterval)
 	require.NoError(t, err, "Error waiting for app to get to running state")
+	log.InfoD("Mysql app validated")
```
pp511, line 149: Same here

test/integration_test/migration_test.go
```diff
@@ -269,6 +275,7 @@ func validateAndDestroyMigration(
 }
 
 func deploymentMigrationTest(t *testing.T) {
+	dash.TestCaseBegin("Deployment migration test", "Validating Async DR with Deployment App", "", nil)
```
pp511, line 278: We need to pass testrailID as the third param here right?

test/integration_test/migration_test.go
```diff
@@ -2210,6 +2220,7 @@ func scheduleClusterPairGeneric(t *testing.T, ctxs []*scheduler.Context,
 		err = scheduleUnidirectionalClusterPair(remotePairName, clusterPairNamespace, projectIDMappings, defaultBackupLocation, defaultSecretName, true, pairReverse)
 		require.NoError(t, err, "failed to set unidirectional cluster pair: %v", err)
 	} else {
+		log.InfoD("Scheduling cluster pair using storkctl generate")
```
pp511, line 2223: We should be using storkctl create and not generate. Isn't it?

test/integration_test/extender_test.go
```diff
@@ -509,9 +539,11 @@ func verifyAntihyperconvergence(t *testing.T, appNodes []node.Node, volumes []st
 	for _, appNode := range appNodes {
 		require.Equal(t, highScore, scores[appNode.Name], "Scheduled node does not have the highest score")
 	}
+	log.InfoD("Verified scheduled node has the highest score")
 }
 
 func equalPodSpreadTest(t *testing.T) {
+	dash.TestCaseBegin("Stork scheduler equal pod spread test", "Verify equal pod spread is achieved using stork for an app", "", nil)
```
pp511, line 546: same concern about first param as above. Similar concern everywhere else.

test/integration_test/extender_test.go
```diff
@@ -451,6 +480,7 @@ func antihyperconvergenceTestPreferRemoteOnlyTest(t *testing.T) {
 }
 
 func preferRemoteNodeFalseHyperconvergenceTest(t *testing.T) {
+	dash.TestCaseBegin("Stork scheduler prefer remote node antihyperconvergence test", "validate antihyperconvergence with preferRemoteNodeOnly flag", "", nil)
```
pp511, line 483: Shouldn't we be passing the exact name of the test as the first param instead of a brief description?

PR #1596, opened 2023-12-13
PB-4958: v1

sgajawada-px:
**This is a POC NOT FOR REVIEW**

**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

```release-note
Issue:
User Impact:
Resolution

```

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -163,12 +163,80 @@ func (a *ApplicationRestoreController) verifyNamespaces(restore *storkapi.Applic
 func (a *ApplicationRestoreController) createNamespaces(backup *storkapi.ApplicationBackup,
 	backupLocation string,
 	restore *storkapi.ApplicationRestore) error {
+	logrus.Info("Starting createNamespaces")
 	var namespaces []*v1.Namespace
+
+	// rtTemplate := &storkapi.ResourceTransformation{}
+	// a.client.Get(context.Background(), types.NamespacedName{Name: restore.Spec.ResourceTransformationTemplate, Namespace: kdmputils.AdminNamespace}, rtTemplate)
+
+	var gvkResourceTransform map[schema.GroupVersionKind][]storkapi.TransformSpecs
+	if restore.Spec.ResourceTransformationTemplate != "" {
+		var err error
+		gvkResourceTransform, err = resourcecollector.GetGVKToTransformSpecMapper(restore.Spec.ResourceTransformationTemplate, kdmputils.AdminNamespace)
+		if err != nil {
+			log.ApplicationRestoreLog(restore).
+				Warnf("Unable to get transformation spec from :%s, skipping transformation for this restore, err: %v", restore.Spec.ResourceTransformationTemplate, err)
+			return err
+		}
+	}
+
+	logrus.Info("gvkResourceTransform: ", gvkResourceTransform)
 
 	nsData, err := a.downloadObject(backup, backupLocation, restore.Namespace, nsObjectName, true)
 	if err != nil {
 		return err
 	}
+
+	logrus.Info("Before nsdata: ", string(nsData))
+
+	if gvkResourceTransform != nil {
+		objects := make([]*unstructured.Unstructured, 0)
+		if err = json.Unmarshal(nsData, &namespaces); err != nil {
+			return err
+		}
+
+		for _, ns := range namespaces {
+			var object unstructured.Unstructured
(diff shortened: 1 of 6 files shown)


pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -163,13 +163,81 @@ func (a *ApplicationRestoreController) verifyNamespaces(restore *storkapi.Applic
 func (a *ApplicationRestoreController) createNamespaces(backup *storkapi.ApplicationBackup,
 	backupLocation string,
 	restore *storkapi.ApplicationRestore) error {
+	logrus.Info("Starting createNamespaces")
 	var namespaces []*v1.Namespace
 
+	// rtTemplate := &storkapi.ResourceTransformation{}
+	// a.client.Get(context.Background(), types.NamespacedName{Name: restore.Spec.ResourceTransformationTemplate, Namespace: kdmputils.AdminNamespace}, rtTemplate)
+
+	var gvkResourceTransform map[schema.GroupVersionKind][]storkapi.TransformSpecs
```
lalat-das, line 172: @sgajawada-px  The array nature of TransformSpecs will be effectively utilised only in future wherein we have multiple set of rules for a single type of resource and it is via a two different selector. As of current UX we don't support this via UI.

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -163,13 +163,81 @@ func (a *ApplicationRestoreController) verifyNamespaces(restore *storkapi.Applic
...
 	nsData, err := a.downloadObject(backup, backupLocation, restore.Namespace, nsObjectName, true)
 	if err != nil {
 		return err
 	}
 
+	logrus.Info("Before nsdata: ", string(nsData))
+
+	if gvkResourceTransform != nil {
+		objects := make([]*unstructured.Unstructured, 0)
+		if err = json.Unmarshal(nsData, &namespaces); err != nil {
```
lalat-das, line 194: This sound like a hack because of the way we store namespace resources., We should have fixed while we are backing up namespaces with GVK info in namespace.json. While doing actual implementation lets see if we have to chnage in backup path and how big that challenge would be

pkg/migration/controllers/resourcetransformation.go
```diff
@@ -121,6 +126,9 @@ func (r *ResourceTransformationController) handle(ctx context.Context, transform
 			return nil
```
sgajawada-px, line 126: TODO: During r.validateSpecPath(transform) need to remove the filter getsupportsk8stypes to allow cr and crds in rt only for the restore flow

pkg/resourcecollector/resourcetransformation.go
```diff
@@ -9,10 +10,72 @@ import (
...
+// GetGVKToTransformSpecMapper {GVK: []TransformSpecs{}}
+func GetGVKToTransformSpecMapper(transformName string, namespace string) (map[schema.GroupVersionKind][]stork_api.TransformSpecs, error) {
+	logrus.Info("Starting GetGVKToTransformSpecMapper")
+	gvkResourceTransform := make(map[schema.GroupVersionKind][]stork_api.TransformSpecs)
+	if transformName == "" || namespace == "" {
+		logrus.Errorf("empty name received for resource transformation/namespace")
+		return gvkResourceTransform, fmt.Errorf("empty name received for resource transformation/namespace")
+	}
+
+	resp, err := storkops.Instance().GetResourceTransformation(transformName, namespace)
```
sgajawada-px, line 52: As this is called n times from
verifyNamespace - download namespace.json
VolumeStage - download storageclass.json, resources.json, crd.json
ApplicationStage - download resources.json, crd.json

Is good to have a cache for RT CR object instead of reaching out to apiserver

pkg/applicationmanager/controllers/applicationrestore.go
sgajawada-px: TODO for the actual implementation

pkg/applicationmanager/controllers/applicationrestore.go
sgajawada-px: Yes it's just good to have support from our end.

pkg/resourcecollector/resourcetransformation.go
lalat-das: We need to check also when the RT template reaches its high watermark interms of rules how big this in-memory copy shoots up. Else in case of multiple restore with multiple RT template this in memory size will bloat the whole cluster memory consumption

PR #1602, opened 2024-01-04
PB-4991: Update ApplicationBackupController to execute the post-exec …

sgajawada-px:
Update ApplicationBackupController to execute the post-exec rule if they were not executed due to the backupCR deleteion when it is in-progress.

**RCA**
If the user cancels the in-progress backup(From px-backup we were update AB CR delete Timestamp) which already in the pre-exec rule stage or the volume beginning stage for any reason like backup snapshot ID not found or Err volume provisioner similar cases we were re-reconciling the request.
This leads to deletion of AB CR because the delete Timestamp is not nil which causing the post exec rule execution missing.

**Current Approached Solution in PR**
If the delete timestamp is no nil, we check CR is in the volume stage stage if so we additionally check if the post exec rules were ran, if not we can initiate a same here.
Deletion of the backup from px-backup also removes the rule CR’s which were required for the post exec rule execution. So a PR is raised for PX-backup(https://github.com/portworx/px-backup/pull/1432) to not delete the rule CRs during the backup deletion.
This activity can be taken during the AB CR finialzer execution.

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

Executing the post exec rule during the finalizer cleanup and delete the CR resource once it is completed.

no

```release-note
Issue: Post-exec rule execution was halted due to the cancellation of an in-progress backup initiated by the px-backup.
User Impact: I/O halted on Application or Halt of post-exec rule execution.
Resolution: Executing and removing the post-execution rule Custom Resource (CR) as part of the cleanup process for application backup Custom Resource (CR).

```

Unit Test Results
Test Rail Link: https://portworx.testrail.net/index.php?/runs/view/8098&group_by=cases:section_id&group_id=27135&group_order=asc
Screenshot from 2024-01-04 06-35-21

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -265,6 +265,43 @@ func (a *ApplicationBackupController) createBackupLocationPath(backup *stork_api
 func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_api.ApplicationBackup) error {
 	if backup.DeletionTimestamp != nil {
 		if controllers.ContainsFinalizer(backup, controllers.FinalizerCleanup) {
+			// Run the post exec rules if the backup is in ApplicationBackupStageVolumes stage(After the ApplicationBackupStagePreExecRule Stage) AND execRulesCompleted check is negative
+			if _, ok := a.execRulesCompleted[string(backup.UID)]; !ok && backup.Status.Stage == stork_api.ApplicationBackupStageVolumes {
+				log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Info("BackpCR is marked for deletion before execting the post exec rules, will execute the same during the finalizer cleanup if any")
+				if terminationChannels, ok := a.terminationChannels[string(backup.UID)]; ok {
+					for _, channel := range terminationChannels {
+						log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Info("Sending termination commands to kill pre-exec pod in non-kdmp driver path")
+						channel <- true
+					}
+				}
+				if backup.Spec.PostExecRule != "" {
+					log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Info("Starting post-exec rule during the backup cr finalizer cleanup")
+					err := a.runPostExecRule(backup)
+					if err != nil {
+						message := fmt.Sprintf("Error running PostExecRule during the finalizer cleanup: %v", err)
+						log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Errorf(message)
+						a.recorder.Event(backup,
+							v1.EventTypeWarning,
+							string(stork_api.ApplicationBackupStatusFailed),
+							message)
+						return fmt.Errorf("%v", message)
+					}
+					a.execRulesCompleted[string(backup.UID)] = true
+				}
+			}
+
+			// Delete post-exec rule CR after executing the postexec rules only for the manual backup as it is managed and created through the px-backup
+			// And also here the deletion will be skipped if the backupCR is created by the schedule or for the restore from the px-backup.
+			if len(backup.Spec.PostExecRule) != 0 && backup.Annotations[utils.PxbackupAnnotationCreateByKey] == utils.PxbackupAnnotationCreateByValue && backup.Annotations["portworx.io/backup-by"] != "backup-schedule" && backup.Annotations["portworx.io/backup-by"] != "restore" {
+				log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Info("Delete post-exec rule CR as it is a manual backup created by the px-backup")
+
+				err := storkops.Instance().DeleteRule(backup.Spec.PostExecRule, backup.Namespace)
+				if err != nil && !k8s_errors.IsNotFound(err) {
+					log.ApplicationBackupLog(backup).WithField("Event", "Finalizer Cleanup").Infof("Error while deleting post exec rule CR: %v", err)
+					return err
+				}
+			}
+
 			canDelete, err := a.deleteBackup(backup)
 			if err != nil {
 				logrus.Errorf("%s: cleanup: %s", reflect.TypeOf(a), err)
@@ -272,6 +309,7 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 			if !canDelete {
 				return nil
 			}
+
 			// get-rid of map entry for termination channel and rule flag
 			if _, ok := a.terminationChannels[string(backup.UID)]; ok {
 				delete(a.terminationChannels, string(backup.UID))
@@ -620,8 +658,21 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 		backupStatusVolMap[statusVolume.Namespace+"-"+statusVolume.PersistentVolumeClaim] = ""
 	}
 
+	namespacedName := types.NamespacedName{Namespace: backup.Namespace, Name: backup.Name}
 	backup.Status.Stage = stork_api.ApplicationBackupStageVolumes
-	namespacedName := types.NamespacedName{}
+
+	backup, err = a.updateBackupCRInVolumeStage(
+		namespacedName,
+		stork_api.ApplicationBackupStatusInProgress,
+		backup.Status.Stage,
+		"Starting the Volume backups",
+		nil,
+	)
+	if err != nil {
+		logrus.Errorf("Error while updateBackupCRInVolumeStage: %v", err)
+		return err
+	}
+
 	if a.IsVolsToBeBackedUp(backup) {
 		isResourceTypePVC := IsResourceTypePVC(backup)
 		var objectMap map[stork_api.ObjectInfo]bool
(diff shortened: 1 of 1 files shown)


sgajawada-px:
Need to check the below scenarios.
1. What if the px-backup created the AB CR but if the CR is removed manually.
2. What if the AB CR creation and deletion done manually only depending upon stork.
3. schedule backup delete while it is in-progress

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -262,6 +262,46 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 			if !canDelete {
 				return nil
 			}
+			// run the post exec rules if the backup is in volume stage and execRulesCompleted is not completed
```
siva-portworx, line 265: Should move this step before deleteBackup api? Since we decided to delete the backup, should wait till backup cancel and deletion to run the post exec rule?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -262,6 +262,46 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
...
+							string(stork_api.ApplicationBackupStatusFailed),
+							message)
+						return fmt.Errorf("%v", message)
+					}
+					a.execRulesCompleted[string(backup.UID)] = true
+				}
+			}
+
+			// Delete post-exec rule CR if any
+			if len(backup.Spec.PostExecRule) != 0 && backup.Annotations[utils.PxbackupAnnotationCreateByKey] == utils.PxbackupAnnotationCreateByValue && backup.Annotations["portworx.io/backup-by"] != "backup-schedule" && backup.Annotations["portworx.io/backup-by"] != "restore" {
```
siva-portworx, line 290: Good check. Can we please some comment, why we are having schedule and restore check? So that it will be useful for  others.

Also, We might need to avoid the running the post-Exec rule, if it created  by restore case rt?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -262,6 +262,46 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
...
+				log.ApplicationBackupLog(backup).Info("Delete post-exec rule CR")
+				//uid := backup.GetAnnotations()["portworx.io/backup-uid"]
+				//ruleCrName := getCrNameWithUID(backup.Spec.PostExecRule, uid)
+
+				err := storkops.Instance().DeleteRule(backup.Spec.PostExecRule, backup.Namespace)
+				log.ApplicationBackupLog(backup).Infof("Error value: %v", err)
+
+				if err != nil && !k8s_errors.IsNotFound(err) {
+					log.ApplicationBackupLog(backup).Infof("Error while deleting post exec rule: %v", err)
```
siva-portworx, line 299: Let us have all the error msg and debug in the context to refer to "during cancel backup" or some other statement.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -885,38 +935,41 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
-					string(stork_api.ApplicationBackupStatusFailed),
-					message)
-
-				backup.Status.Stage = stork_api.ApplicationBackupStageFinal
-				backup.Status.FinishTimestamp = metav1.Now()
-				backup.Status.LastUpdateTimestamp = metav1.Now()
-				backup.Status.Status = stork_api.ApplicationBackupStatusFailed
-				backup.Status.Reason = message
-				err = a.client.Update(context.TODO(), backup)
+	if !a.execRulesCompleted[string(backup.UID)] {
```
siva-portworx, line 938: @lalat-das Was there any reason, why we did not had it check in the previous fix? Is it because, it is the last step?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -885,38 +935,41 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+			// Let's kill the pre-exec rule pod here so that application specific
+			// data  stream freezing logic works. Certain app actually unleash the WRITE when session ends.
+			// At this point we are dead sure that volume snapshot for all PVCs in the APP is done.. if not then
+			// there is issue...
+			// For detail refer pb-3823
+			for _, channel := range terminationChannels {
+				logrus.Infof("Sending termination commands to kill pre-exec pod in kdmp or mixed driver path")
+				channel <- true
+			}
+			//terminationChannels = nil
```
siva-portworx, line 949: Remove the commented code.

siva-portworx:
Since it is a change in the Rule, I would suggest to have functional test case define to cover all the rule cases.

pkg/applicationmanager/controllers/applicationbackup.go
sgajawada-px: Agree, post exec rule execution doesn't required to be on hold till the backup deletion will update accordingly.

pkg/applicationmanager/controllers/applicationbackup.go
sgajawada-px: 1. Sure will update the comments.
2. Also, We might need to avoid the running the post-Exec rule, if it created by restore case rt? - Yes, existing check will prevent this (**backup.Annotations["portworx.io/backup-by"] != "restore"**) is there anything else am missing ?

pkg/applicationmanager/controllers/applicationbackup.go
sgajawada-px: Sure

sgajawada-px: This commented code is an existing one not part of the this PR change.

sgajawada-px:
> Since it is a change in the Rule, I would suggest to have functional test case define to cover all the rule cases.

I assume functional testcases from the px-backup.
1. Need to check the existing test cases which can break due to this change.
2. Add the new testcases to test the current fix.
@siva-portworx is my understanding correct ??

siva-portworx:
> > Since it is a change in the Rule, I would suggest to have functional test case define to cover all the rule cases.
> 
> I assume functional testcases from the px-backup.
> 
> 1. Need to check the existing test cases which can break due to this change.
> 2. Add the new testcases to test the current fix.
>    @siva-portworx is my understanding correct ??

I want to have test case covering all the rule scenario. Because we had some regression in the rule fixed perviously.

sgajawada-px:
Unit Testcases for the Fix are document in below sheet
https://docs.google.com/spreadsheets/d/10rtChd5sHgvzNsqvh8tUHHRAn86nZHV4UIbVlyTxAJM/edit?usp=sharing

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Ok, I meant of avoiding the running post-exec rule if the applicationbackup CR was created by restore operation.
But I think the check in the line 254 might not satisfy for restore's applicationbakup CR. So it should be fine.

siva-portworx: One concern is that when the application backup CR is deleted, px-backup will also delete the Rule CR as well rt?
Will it not cause a issue that Rule CR might be available at line 264?

sgajawada-px: A PR is also raised for the same https://github.com/portworx/px-backup/pull/1432

siva-portworx (approved):
LGTM

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -265,13 +265,52 @@ func (a *ApplicationBackupController) createBackupLocationPath(backup *stork_api
...
+							message)
+						return fmt.Errorf("%v", message)
+					}
+					a.execRulesCompleted[string(backup.UID)] = true
+				}
+			}
+
+			// Delete post-exec rule CR after executing the postexec rules only for the manual backup as it is managed and created through the px-backup
+			// And also here the deletion will be skipped if the backupCR is created by the schedule or for the restore from the px-backup.
+			if len(backup.Spec.PostExecRule) != 0 && backup.Annotations[utils.PxbackupAnnotationCreateByKey] == utils.PxbackupAnnotationCreateByValue && backup.Annotations["portworx.io/backup-by"] != "backup-schedule" && backup.Annotations["portworx.io/backup-by"] != "restore" {
```
vsundarraj-px, line 295: can you move this in cleanupResources() to handle all Cr cleanup.
Also for kubevirt support we have added support do delete the ruleCR created by stork. You may want to check if there is no conflict.

vsundarraj-px: You already mentioned px-backup will delete the ruleCr so there will be conflict stork deletes the Cr too and pxBackup deletes too. There should be a way to know that delete is triggered by px-backup. In such case, we should avoid deleting the Cr. In my opinion we should defer deleting ruleCr at all unless its created by stork. 
If the applicationBackupCr created by Px-backup but user has issued deleted outside of px-backup, then we should leave the user to delete the Cr as well.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -309,6 +354,12 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 	if backup.Status.Stage == stork_api.ApplicationBackupStageFinal {
 		return nil
 	}
+
+	// Initialize execRulesCompleted for the backup UID
```
vsundarraj-px, line 358: This is redundant code. You already checking ifExists..

pkg/applicationmanager/controllers/applicationbackup.go
vsundarraj-px: this was the comment I had given earlier, guess it was missed incorporating.

vsundarraj-px (changes requested):

pkg/applicationmanager/controllers/applicationbackup.go
sgajawada-px: That is in finalizer here it is during initialization.

PR #1603, opened 2024-01-08
[PWX-35512] Preventing pods from being assigned on nodes not available to Volume Driver

olavangad-px:
**What type of PR is this?**
>bug
>unit-test

As discussed in [PWX-35512](https://portworx.atlassian.net/browse/PWX-35512), if for some reason a node received as part of the Prioritize request is not available to the driver, we should avoid scheduling pods on that node. This is achieved by assigning the node a 0 score so that this node is ignored by the k8s scheduler during scheduling.

No

Maybe

Yes, 23.11

**Testing Details**
I have added unit tests for both hyperconverged and anti-hyperconverged cases.
The QA can replicate the original issue to ensure that the proposed fix is working as expected.

[PWX-35512]: https://portworx.atlassian.net/browse/PWX-35512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -698,13 +698,13 @@ func (e *Extender) processPrioritizeRequest(w http.ResponseWriter, req *http.Req
 				storklog.PodLog(pod).Debugf("Volume %v allocated in regions: %v", volume.VolumeName, regionInfo.PreferredLocality)
 
 				for k8sNodeIndex, node := range args.Nodes.Items {
-					if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; !keyExists {
+					if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; keyExists {
+						priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
+					} else {
 						// the k8sNodeIndex's corresponding node isn't available to the driver,
 						// so we shouldn't schedule pods on this node
 						// we will assign a negative score here and set it to 0 when encoding the response
 						priorityMap[node.Name] = -1
-					} else {
-						priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
 					}
 				}
 			}
@@ -769,12 +769,7 @@ func (e *Extender) updateForAntiHyperconvergence(
 		}
 	}
 	for k8sNodeIndex, node := range args.Nodes.Items {
-		if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; !keyExists {
-			// the k8sNodeIndex's corresponding node isn't available to the driver,
-			// so we shouldn't schedule pods on this node
-			// we will assign a negative score here and set it to 0 when encoding the response
-			priorityMap[node.Name] = -1
-		} else {
+		if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; keyExists {
 			// storageNode.StorageID = datanodeID
 			// Give defaultScore to the nodes where NeedsAntiHyperconvergence volume exist
 			// to give them a lower score
@@ -786,6 +781,8 @@ func (e *Extender) updateForAntiHyperconvergence(
 				// a score higher than the default score
 				priorityMap[node.Name] += int(nodePriorityScore)
 			}
+		} else {
+			priorityMap[node.Name] = -1
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/extender/extender.go
```diff
@@ -696,8 +698,14 @@ func (e *Extender) processPrioritizeRequest(w http.ResponseWriter, req *http.Req
...
-					storageNode := k8sNodeIndexStorageNodeMap[k8sNodeIndex]
-					priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
+					if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; !keyExists {
+						// the k8sNodeIndex's corresponding node isn't available to the driver,
+						// so we shouldn't schedule pods on this node
+						// we will assign a negative score here and set it to 0 when encoding the response
+						priorityMap[node.Name] = -1
+					} else {
+						priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
+					}
```
pp511, line 708: nit.. This does work. However I think it would have been cleaner if the "if" and "else" were swapped.
 What I mean is this
```
if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; keyExists {
  priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
}
else{
  priorityMap[node.Name] = -1
}
```

pkg/extender/extender.go
```diff
@@ -759,17 +769,23 @@ func (e *Extender) updateForAntiHyperconvergence(
...
-		// Give defaultScore to the nodes where NeedsAntiHyperconvergence volume exist
-		// to give them a lower score
-		if val, ok := needsAntiHyperconvergenceReplicaNodes[storageNode.StorageID]; ok && val {
-			priorityMap[node.Name] = int(defaultScore)
-		} else if storageNode.Status == volume.NodeOnline {
-			// In a scenario where regular volumes do not exist
-			// Raise the score of non replica nodes to give them
-			// a score higher than the default score
-			priorityMap[node.Name] += int(nodePriorityScore)
+		if storageNode, keyExists := k8sNodeIndexStorageNodeMap[k8sNodeIndex]; !keyExists {
```
pp511, line 772: nit.. Similar suggestion as above

pkg/extender/extender.go
```diff
@@ -715,6 +723,8 @@ sendResponse:
 		score, ok := priorityMap[node.Name]
 		if !ok || score == 0 {
 			score = int(defaultScore)
+		} else if score == -1 {
+			score = int(invalidNodeScore)
```
pp511, line 727: nit..We assign a score of 0 to offline nodes and later set it to defaultScore (5).
We assign -1 to invalid nodes and later set it to invalidNodeScore (0).

Instead shouldn't we just update a score of 5 to offline nodes and 0 to invalid nodes and just use those values from the priorityMap here? I think it would be just a bit cleaner. Wdyt?

pp511 (commented):
Changes look good. Just some nits to make the code cleaner.

pkg/extender/extender.go
olavangad-px: Sure will address

pkg/extender/extender.go
olavangad-px: Will address

pkg/extender/extender.go
olavangad-px: In the current implementation priorityMap is initialized with all values set to 0.
What you are saying makes sense however if we initialise the map with score of 5, we will have to make changes in unit tests and integration tests wherever required (since node which earlier had a score of 2*nodePriorityScore will have a score of 2*nodePriorityScore+defaultScore now and so on)

Since we want to get this change in on a tight deadline I feel it makes more sense to address this later

olavangad-px: Makes sense.
But this won't necessarily make the code cleaner.
1. The below check will still have to remain 
```
if !ok {
score = int(defaultScore)
}
```

2. In line # 595 , we will have to handle updating the 0 score (assigned to every node when priorityMap was initialized) to defaultScore here separately.
```
if disableHyperconvergence {
		goto sendResponse
	}
```

3. Also making that change will involve calculating and changing unit tests and integration tests as well since the final scores for nodes will change a bit.

For example, 
```
priorityMap[node.Name] += int(e.getNodeScore(node, volume, &rackInfo, &zoneInfo, &regionInfo, storageNode))
```
If we return defaultScore instead of 0 from e.getNodeScore(), then the final scores will vary based on the setup because we are adding the scores.

pp511: Yeah. Let's not complicate things for now.

pp511 (approved):
lgtm

PR #1604, opened 2024-01-08
PTX-21639: Namespace delete error should not error out if namespace does not exists in migration tests cleanup.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
>integration-test

Namespace delete error should not error out if namespace does not exists in migration tests cleanup.

**Test**
https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2023.11-dev/job/migration-op-k8s-1-25-0/54/

### DIFF
--- a/test/integration_test/migration_test.go
+++ b/test/integration_test/migration_test.go
@@ -2533,7 +2533,9 @@ func blowNamespacesForTest(t *testing.T, instanceID, appKey string, skipDestDele
 
 	namespace := fmt.Sprintf("%s-%s", appKey, instanceID)
 	err = core.Instance().DeleteNamespace(namespace)
-	require.NoError(t, err, "failed to delete namespace %s on destination cluster", namespace)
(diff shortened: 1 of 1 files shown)


pp511 (approved):
Looks ok. But not able to understand when/why we run into an issue like this? Is the test not creating a ns and we are still trying to delete it? In that case shouldn't the test be fixed?

diptiranjanpx:
> Looks ok. But not able to understand when/why we run into an issue like this? Is the test not creating a ns and we are still trying to delete it? In that case shouldn't the test be fixed?

The ns in dest sides are created by bidirectional cluster pair creation. At each test level, making the test aware about if the clusterpair creation is bidirectional or unidirectional or old way of applying yaml and then taking the decision of deleting the namespace would have been too much change and function definition changes. This is the reason for going with the above change.

PR #1607, opened 2024-01-12
PB-5270 PB-5271 : fail backup and restore if admin namespace is not present

strivedi-px:
**What type of PR is this?**
bug-fix

If the custom admin namespace is set and not present, backup as well as restore should fail. However, currently they work by creating the set admin namespace themselves and proceeding ahead. However, the admin namespace related operations should ideally be handled by the user and not through automation. This pull-request errors out the backup and restore operations if the set admin namespace is not found. Once the user creates the admin namespace manually again, the backup and restore should work as expected.

No

```release-note
Issue:
User Impact:
Resolution

```


strivedi-px:
Looks like this PR would not be needed to fix this issue. Have created a PR in px-backup that should take care. https://github.com/portworx/px-backup/pull/1451

PR #1614, opened 2024-01-30
feat: removed creation of default vsc from stork

dbinnal-px:
For KDMP clean up we are removing creation of default VolumeSnapshotClass from stork. As we don’t know whether that particular csi driver has a capability of volumeSnapshot or not
User has to create a VolumeSnapshotClass if required by them.

**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

- Removes creation of default VolumeSnapshotClass from stork

- Before updating the stork with latest changes, stork creating default VSC
Screenshot 2024-01-30 at 4 18 24 PM

- Deleted stork created VSC class, and restarted the stork with latest changes, stork did not create VSC by default.
Screenshot 2024-01-30 at 4 16 54 PM

### DIFF
--- a/drivers/volume/csi/csi.go
+++ b/drivers/volume/csi/csi.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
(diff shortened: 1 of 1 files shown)


siva-portworx:
- Please test results to the PR. 
- Please validate the default stork VSC are not getting created. 
- Also if the some stork VSC are already present, restarting of the stork should not delete them.

siva-portworx (approved):
Please add the test result. Rest looks good to me.

pallav-px (approved):

PR #1631, opened 2024-02-12
[PWX-32202] storkctl perform failover CLI changes

olavangad-px:
**What type of PR is this?**
>feature
>unit-test

To add storkctl perform failover cli command. 
This command creates an action CR with actionType:failover for which we will add logic in the ActionController to perform the actual failover.

Another change is that we will create static copy of migrationSchedules in remote cluster irrespective of autoSuspend value

Yes, it adds a new storkctl command `storkctl perform failover`
```
[root@olavangad-84-0 ~]# storkctl perform failover -h
Initiate failover of the given migration schedule

Usage:
  storkctl perform failover [flags]

Flags:
      --exclude-namespaces strings   Specify the comma-separated list of subset namespaces to be skipped during the failover. By default, all namespaces part of the migrationSchedule are failed over
  -h, --help                         help for failover
      --include-namespaces strings   Specify the comma-separated list of subset namespaces to be failed over. By default, all namespaces part of the migrationSchedule are failed over
  -m, --migration-reference string   Specify the migration schedule to failover. Also specify the namespace of this migrationSchedule using the -n flag
      --skip-deactivate-source       If present, applications in the source cluster will not be scaled down as part of the failover.

Global Flags:
      --alsologtostderr                   log to standard error as well as files
      --burst int                         restrict number of k8s API requests from stork (default 2000)
      --context string                    name of the kubeconfig context to use
      --kubeconfig string                 path to the kubeconfig file to use for CLI requests
      --log_backtrace_at traceLocations   when logging hits line file:N, emit a stack trace
      --log_dir string                    If non-empty, write log files in this directory
      --log_link string                   If non-empty, add symbolic links in this directory to the log files
      --logbuflevel int                   Buffer log messages logged at this level or lower (-1 means don't buffer; 0 means buffer INFO only; ...). Has limited applicability on non-prod platforms.
      --logtostderr                       log to standard error instead of files
  -n, --namespace string                  If present, the namespace scope for this CLI request (default "default")
  -o, --output string                     output format. One of: table|json|yaml (default "table")
      --qps int                           restrict number of k8s API requests from stork (default 1000)
      --stderrthreshold severityFlag      logs at or above this threshold go to stderr (default 2)
  -v, --v Level                           log level for V logs
      --vmodule vModuleFlag               comma-separated list of pattern=N settings for file-filtered logging
  -w, --watch                             watch stork resources

```

Yes, TBD

Yes, 24.2.0

**Testing Details**:
1. Have tested that remote copy of migrationSchedule is created irrespective of autoSuspend value.
2. Have added comprehensive unit tests for storkctl perform failover and tested the corner cases manually as well.

### DIFF
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -181,6 +181,17 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 		} else if err != nil {
 			return err
 		}
+		if remoteMigrSched.GetAnnotations() == nil || remoteMigrSched.GetAnnotations()[StorkMigrationScheduleCopied] != "true" {
+			// handling the corner case where user has already created a migrationSchedule with the same name
+			// and namespace in remote cluster before upgrading to stork 24.2.0
+			// we will fail the migrationSchedule in this case
+			err = fmt.Errorf("a non static-copy migrationSchedule with the same name and namespace found in the remote cluster")
+			m.recorder.Event(migrationSchedule,
+				v1.EventTypeWarning,
+				string(stork_api.MigrationStatusFailed),
+				err.Error())
+			return err
+		}
 		if migrationSchedule.Spec.AutoSuspend && remoteMigrSched.Status.ApplicationActivated {
 			suspend := true
 			migrationSchedule.Spec.Suspend = &suspend
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -163,21 +163,20 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 			if err != nil && !errors.IsAlreadyExists(err) {
 				return err
 			}
-			// create new migrationschedule on remote cluster
-			remoteMigrSched = migrationSchedule.DeepCopy()
-			remoteMigrSched.ResourceVersion = ""
-			remoteMigrSched.UID = ""
-			if remoteMigrSched.Annotations == nil {
-				remoteMigrSched.Annotations = make(map[string]string)
+			// create new migrationSchedule on remote cluster
+			var remoteMigrSchedObj = migrationSchedule.DeepCopy()
+			remoteMigrSchedObj.ResourceVersion = ""
+			remoteMigrSchedObj.UID = ""
+			if remoteMigrSchedObj.Annotations == nil {
+				remoteMigrSchedObj.Annotations = make(map[string]string)
 			}
-			remoteMigrSched.Annotations[StorkMigrationScheduleCopied] = "true"
+			remoteMigrSchedObj.Annotations[StorkMigrationScheduleCopied] = "true"
 			suspend := true
-			remoteMigrSched.Spec.Suspend = &suspend
-			remoteMigrSched.Status = stork_api.MigrationScheduleStatus{}
-			if _, err := remoteOps.CreateMigrationSchedule(remoteMigrSched); err != nil {
+			remoteMigrSchedObj.Spec.Suspend = &suspend
+			remoteMigrSchedObj.Status = stork_api.MigrationScheduleStatus{}
+			if remoteMigrSched, err = remoteOps.CreateMigrationSchedule(remoteMigrSchedObj); err != nil {
 				return err
 			}
-
 		} else if err != nil {
 			return err
 		}
@@ -201,8 +200,6 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 				"Suspended",
 				msg)
 			log.MigrationScheduleLog(migrationSchedule).Warn(msg)
-			// deactivate apps in namespace
-			// TODO: suspend deploy/sts,crds
 			return m.client.Update(context.TODO(), migrationSchedule)
 		}
 	}
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -149,50 +149,50 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 				err.Error())
 			return nil
 		}
-		if migrationSchedule.Spec.AutoSuspend {
-			var remoteMigrSched *stork_api.MigrationSchedule
-			remoteMigrSched, err = remoteOps.GetMigrationSchedule(migrationSchedule.Name, migrationSchedule.Namespace)
-			if errors.IsNotFound(err) {
-				namespace, err := core.Instance().GetNamespace(migrationSchedule.Namespace)
-				if err != nil {
-					return err
-				}
-				namespace.ResourceVersion = ""
-				_, err = coreOps.CreateNamespace(namespace)
-				if err != nil && !errors.IsAlreadyExists(err) {
-					return err
-				}
-				// create new migrationschedule on remote cluster
-				remoteMigrSched = migrationSchedule.DeepCopy()
-				remoteMigrSched.ResourceVersion = ""
-				remoteMigrSched.UID = ""
-				if remoteMigrSched.Annotations == nil {
-					remoteMigrSched.Annotations = make(map[string]string)
-				}
-				remoteMigrSched.Annotations[StorkMigrationScheduleCopied] = "true"
-				suspend := true
-				remoteMigrSched.Spec.Suspend = &suspend
-				remoteMigrSched.Status = stork_api.MigrationScheduleStatus{}
-				if _, err := remoteOps.CreateMigrationSchedule(remoteMigrSched); err != nil {
-					return err
-				}
 
-			} else if err != nil {
+		// create migrationSchedule in remote cluster irrespective of autoSuspend value
+		var remoteMigrSched *stork_api.MigrationSchedule
+		remoteMigrSched, err = remoteOps.GetMigrationSchedule(migrationSchedule.Name, migrationSchedule.Namespace)
+		if errors.IsNotFound(err) {
+			namespace, err := core.Instance().GetNamespace(migrationSchedule.Namespace)
+			if err != nil {
 				return err
 			}
-			if remoteMigrSched.Status.ApplicationActivated {
-				suspend := true
-				migrationSchedule.Spec.Suspend = &suspend
-				msg := "Suspending migration schedule since migrated apps on remote cluster are active"
-				m.recorder.Event(migrationSchedule,
-					v1.EventTypeWarning,
-					"Suspended",
-					msg)
-				log.MigrationScheduleLog(migrationSchedule).Warn(msg)
-				// deactivate apps in namespace
-				// TODO: suspend deploy/sts,crds
-				return m.client.Update(context.TODO(), migrationSchedule)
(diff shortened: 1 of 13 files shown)


pkg/migration/controllers/migrationschedule.go
```diff
@@ -149,50 +149,61 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
...
+			remoteMigrSched.ResourceVersion = ""
+			remoteMigrSched.UID = ""
+			if remoteMigrSched.Annotations == nil {
+				remoteMigrSched.Annotations = make(map[string]string)
+			}
+			remoteMigrSched.Annotations[StorkMigrationScheduleCopied] = "true"
+			suspend := true
+			remoteMigrSched.Spec.Suspend = &suspend
+			remoteMigrSched.Status = stork_api.MigrationScheduleStatus{}
+			if _, err := remoteOps.CreateMigrationSchedule(remoteMigrSched); err != nil {
```
diptiranjanpx, line 177: Only return of error is not  AlreadyExists.
Also use the migration schedule copy returned by the server from there onwards instead of `remoteMigrSched`.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -149,50 +149,61 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
...
+			suspend := true
+			migrationSchedule.Spec.Suspend = &suspend
+			msg := "Suspending migration schedule since migrated apps on remote cluster are active"
+			m.recorder.Event(migrationSchedule,
+				v1.EventTypeWarning,
+				"Suspended",
+				msg)
+			log.MigrationScheduleLog(migrationSchedule).Warn(msg)
+			// deactivate apps in namespace
+			// TODO: suspend deploy/sts,crds
```
diptiranjanpx, line 205: nit: Let's remove this line.

pkg/storkctl/action.go
```diff
@@ -20,99 +20,144 @@ const (
...
-	actionList, err := storkops.Instance().ListActions(namespace)
-	if err != nil {
-		util.CheckErr(err)
-		return nil
+// given the name, namespace, actionType and actionParameters create an Action CR
+func createActionCR(actionName string, namespace string, actionType storkv1.ActionType, actionParameters storkv1.ActionParameter) (*storkv1.Action, error) {
+	action := storkv1.Action{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      actionName,
+			Namespace: namespace,
```
diptiranjanpx, line 142: Should we put an annotation "created-by: stork" to figure out if it has been created by storkctl ?

pkg/storkctl/action.go
```diff
@@ -20,99 +20,144 @@ const (
...
-	return strings.Join([]string{string(action), time.Now().Format(nameTimeSuffixFormat)}, "-")
+func getActionStatusMessage(action *storkv1.Action) string {
+	return fmt.Sprintf("To check %v status use the command : `storkctl get %v %v -n %v`", action.Spec.ActionType, action.Spec.ActionType, action.Name, action.Namespace)
 }
 
-func getDescribeActionMessage(action *storkv1.Action) string {
-	return "To check Action status use: " +
-		fmt.Sprintf("kubectl describe action %v -n %v", action.Name, action.Namespace)
+func getActionName(actionType storkv1.ActionType, referenceResourceName string) string {
+	return strings.Join([]string{string(actionType), referenceResourceName, GetCurrentTime().Format(nameTimeSuffixFormat)}, "-")
```
diptiranjanpx, line 162: let's trim <actionType>-<referenceResourceName> so that the whole string including the time fits into the max chars. Also let's put these kind of functions in utils.

pkg/migration/controllers/migrationschedule.go
olavangad-px: You mean the TODO line?

pkg/migration/controllers/migrationschedule.go
olavangad-px: This particular branch will be entered only if remote migration schedule doesn't already exist. 
So err caught here cannot be AlreadyExists.

2nd point makes sense. Will address.

pkg/storkctl/action.go
olavangad-px: Since we are not adding such an annotation in other migration objects clusterpair, migrationschedule 
I think we can skip this for now.
If we decide in the future we can make the change across all objects.

pkg/storkctl/action.go
olavangad-px: I will trim it according to 253 chars

diptiranjanpx (approved):

pkg/storkctl/action_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+package storkctl
+
+import (
+	"fmt"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"testing"
+	"time"
```
pp511, line 12: nit..Let's group two groups here too.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -149,50 +149,58 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
...
-				remoteMigrSched.Annotations[StorkMigrationScheduleCopied] = "true"
-				suspend := true
-				remoteMigrSched.Spec.Suspend = &suspend
-				remoteMigrSched.Status = stork_api.MigrationScheduleStatus{}
-				if _, err := remoteOps.CreateMigrationSchedule(remoteMigrSched); err != nil {
-					return err
-				}
 
-			} else if err != nil {
+		// create migrationSchedule in remote cluster irrespective of autoSuspend value
```
pp511, line 153: I am missing the point. Help me understand why we need to copy the migrationschedules?

pp511 (approved):
Looks good. Just one question and some nit comments.

pkg/storkctl/action.go
```diff
@@ -2,6 +2,7 @@ package storkctl
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/validation"
```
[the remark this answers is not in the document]
olavangad-px: will address

pkg/storkctl/action.go
```diff
@@ -20,99 +21,150 @@ const (
...
+			if len(referenceMigrationSchedule) == 0 {
+				util.CheckErr(fmt.Errorf("referenceMigrationSchedule name needs to be provided for failover"))
+				return
+			}
+			// namespace of the migrationSchedule is provided by user using the -n / --namespace global flag
+			migrationScheduleNs := cmdFactory.GetNamespace()
+			migrSchedObj, err := storkops.Instance().GetMigrationSchedule(referenceMigrationSchedule, migrationScheduleNs)
 			if err != nil {
-				util.CheckErr(err)
+				util.CheckErr(fmt.Errorf("unable to find the referenceMigrationSchedule %v in the %v namespace", referenceMigrationSchedule, migrationScheduleNs))
```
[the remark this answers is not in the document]
olavangad-px: makes sense. will address

pkg/storkctl/action.go
```diff
@@ -20,99 +21,150 @@ const (
...
+			// at most one of exclude-namespaces or include-namespaces can be provided
+			if len(includeNamespaceList) != 0 && len(excludeNamespaceList) != 0 {
+				util.CheckErr(fmt.Errorf("can provide only one of --include-namespaces or --exclude-namespaces values at once"))
+				return
+			} else if len(includeNamespaceList) != 0 {
+				if isSubset, nonSubsetStrings := isSubset(includeNamespaceList, migrationNamespaces); isSubset {
+					// Branch 1: Only failover some of the namespaces being migrated by the given migrationSchedule
+					namespaceList = includeNamespaceList
+				} else {
+					util.CheckErr(fmt.Errorf("provided namespaces %v are not a subset of the namespaces being migrated by the given migrationSchedule", nonSubsetStrings))
```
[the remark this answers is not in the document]
olavangad-px: okay will address

pkg/storkctl/action.go
```diff
@@ -20,99 +21,150 @@ const (
...
+			printMsg(getActionStatusMessage(action), ioStreams.Out)
 		},
 	}
-	return failoverCommand
-}
-
-func isActionIncomplete(action *storkv1.Action) bool {
-	return action.Status.Status == storkv1.ActionStatusScheduled || action.Status.Status == storkv1.ActionStatusInProgress
+	performFailoverCommand.Flags().BoolVar(&skipDeactivateSource, "skip-deactivate-source", false, "If present, applications in the source cluster will not be scaled down as part of the failover.")
+	performFailoverCommand.Flags().StringVarP(&referenceMigrationSchedule, "migration-reference", "m", "", "Specify the migration schedule to failover. Also specify the namespace of this migrationSchedule using the -n flag")
```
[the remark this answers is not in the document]
olavangad-px: yes, will go through the entire PR once again and make sure that i use either one of migration schedule or Migration Schedule everywhere

pkg/storkctl/action_test.go
olavangad-px: will review in all the files and make sure this is taken care of

pkg/migration/controllers/migrationschedule.go
olavangad-px: perform failover command is run in the context of destination cluster with the flag --migration-reference <migrationScheduleName>
And the migrationSchedule is then used to get the clusterPairName and create last mile migration object. 
Obviously for this to work, the reference migrationschedule should exist in the destination cluster.

As of stork 23.11, migrationschedule's staticCopy is created in the remote cluster only in case the original source migrationSchedule has autoSuspend set to true.
From stork 24.2 onwards, we will always create the staticCopy in the remote cluster.

pkg/storkctl/action.go
```diff
@@ -20,99 +21,150 @@ const (
...
+					return
 				}
 			}
-			if failedToStart {
+			migrationNamespaceList := migrSchedObj.Spec.Template.Spec.Namespaces
+			migrationNamespaceSelectors := migrSchedObj.Spec.Template.Spec.NamespaceSelectors
+			// update the migrationNamespaces list by fetching namespaces based on provided label selectors
+			migrationNamespaces, err := getMigrationNamespaces(migrationNamespaceList, migrationNamespaceSelectors)
+			if err != nil {
+				util.CheckErr(fmt.Errorf("unable to get the namespaces based on the --namespace-selectors in the provided migrationSchedule: %v", err))
```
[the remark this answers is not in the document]
olavangad-px: The helper function being called here, `getMigrationNamespaces` returns the migrationNamespaceList itself in case selectors is empty. So no possibility of an error in that scenario.
The function's responsibility is to append the namespaces based on the selectors to the provided list of namespaces. So errors can only arise if we are unable to get the namespaces based on the selectors.
```
namespaces, err := core.Instance().ListNamespaces(map[string]string{key: val})
```
 That's why the error message conveys the same.

PR #1637, opened 2024-02-15
set default version in Makefile

Rohit-PX:
**What type of PR is this?**
improvement

Modify the default version of stork in Makefile so that build pipelines can pick it up easily.

no

no

24.2.0


mborodin-px (approved):
LGTM

PR #1638, opened 2024-02-17
PTX-21395 Added jenkins build pipeline

mborodin-px:
It adds a new build Jenkins pipeline which is going to be triggered for all release branches

no

no

yes


Rohit-PX (approved):
LGTM.

Makefile
```diff
@@ -44,7 +44,7 @@ ifeq ($(BUILD_TYPE),debug)
 BUILDFLAGS += -gcflags "-N -l"
 endif
 
-RELEASE_VER := 2.7.0
+RELEASE_VER := 99.9.9
```
adityadani, line 47: This should be an actual version? I would say lets keep this as 24.0.0, and every year we bump up the first number. @diptiranjanpx wdyt?

adityadani (approved):

Makefile
mborodin-px: @adityadani for all branches it should be `99.9.9` in my release pipelines I override this value to whatever version taken from branch name, for ex. `release-24.1.0` it will pass `RELEASE_VER=24.1.0`, so you don't need to commit this version every time in the release branches.

PR #1639, opened 2024-02-18
PWX-33044: add kubevirt label to the PVCs added directly to the VMs

pureneelesh:
**What type of PR is this?**
integration-test

PX performs its KubeVirt-specific logic only if the volume is a KubeVirt volume based on the volume labels. When the volume is created via a dataVolume template such labels are automatically added to the PVC. However, if a PVC is added directly to the VM spec, no such labels are present and therefore no VPS is added to the PX volume.

This patch adds a label with "kubevirt" in key to the PVCs specified directly in the VM spec.

Note that the test is currently passing even without this label. That is because the namespace just happens to have "kubevirt" in the name and the namespace appears as one of the label values. If the same VM spec is used from a different namespace without "kubevirt" in the name, the VPS detection logic is skipped and replica collaction does not work.

no

no

no


Rohit-PX (approved):
LGTM.

PR #1648, opened 2024-02-22
Scale kubevirt

Rohit-PX:
**What type of PR is this?**
integration-test

* Add scaling for existing tests with respect to kubevirt VMs being deployed
* Create `ClusterRole` and `ClusterRoleBinding` dynamically for datavolume permissions

no

no 

no

### DIFF
--- a/test/integration_test/kubevirt_test.go
+++ b/test/integration_test/kubevirt_test.go
@@ -5,16 +5,22 @@ package integrationtest
 
 import (
 	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
 	"github.com/portworx/sched-ops/k8s/core"
-	"github.com/portworx/sched-ops/k8s/kubevirt"
+	kubevirt "github.com/portworx/sched-ops/k8s/kubevirt"
+	"github.com/portworx/sched-ops/k8s/rbac"
 	"github.com/portworx/sched-ops/task"
 	"github.com/portworx/torpedo/drivers/scheduler"
 	"github.com/portworx/torpedo/pkg/log"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var templatePVCSpecs = map[string]string{
@@ -34,7 +40,10 @@ const (
 	kubevirtCDIStorageConditionAnnotation = "cdi.kubevirt.io/storage.condition.running.reason"
 	kubevirtCDIStoragePodPhaseAnnotation  = "cdi.kubevirt.io/storage.pod.phase"
 
-	volumeBindingImmediate = "kubevirt-templates"
+	volumeBindingImmediate       = "kubevirt-templates"
+	dataVolumeClusterrole        = "datavol-clusterrole"
+	dataVolumeClusterroleBinding = "datavol-clusterrolebinding"
+	dataDiskClusterroleBinding   = "datadisk-clusterrolebinding"
 )
 
 func TestKubevirt(t *testing.T) {
@@ -48,6 +57,13 @@ func TestKubevirt(t *testing.T) {
 
 	err = createDatadiskTemplates(t)
 	require.NoError(t, err, "Error creating kubevirt templates")
+
+	err = createClusterroleDataVolume(dataVolumeClusterrole)
+	require.NoError(t, err, "Error creating cluste role for datavolume clone")
+
+	err = createClusterroleBindingDatavolume(dataVolumeClusterrole, kubevirtDatadiskNamespace)
+	require.NoError(t, err, "Error creating/updating data volume clusterrole binding %s with namespace: %s",
+		dataVolumeClusterroleBinding, kubevirtDatadiskNamespace)
 
 	t.Run("kubevirtDeployFedoraVMWithClonePVC", kubevirtDeployFedoraVMWithClonePVC)
 	t.Run("kubevirtDeployWindowsServerWithClonePVC", kubevirtDeployWindowsServerWithClonePVC)
@@ -66,16 +82,12 @@ func kubevirtDeployFedoraVMWithClonePVC(t *testing.T) {
 	defer updateTestRail(&testResult, testrailID, runID)
 	defer updateDashStats(t.Name(), &testResult)
 	instanceID := "vm"
-	appKey := "kubevirt-fedora"
+	appKeys := []string{"kubevirt-fedora"}
 
-	ctxs := kubevirtVMsDeployAndValidate(
-		t,
-		instanceID,
-		[]string{appKey},
-	)
+	allCtxs := kubevirtVMScaledDeployAndValidate(t, instanceID, appKeys, kubevirtScale)
 
 	log.Infof("Destroying apps")
-	destroyAndWait(t, ctxs)
+	destroyAndWait(t, allCtxs)
 
 	// If we are here then the test has passed
 	testResult = testResultPass
@@ -88,49 +100,16 @@ func kubevirtDeployWindowsServerWithClonePVC(t *testing.T) {
 	defer updateTestRail(&testResult, testrailID, runID)
 	defer updateDashStats(t.Name(), &testResult)
 	instanceID := "vm"
-	appKey := "kubevirt-windows-22k-server"
+	appKeys := []string{"kubevirt-windows-22k-server"}
 
-	ctxs := kubevirtVMsDeployAndValidate(
-		t,
-		instanceID,
-		[]string{appKey},
-	)
+	allCtxs := kubevirtVMScaledDeployAndValidate(t, instanceID, appKeys, kubevirtScale)
 
 	log.Infof("Destroying apps")
-	destroyAndWait(t, ctxs)
+	destroyAndWait(t, allCtxs)
 
 	// If we are here then the test has passed
 	testResult = testResultPass
 	log.Infof("Test status at end of %s test: %s", t.Name(), testResult)
-}
-
-func kubevirtVMsDeployAndValidate(
-	t *testing.T,
-	instanceID string,
-	appKeys []string,
-) []*scheduler.Context {
-	ctxs, err := schedulerDriver.Schedule(instanceID,
-		scheduler.ScheduleOptions{
-			AppKeys: appKeys,
-		})
-	require.NoError(t, err, "Error scheduling tasks")
-	require.Equal(t, len(appKeys), len(ctxs), "wrong number of tasks started")
-
-	for _, ctx := range ctxs {
-		err = schedulerDriver.WaitForRunning(ctx, 30*time.Minute, defaultWaitInterval)
-		require.NoError(t, err, "Error waiting for app %s to get to running state", ctx.App.Key)
-
-		namespace := appKey + "-" + instanceID
-		vms, err := kubevirt.Instance().ListVirtualMachines(namespace)
-		require.NoError(t, err, "Error listing virtual machines")
-
-		for _, vm := range vms.Items {
-			require.Equal(t, vm.Status.Created, true, "VM %s not created yet", vm.Name)
-			require.Equal(t, vm.Status.Ready, true, "VM %s not ready yet", vm.Name)
-			logrus.Infof("VM %s has %d disks", vm.Name, len(vm.Spec.Template.Spec.Volumes))
-		}
-	}
-	return ctxs
 }
 
 func kubevirtDeployFedoraVMWithClonePVCWaitFirstConsumer(t *testing.T) {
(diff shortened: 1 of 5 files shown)


adityadani (commented):
We need to run scale at two levels:
- Number of VMs.
- Number of DataVolumes per VM -> What is the current number of DataVolumes per VM? We should at least go upto 5 PVCs per VM.

Rohit-PX:
> We need to run scale at two levels:
> 
> * Number of VMs.
> * Number of DataVolumes per VM -> What is the current number of DataVolumes per VM? We should at least go upto 5 PVCs per VM.

Currently, only a one of tests use data volumes (disks other than root disks) and those are hardcoded in the yaml specs deployed for the VM. Will have to modify that test to make this parameterized. I will send out another PR for that.

test/integration_test/kubevirt_test.go
```diff
@@ -5,22 +5,30 @@ package integrationtest
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var templatePVCSpecs = map[string]string{
 	"fedora": "kubevirt-templates",
 }
 
+var kubevirtScale = migrationScaleCount
```
pureneelesh, line 30: I don't understand why this is set to migrationScaleCount. Can you please add a comment?

test/integration_test/kubevirt_test.go
```diff
@@ -49,6 +60,13 @@ func TestKubevirt(t *testing.T) {
 	err = createDatadiskTemplates(t)
 	require.NoError(t, err, "Error creating kubevirt templates")
 
+	err = createClusterroleDataVolume(dataVolumeClusterrole)
+	require.NoError(t, err, "Error creating cluste role for datavolume clone")
+
+	err = createClusterroleBindingDatavolume(dataVolumeClusterrole, kubevirtDatadiskNamespace)
+	require.NoError(t, err, "Error creating/updating data volumeclusterrole binding %s with namespace: %s",
```
pureneelesh, line 67: nit: need space between volume and clusterrole

test/integration_test/kubevirt_test.go
```diff
@@ -67,15 +85,12 @@ func kubevirtDeployFedoraVMWithClonePVC(t *testing.T) {
...
 	instanceID := "vm"
 	appKey := "kubevirt-fedora"
+	var allCtxs []*scheduler.Context
 
-	ctxs := kubevirtVMsDeployAndValidate(
-		t,
-		instanceID,
-		[]string{appKey},
-	)
+	kubevirtVMScaledDeployAndValidate(t, instanceID, appKey, kubevirtScale, &allCtxs)
```
pureneelesh, line 90: Just curious. Why not add the scale param to the existing kubevirtVMsDeployAndValidate function instead of creating a new function? Also, why return the allCtxs C-style with pointer arg? Why not just return it as a return param which is more go-style?

test/integration_test/kubevirt_test.go
```diff
@@ -108,7 +120,14 @@ func kubevirtVMsDeployAndValidate(
 	t *testing.T,
 	instanceID string,
 	appKeys []string,
+	wg *sync.WaitGroup,
 ) []*scheduler.Context {
+	defer wg.Done()
+	// Before validating the VM create required clusterrolebindings for datavolume cloning in the VM namespace
+	testNamespace := appKeys[0] + "-" + instanceID
```
pureneelesh, line 127: We can't assume that there is only on app in the appKeys. Existing tests already pass multiple apps here.

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +258,65 @@ func createDatadiskTemplates(t *testing.T) error {
...
+func createClusterroleDataVolume(clusterroleName string) error {
+	clusterRoleDatavolume := &rbacv1.ClusterRole{
+		ObjectMeta: meta.ObjectMeta{
+			Name: clusterroleName,
+		},
+		Rules: []rbacv1.PolicyRule{{
+			Verbs:     []string{"*"},
+			APIGroups: []string{"*"},
+			Resources: []string{"*"},
+		},
```
pureneelesh, line 289: Hmmm... this is all powerful cluster role. It is not just for the data volume. So, the name is misleading. But should we be doing this? Could this potentially hide bugs?

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +258,65 @@ func createDatadiskTemplates(t *testing.T) error {
...
+// creates/updates clusterrolebindings by adding default ServiceAccount in the given namespace
+func createClusterroleBindingDatavolume(clusterRoleName, namespace string) error {
+	newSubject := rbacv1.Subject{Kind: "ServiceAccount", Name: "default", Namespace: namespace}
+
+	// append subjects if clusterrolebinding already exists else create new cluster role binding
+	cB, err := rbac.Instance().GetClusterRoleBinding(dataVolumeClusterroleBinding)
+	if err == nil && cB != nil {
+		cB.Subjects = append(cB.Subjects, newSubject)
+		_, err = rbac.Instance().UpdateClusterRoleBinding(cB)
+	} else {
```
pureneelesh, line 308: We should do this only if err != NotFound?

test/integration_test/kubevirt_test.go
Rohit-PX: 1. Added a new function instead of using the existing `kubevirtVMsDeployAndValidate` so that if future tests need to call this method, they can still use it w/o worrying about the scale factor
2. Agreed, fixed `allCtxs` to be a returned param from the function

test/integration_test/kubevirt_test.go
Rohit-PX: Fixed.

test/integration_test/kubevirt_test.go
Rohit-PX: This was an error, fixed it.

test/integration_test/kubevirt_test.go
Rohit-PX: Not sure of this comment. The tests in this suite seem to be only one appkey.

test/integration_test/kubevirt_test.go
Rohit-PX: If we are here then it means we could not get the cluster role binding so we create one here, in line:313 we are already checking for the `AlreadyExists` error so don't think  `if err != NotFound` is required here.

test/integration_test/kubevirt_test.go
pureneelesh: Please check all the callers of this function. e.g.
https://github.com/libopenstorage/stork/blob/28c950b28ca3a41dd7bb0a5351c1d1ba124f6877/test/integration_test/kubevirt_hyperconv_test.go#L103

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +251,67 @@ func createDatadiskTemplates(t *testing.T) error {
...
+
+func createClusterroleDataVolume(clusterroleName string) error {
+	clusterRoleDatavolume := &rbacv1.ClusterRole{
+		ObjectMeta: meta.ObjectMeta{
+			Name: clusterroleName,
+		},
+		Rules: []rbacv1.PolicyRule{{
+			Verbs:     []string{"*"},
+			APIGroups: []string{"cdi.kubevirt.io"},
+			Resources: []string{"*"},
```
pureneelesh, line 283: Can we tighten it further and add only the verbs and resources that were missing as per the error message? I believe all we need is "clone" permission.

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +251,67 @@ func createDatadiskTemplates(t *testing.T) error {
 	}
 	return nil
 }
+
+func kubevirtVMScaledDeployAndValidate(t *testing.T, instanceID, appKey string, scale int) []*scheduler.Context {
```
pureneelesh, line 255: I think this function should take multiple appKeys since there are tests that need that. I still think it would be better to add the scale factor to the existing function instead of creating this new function which lacks the ability to deploy multiple apps. 

If the scale is not needed, it is easy to pass 1 as the scale factor.

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +251,67 @@ func createDatadiskTemplates(t *testing.T) error {
...
+		ctxs := kubevirtVMsDeployAndValidate(
+			t,
+			instanceID+"-"+fmt.Sprintf("%d", i),
+			[]string{appKey},
+			&wg,
+		)
+		logrus.Infof("Validated VM: %d", i)
+		allCtxs = append(allCtxs, ctxs...)
+	}
+	wg.Wait()
```
pureneelesh, line 271: I am wondering if sync.WaitGroup is an overkill here. WaitGroup is needed when multiple goroutines are working in parallel and you want all of them to finish before proceeding. But I am not seeing any concurrent work being done here. Am I missing something?

test/integration_test/kubevirt_test.go
Rohit-PX: Agree, fixed this to validate VMs from all namespaces from the context not just the first one.

test/integration_test/kubevirt_test.go
Rohit-PX: I had go routines earlier but forgot to remove the sync group. Fixed.

test/integration_test/kubevirt_test.go
Rohit-PX: Tightened the role by adding the following resources `datavolumes/source` but for verbs we do need `*` for cloning. 
Refer this document: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/virtualization/storage#virt-enabling-user-permissions-to-clone-datavolumes

test/integration_test/kubevirt_test.go
Rohit-PX: Fixed.

Rohit-PX:
Full test suite run: https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%20Cloud%20Jobs/job/stork-ocp-kubevirt-deploy/64/console

Rohit-PX:
Full test suite run: https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%20Cloud%20Jobs/job/stork-ocp-kubevirt-deploy/64/

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +216,87 @@ func createDatadiskTemplates(t *testing.T) error {
 	}
 	return nil
 }
+
+func kubevirtVMScaledDeployAndValidate(t *testing.T, instanceID string, appKeys []string, scale int) []*scheduler.Context {
+	var allCtxs []*scheduler.Context
+	logrus.Infof("Deploying %d VMs and validating them.", scale)
```
pureneelesh, line 222: nit: this is misleading since appKeys could have multiple apps. So, the number of VMs is `len(appKeys) * scale`

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +216,87 @@ func createDatadiskTemplates(t *testing.T) error {
...
+		for _, ctx := range ctxs {
+			err = schedulerDriver.WaitForRunning(ctx, 30*time.Minute, defaultWaitInterval)
+			require.NoError(t, err, "Error waiting for app %s to get to running state", ctx.App.Key)
+
+			vms, err := kubevirt.Instance().ListVirtualMachines(ctx.App.NameSpace)
+			require.NoError(t, err, "Error listing virtual machines")
+			for _, vm := range vms.Items {
+				require.Equal(t, vm.Status.Created, true, "VM %s not created yet", vm.Name)
+				require.Equal(t, vm.Status.Ready, true, "VM %s not ready yet", vm.Name)
+				logrus.Infof("VM %s has %d disks", vm.Name, len(vm.Spec.Template.Spec.Volumes))
```
pureneelesh, line 250: add the namespace to the error messages to make the errors easier to debug

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +216,87 @@ func createDatadiskTemplates(t *testing.T) error {
 	}
 	return nil
 }
+
+func kubevirtVMScaledDeployAndValidate(t *testing.T, instanceID string, appKeys []string, scale int) []*scheduler.Context {
+	var allCtxs []*scheduler.Context
+	logrus.Infof("Deploying %d VMs and validating them.", scale)
+	for i := 1; i <= scale; i++ {
+		logrus.Infof("Deploying VM: %d", i)
```
pureneelesh, line 224: since we are deploying multiple VMs in each iteration (one per appKey), we could use a better message here
e.g. `Deploying %d VMs in the iteration %d of %d", len(apKeys), i, scale`

test/integration_test/kubevirt_test.go
```diff
@@ -246,3 +216,87 @@ func createDatadiskTemplates(t *testing.T) error {
...
+			vms, err := kubevirt.Instance().ListVirtualMachines(ctx.App.NameSpace)
+			require.NoError(t, err, "Error listing virtual machines")
+			for _, vm := range vms.Items {
+				require.Equal(t, vm.Status.Created, true, "VM %s not created yet", vm.Name)
+				require.Equal(t, vm.Status.Ready, true, "VM %s not ready yet", vm.Name)
+				logrus.Infof("VM %s has %d disks", vm.Name, len(vm.Spec.Template.Spec.Volumes))
+			}
+
+		}
+		logrus.Infof("Validated VM: %d", i)
```
pureneelesh, line 254: improve message like the above

pureneelesh (approved):

test/integration_test/kubevirt_test.go
Rohit-PX: Fixed.

PR #1649, opened 2024-02-22
PWX-35855: Action CR reconciler changes with multiple stages.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
feature

1. Added action CR reconciler workflow with respect to different actions and multiple stages.
2. Added the code for DeactivateDestination of failback action.
3. Unit tests

Tests:
Manually run following cases and results have been updated in the ticket.
```
1. With an old successful migration.
2. With the latest migration failed.
3. With migration is in progress.
4. With more namespaces in action CR with reference to migration schedule (not a subset).
5. With latest successful migration within the schedule policy interval range.

```

### DIFF
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -205,7 +205,8 @@ func getLatestMigrationPolicyAndStatus(migrationSchedule storkv1.MigrationSchedu
 }
 
 func isMigrationComplete(status storkv1.MigrationStatusType) bool {
-	if status == storkv1.MigrationStatusPending ||
+	if status == storkv1.MigrationStatusInitial ||
+		status == storkv1.MigrationStatusPending ||
 		status == storkv1.MigrationStatusInProgress {
 		return false
 	}
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -26,9 +26,10 @@ import (
 )
 
 const (
-	validateCRDInterval time.Duration = 5 * time.Second
-	validateCRDTimeout  time.Duration = 1 * time.Minute
-	actionExpiryTime    time.Duration = 24 * time.Hour
+	validateCRDInterval          time.Duration = 5 * time.Second
+	validateCRDTimeout           time.Duration = 1 * time.Minute
+	actionExpiryTime             time.Duration = 24 * time.Hour
+	latestMigrationThresholdTime time.Duration = 1 * time.Hour
 )
 
 func NewActionController(mgr manager.Manager, d volume.Driver, r record.EventRecorder) *ActionController {
@@ -69,30 +70,61 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
 		return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
 	}
 
-	if action.Status.Status == storkv1.ActionStatusScheduled {
-		if err = ac.handle(context.TODO(), action); err != nil {
-			log.ActionLog(action).Errorf("ActionController handle failed with %s", err)
-			ac.recorder.Event(action, v1.EventTypeWarning, "ActionController", err.Error())
-			return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
-		}
+	if action.Status.Stage == storkv1.ActionStageFinal {
+		// Will not requeue once action is in final stage
+		return reconcile.Result{}, nil
 	}
-	ac.pruneActionIfExpired(action)
+
+	if !controllers.ContainsFinalizer(action, controllers.FinalizerCleanup) {
+		controllers.SetFinalizer(action, controllers.FinalizerCleanup)
+		return reconcile.Result{Requeue: true}, ac.client.Update(context.TODO(), action)
+	}
+
+	if err = ac.handle(context.TODO(), action); err != nil {
+		log.ActionLog(action).Errorf("ActionController handle failed with %s", err)
+		ac.recorder.Event(action, v1.EventTypeWarning, "ActionController", err.Error())
+		return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
+	}
+
+	if action.Spec.ActionType == storkv1.ActionTypeNearSyncFailover {
+		ac.pruneActionIfExpired(action)
+	}
 
 	return reconcile.Result{RequeueAfter: controllers.DefaultRequeue}, nil
 }
 
 func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) error {
-	ac.updateStatus(action, storkv1.ActionStatusInProgress)
+	if action.DeletionTimestamp != nil {
+		if action.GetFinalizers() != nil {
+			controllers.RemoveFinalizer(action, controllers.FinalizerCleanup)
+			return ac.client.Update(ctx, action)
+		}
+		return nil
+	}
+
 	switch action.Spec.ActionType {
 	case storkv1.ActionTypeNearSyncFailover:
-		log.ActionLog(action).Info("started failover")
-		err := ac.volDriver.Failover(action)
-		if err != nil {
-			ac.updateStatus(action, storkv1.ActionStatusFailed)
-			return err
+		ac.updateStatus(action, storkv1.ActionStatusInProgress)
+		if action.Status.Status == storkv1.ActionStatusScheduled {
+			log.ActionLog(action).Info("started failover")
+			err := ac.volDriver.Failover(action)
+			if err != nil {
+				ac.updateStatus(action, storkv1.ActionStatusFailed)
+				return err
+			}
+			resourceutils.ScaleReplicas(action.Namespace, true, ac.printFunc(action, "ScaleReplicas"), ac.config)
+			ac.updateStatus(action, storkv1.ActionStatusSuccessful)
 		}
-		resourceutils.ScaleReplicas(action.Namespace, true, ac.printFunc(action, "ScaleReplicas"), ac.config)
-		ac.updateStatus(action, storkv1.ActionStatusSuccessful)
+	case storkv1.ActionTypeFailback:
+		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
+		switch action.Status.Stage {
+		case storkv1.ActionStageInitial:
+			ac.verifyMigrationScheduleBeforeFailback(action)
+		case storkv1.ActionStageScaleDownDestination:
+			ac.performDeactivateFailBackStage(action)
+		case storkv1.ActionStageFinal:
+			return nil
+		}
 	default:
 		ac.updateStatus(action, storkv1.ActionStatusFailed)
 		return fmt.Errorf("invalid value received for Action.Spec.ActionType")
@@ -126,6 +158,15 @@ func (ac *ActionController) updateStatus(action *storkv1.Action, actionStatus st
 	}
 }
 
+func (ac *ActionController) updateAction(action *storkv1.Action) error {
+	err := ac.client.Update(context.TODO(), action)
+	if err != nil {
+		log.ActionLog(action).Errorf("failed to update action status to %v with error %v", action.Status, err)
+		return err
+	}
+	return nil
+}
+
 // delete any action older than actionExpiryTime
 func (ac *ActionController) pruneActionIfExpired(action *storkv1.Action) {
 	if action.Status.Status == storkv1.ActionStatusFailed || action.Status.Status == storkv1.ActionStatusSuccessful {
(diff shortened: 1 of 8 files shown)


pkg/action/actioncontroller.go
```diff
@@ -149,3 +190,24 @@ func (ac *ActionController) createCRD() error {
...
+	for schedulePolicyType, policyMigration := range migrationSchedule.Status.Items {
+		if len(policyMigration) > 0 {
+			lastMigrationStatus = policyMigration[len(policyMigration)-1]
+			return schedulePolicyType, lastMigrationStatus
+		}
+	}
+	return schedulePolicyType, lastMigrationStatus
+}
+
+func isMigrationComplete(status storkv1.MigrationStatusType) bool {
```
olavangad-px, line 207: I feel we can add status == storkv1.MigrationStatusInitial as well in the condition here 
So new condition should be 
```
if status == storkv1.MigrationStatusInitial ||
                status == storkv1.MigrationStatusPending ||
		status == storkv1.MigrationStatusInProgress {
		return false
	}
```

pkg/action/failback.go
```diff
@@ -0,0 +1,216 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (ac *ActionController) verifyMigrationScheduleBeforeFailback(action *storkv1.Action) {
+
+	if action.Status.Status == storkv1.ActionStatusSuccessful {
+		action.Status.Stage = storkv1.ActionStageScaleDownDestination
+		action.Status.Status = storkv1.ActionStatusInitial
+		action.Status.FinishTimestamp = metav1.Now()
+		ac.updateAction(action)
```
olavangad-px, line 22: I see that updateAction method's call returns an error. Do we not need to handle that error??

pkg/action/failback.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/libopenstorage/stork/pkg/resourceutils"
+	"github.com/libopenstorage/stork/pkg/schedule"
+	"github.com/libopenstorage/stork/pkg/utils"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (ac *ActionController) verifyMigrationScheduleBeforeFailback(action *storkv1.Action) {
+
```
olavangad-px, line 17: nit: I think adding a comment here describing the method's responsibility and stage of failback this method corresponds to would be helpful

olavangad-px: On a second thought the method's name is self-descriptive. Maybe a comment is not required then

pkg/utils/utils.go
```diff
@@ -334,3 +335,38 @@ func addResourceVersion(patchBytes []byte, resourceVersion string) ([]byte, erro
...
+	}
+
+	migrationNamespaces := make([]string, 0, len(uniqueNamespaces))
+	for namespace := range uniqueNamespaces {
+		migrationNamespaces = append(migrationNamespaces, namespace)
+	}
+	return migrationNamespaces, nil
+}
+
+func IsSubList(list1 []string, list2 []string) bool {
```
olavangad-px, line 365: Although I think in our case it practically doesn't matter, the time complexity of this method is O(m*n), while we could instead use a map and do it in O(m+n) where m and n are lengths of list1 and list2.
I had written the below function as part of #1631 
```
// isSubset returns true if the first slice is subset of the second slice.
// If false it also returns the list of non subset strings.
func isSubset(listA []string, listB []string) (bool, []string) {
	nonSubsetStrings := make([]string, 0)
	superset := make(map[string]bool)
	for _, str := range listB {
		superset[str] = true
	}
	for _, str := range listA {
		if !superset[str] {
			nonSubsetStrings = append(nonSubsetStrings, str)
		}
	}
	return len(nonSubsetStrings) == 0, nonSubsetStrings
}
```

pkg/action/failback.go
```diff
@@ -0,0 +1,216 @@
...
+				v1.EventTypeWarning,
+				string(storkv1.ActionStatusFailed),
+				msg)
+			ac.updateAction(action)
+			return
+		}
+		if alreadyTriggered {
+			log.ActionLog(action).Infof("The latest migration %s is in threshold range", latestMigration.Name)
+			// Suspend migration schedule
+			log.ActionLog(action).Infof("Suspending the migration schedule %s", migrationSchedule.Name)
```
olavangad-px, line 130: nit: i don't think this particular log is needed since we have a similar one in the below line#135

pkg/action/failback.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if action.Status.Status == storkv1.ActionStatusSuccessful {
+		action.Status.Stage = storkv1.ActionStageScaleDownDestination
+		action.Status.Status = storkv1.ActionStatusInitial
+		action.Status.FinishTimestamp = metav1.Now()
+		ac.updateAction(action)
+		return
+	} else if action.Status.Status == storkv1.ActionStatusFailed {
+		action.Status.Stage = storkv1.ActionStageFinal
+		action.Status.Status = storkv1.ActionStatusInitial
```
olavangad-px, line 26: would setting the status to ActionStatusFailed (instead of ActionStatusInitial)along with stage ActionStatusFinal make more sense?

diptiranjanpx: Yes, modified accordingily.

pkg/action/failback.go
diptiranjanpx: Done

pkg/action/actioncontroller.go
diptiranjanpx: Done

pkg/action/failback.go
diptiranjanpx: After updateAction , we anyway return in current cases.
Wherever the error needs to be handled, the respective code will get added at that time.

As of now, if it fails the stage remains same, so the stage will be rerun.

pkg/action/failback.go
```diff
@@ -0,0 +1,213 @@
...
+		ac.recorder.Event(action,
+			v1.EventTypeWarning,
+			string(storkv1.ActionStatusScheduled),
+			msg)
+		ac.updateAction(action)
+		return
+	}
+
+	for _, ns := range namespaces {
+		err := resourceutils.ScaleReplicasReturningError(ns, true, true, ac.printFunc(action, "ScaleReplicas"), ac.config)
```
olavangad-px, line 199: I am thinking of another case. What if user performs a failover from C1->C2, and then adds new apps in C2 before doing a failover from C2->C1. In this case the new apps will not have a migrationReplicas annotation. I don't think this case is handled by current updateReplicas method.
Also another point if the originalReplicaCount is 0, do we still want to update the replicaAnnotation which will then be carried forward to the source cluster C1 ??

diptiranjanpx: I see your point. To handle this case, one way can be, we can reuse the same flow of checking what is getting activated and then deactivate the corresponding app (similar to failover). I will do that in another PR.

olavangad-px (approved):
LGTM

pkg/action/failback.go
```diff
@@ -0,0 +1,213 @@
...
+			if *migrationSchedule.Spec.Suspend {
+				migrationSchedule.Spec.Suspend = &suspend
+			}
+			msg := fmt.Sprintf("Suspending migration schedule %s before doing failback operation", migrationSchedule.Name)
+			log.ActionLog(action).Infof(msg)
+			_, err := storkops.Instance().UpdateMigrationSchedule(migrationSchedule)
+			if err != nil {
+				log.ActionLog(action).Errorf("Error suspending migration schedule %s: %v", migrationSchedule.Name, err)
+			}
+			action.Status.Status = storkv1.ActionStatusSuccessful
```
pp511, line 139: We will update to ActionStatusSuccessful despite the error in the statement above?

pkg/action/failback.go
```diff
@@ -0,0 +1,213 @@
...
+			ac.updateAction(action)
+		} else {
+			// Fail the action failback if latest successful migration is older than latestMigrationThresholdTime
+			// And migrationschedule is in suspended state
+			log.ActionLog(action).Infof("The latest migration %s is not in threshold range", latestMigration.Name)
+			msg := fmt.Sprintf("Failing failback operation as the latest migration %s is not in threshold range", latestMigration.Name)
```
pp511, line 145: nit..threshold time range. 
This is just to be explicit and avoid confusion. Also are we going to publish the latestMigrationThresholdTime in our docs?
Maybe we can just say threshold time range of 1 hr...

pkg/schedule/schedule.go
```diff
@@ -148,6 +148,88 @@ func checkTrigger(
 	return false, nil
 }
 
+func hasTriggered(
+	lastTrigger time.Time,
+	lastTriggerEarliest time.Time,
+	now time.Time,
+) (bool, error) {
+	// If the last trigger is before the  we had triggered before the scheduled time this month, don't
+	// triggered again
```
pp511, line 157: take another look at this comment and correct it.

pp511 (commented):
Did take one pass at it. Will try to take another. Also didn't review updatereplicas.go yet. Overall looks ok. Just some questions till then.

pkg/action/actioncontroller.go
```diff
@@ -69,30 +70,61 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
...
+	}
+
 	switch action.Spec.ActionType {
 	case storkv1.ActionTypeNearSyncFailover:
-		log.ActionLog(action).Info("started failover")
-		err := ac.volDriver.Failover(action)
-		if err != nil {
-			ac.updateStatus(action, storkv1.ActionStatusFailed)
-			return err
+		ac.updateStatus(action, storkv1.ActionStatusInProgress)
```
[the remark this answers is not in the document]
diptiranjanpx: So earlier this code block was part of reconciller function and it was common for any action. Now with this change, I moved that to handle() and made it only specific for near-sync. It's not new code.

pkg/action/actioncontroller.go
```diff
@@ -149,3 +190,25 @@ func (ac *ActionController) createCRD() error {
...
+func getLatestMigrationPolicyAndStatus(migrationSchedule storkv1.MigrationSchedule) (storkv1.SchedulePolicyType, *storkv1.ScheduledMigrationStatus) {
+	var lastMigrationStatus *storkv1.ScheduledMigrationStatus
+	var schedulePolicyType storkv1.SchedulePolicyType
+	for schedulePolicyType, policyMigration := range migrationSchedule.Status.Items {
+		if len(policyMigration) > 0 {
+			lastMigrationStatus = policyMigration[len(policyMigration)-1]
+			return schedulePolicyType, lastMigrationStatus
+		}
+	}
+	return schedulePolicyType, lastMigrationStatus
```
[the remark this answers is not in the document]
diptiranjanpx: MigrationScheduleStatus.Items is a map[SchedulePolicyType][]*ScheduledMigrationStatus. 

Because migrationschedule supports all the different types of policies the status can have the key of any policy type. ( I agree that it will be mostly a single policy type but still items will be a map).
The idea is to get for any key whose entries are more than one (that means thats the policy type getting used) as the policyKey and among those pick out the last appended one which is the latest. 

The similar looping also gets used in migrationController code.

pkg/schedule/schedule.go
diptiranjanpx: Modified.

pkg/action/failback.go
diptiranjanpx: returning if error happens so that the stage gets rerun again.

pkg/action/failback.go
diptiranjanpx: Now it is not 1 hr time. It's based on the schedule policy. We are looking for a successful migration within the time range ofallowed earliest time based on schedule policy from now and now.

I have modified the message.

pkg/action/actioncontroller.go
```diff
@@ -149,3 +190,25 @@ func (ac *ActionController) createCRD() error {
...
+}
+
+func isMigrationComplete(status storkv1.MigrationStatusType) bool {
+	if status == storkv1.MigrationStatusInitial ||
+		status == storkv1.MigrationStatusPending ||
+		status == storkv1.MigrationStatusInProgress {
+		return false
+	}
+	return true
+}
```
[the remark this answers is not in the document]
diptiranjanpx: done

pkg/action/failback.go
```diff
@@ -0,0 +1,214 @@
...
+				v1.EventTypeWarning,
+				string(storkv1.ActionStatusScheduled),
+				msg)
+			ac.updateAction(action)
+			return
+		}
+
+	}
+
+	// Return if any migration is ongoing
```
olavangad-px, line 71: Sorry for not asking this before. 
Can we not directly use the latestMigration fetched below for this scenario?
if latestMigration.Status is !isMigrationComplete then we assume that we have to wait for it's completion.

diptiranjanpx: Will take care of it. Thanks

PR #1656, opened 2024-03-04
[PWX-36291] Helper functions for scaling down applications during failover/failback

olavangad-px:
**What type of PR is this?**
>feature

This PR deals with the methods required for scaling down applications during failover and failback. 
Note: These functions are not being used yet. The upcoming PR will use these methods.

No

No

Yes, 24.2.0

**Testing Details**
For failover, we will first find out the resources which will be activated in the destination cluster and scale down the corresponding resources in the source cluster.
To simulate this for testing I used the below code in one of the storkctl commands and verified that a deployment,replicaSet & a mongocommunitydb resource (CRD) are scaled down in source cluster as expected and required annotations are added as well.
```
srcConf, _ := getConfig(srcConfig).ClientConfig()
destConf, _ := cmdFactory.GetConfig()
activationNamespaces := []string{cmdFactory.GetNamespace()}
resourceBeingActivated, _ := resourceutils.GetResourcesBeingActivated(activationNamespaces, destConf)
_ = resourceutils.ScaleDownGivenResources(activationNamespaces, resourceBeingActivated, srcConf)
```

### DIFF
--- a/pkg/resourceutils/listResources.go
+++ b/pkg/resourceutils/listResources.go
@@ -1,0 +1,157 @@
+package resourceutils
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	migration "github.com/libopenstorage/stork/pkg/migration/controllers"
+	"github.com/libopenstorage/stork/pkg/resourcecollector"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8sdynamic "k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+)
+
+var statefulSetGVK = metav1.GroupVersionKind{Group: "apps", Version: "v1", Kind: "StatefulSet"}
+var deploymentGVK = metav1.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"}
+var replicaSetGVK = metav1.GroupVersionKind{Group: "apps", Version: "v1", Kind: "ReplicaSet"}
+var deploymentConfigGVK = metav1.GroupVersionKind{Group: "apps.openshift.io", Version: "v1", Kind: "DeploymentConfig"}
+var cronJobGVK = metav1.GroupVersionKind{Group: "batch", Version: "v1", Kind: "CronJob"}
+var virtualMachineGVK = metav1.GroupVersionKind{Group: "kubevirt.io", Version: "v1", Kind: "VirtualMachine"}
+var ibpPeerGVK = metav1.GroupVersionKind{Group: "ibp.com", Version: "v1alpha1", Kind: "IBPPeer"}
+var ibpcaGVK = metav1.GroupVersionKind{Group: "ibp.com", Version: "v1alpha1", Kind: "IBPCA"}
+var ibpConsoleGVK = metav1.GroupVersionKind{Group: "ibp.com", Version: "v1alpha1", Kind: "IBPOrderer"}
+var ibpOrdererGVK = metav1.GroupVersionKind{Group: "ibp.com", Version: "v1alpha1", Kind: "IBPConsole"}
+
+func getRelevantGVKs() []metav1.GroupVersionKind {
+	return []metav1.GroupVersionKind{statefulSetGVK, deploymentGVK, replicaSetGVK, deploymentConfigGVK, cronJobGVK, virtualMachineGVK, ibpPeerGVK, ibpcaGVK, ibpOrdererGVK, ibpConsoleGVK}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/resourceutils/listResources.go
```diff
@@ -0,0 +1,167 @@
...
+	crdList, err := storkClient.ListApplicationRegistrations()
+	if err != nil {
+		return nil, err
+	}
+	resourceReplicaMap := make(map[string]map[metav1.GroupVersionKind]map[string]string)
+	for _, ns := range activationNamespaces {
+		//Initialise namespace -> resource kind maps
+		resourceReplicaMap[ns] = make(map[metav1.GroupVersionKind]map[string]string)
+
+		// 1. ApplicationResources
```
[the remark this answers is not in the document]
olavangad-px: I have already created functions to scale down individual resources for each of application, crd and cronJob.
In this function I am just looping over the resources and calling the corresponding update method.
Should I still create new functions just to loop over the resources ?

diptiranjanpx (approved):

pkg/resourceutils/listResources.go
```diff
@@ -0,0 +1,192 @@
...
+					return nil, fmt.Errorf("error getting the annotations of the resource %v %v/%v: %v", strings.ToLower(cronJobGVK.Kind), object.GetNamespace(), object.GetName(), err)
+				}
+				if !found {
+					continue
+				}
+				// pick only if stork-migrated annotation exists in the resource
+				if val, present := annotations[migration.StorkMigrationAnnotation]; present {
+					resourceReplicaMap[ns][cronJobGVK][object.GetName()] = val
+				}
+			}
```
pp511, line 138: Did we need to create another map here? Couldn't we use the map we have creared in Step1?

pkg/resourceutils/scaleDown.go
```diff
@@ -0,0 +1,270 @@
...
+
+func scaleDownApplicationResource(dynamicClient k8sdynamic.ResourceInterface, object unstructured.Unstructured, resourceType metav1.GroupVersionKind) error {
+	content := object.UnstructuredContent()
+	currentReplicas, found, err := unstructured.NestedInt64(content, "spec", "replicas")
+	if err != nil {
+		return err
+	}
+	if !found {
+		currentReplicas = 1
+	}
```
pp511, line 98: Explain this.

pkg/resourceutils/scaleDown.go
```diff
@@ -0,0 +1,270 @@
...
+					if annotationValue != "" {
+						// If the actual value is equal to suspend.Value (i.e. 0 value) and suspend path annotation value is not an empty string,
+						// then don't override the annotation
+						currentValue = annotationValue
+					}
+				}
+				val = currentValue
+			} else if suspend.Type == "bool" {
+				if val, err := strconv.ParseBool(suspend.Value); err != nil {
+					disableVersion = true
```
pp511, line 218: Shouldn't we log that there was a parsing error here?

pkg/resourceutils/scaleDown.go
```diff
@@ -0,0 +1,270 @@
...
+	opts := &metav1.UpdateOptions{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       resourceType.Kind,
+			APIVersion: resourceType.Group + "/" + resourceType.Version},
+	}
+	_, err = dynamicClient.Update(context.TODO(), &object, *opts, "")
+	if err != nil {
+		return fmt.Errorf("unable to update resource %v %v/%v: %v", strings.ToLower(resourceType.Kind), object.GetNamespace(), object.GetName(), err)
+	}
+	fmt.Printf("successfully updated resource %v %v/%v \n", strings.ToLower(resourceType.Kind), object.GetNamespace(), object.GetName())
```
pp511, line 136: Shouldn't we be using log. in case of Printf?

pkg/resourceutils/scaleDown.go
olavangad-px: I have followed the scaling down logic from prepareApplicationResources method in the migration flow.
https://github.com/libopenstorage/stork/blob/5062314469757f8e10b41a6c97c693c07995069c/pkg/migration/controllers/migration.go#L1474

pkg/resourceutils/scaleDown.go
olavangad-px: Makes sense. Will add a log statement

pkg/resourceutils/scaleDown.go
olavangad-px: Yes makes sense. Will update everywhere applicable

pkg/resourceutils/listResources.go
olavangad-px: The resource replica map is a 3 level map
First Key is Namespace
Second Key is the gvk
Third Key is the resource name
And finally the value is the replica value we will set on activation.
Though the value is not required in our current scenario for scaling down, I created the method like this because then it can be plugged in into failover destination activation dry run.

Coming to your point, the map creation is required because resourceReplicaMap[ns][cronJobGVK] is cronJob specific map and we will populate it with the cronJob resources and their new replica values in this particular namespace.

pp511 (approved):

PR #1657, opened 2024-03-05
fix: PB-5878 - Handle freeze/thaw for case with two VMs with same nam…

vsundarraj-px:
Handle freeze/thaw for case with two VMs with same nam

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

Fixes Bug PB-5878 where autoExecRules fail if there are more than one vm with same name but different namespace

no

```release-note
Issue: execRules fails to run if there are two VMs with same name during backup
User Impact: px-backup will fail.
Resolution: Fix 

```

release-24.1.0

### DIFF
--- a/pkg/resourcecollector/virtualmachine.go
+++ b/pkg/resourcecollector/virtualmachine.go
@@ -1,6 +1,7 @@
 package resourcecollector
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


siva-portworx (approved):
LGTM

pkg/resourcecollector/virtualmachine.go
```diff
@@ -25,7 +26,8 @@ const (
 	//VMUnFreezeCmd is the template for VM unfreeze command
 	VMUnFreezeCmd = "/usr/bin/virt-freezer --unfreeze --name %s --namespace %s"
 	// VMContainerName is the name of the container to use for freeze/thaw
-	VMContainerName = "compute"
+	VMContainerName             = "compute"
+	VMPodSelectorCreatedByLabel = "kubevirt.io/created-by"
```
prashanthpx, line 30: Would this label be unique for VM created across different ns?

vsundarraj-px: yes

PR #1665, opened 2024-03-13
PWX-35885: Doing update on latest copy of volumesnapshotschedule to avoid update failures not happening on latest version.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

Doing GetAndUpdate the volumesnapshotschedule.
Getting the latest copy and doing update is one more api call but currently it does not affect each reconciler cycle .  It happens in those cycles where
1. New volumesnapshot need to be created
2. Pruning of previous snapshots are required.

Generally the policy cycles are much larger compared to reconciler cycle, hence it is ok to do one more api call when the updating is absolutely required.


pp511 (approved):
lgtm

PR #1666, opened 2024-03-13
[PWX-35853] Failover Controller all stages

olavangad-px:
**What type of PR is this?**
>feature

Added all the stages of failover handling to the action controller. 
Note: as of this PR, updateReplicas method used is incorrect because it uses the singleton clients to update/list resources while we should create the clients using given config since in failover's rollback stage we need to activate resources in remote cluster. Changes for fixing that will be taken up in the next PR.
Testing details for the scenarios covered with this PR are attached in the corresponding Jira. https://portworx.atlassian.net/browse/PWX-35853

No

No

Yes 24.2.0

### DIFF
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -26,10 +26,9 @@ import (
 )
 
 const (
-	validateCRDInterval          time.Duration = 5 * time.Second
-	validateCRDTimeout           time.Duration = 1 * time.Minute
-	actionExpiryTime             time.Duration = 24 * time.Hour
-	latestMigrationThresholdTime time.Duration = 1 * time.Hour
+	validateCRDInterval time.Duration = 5 * time.Second
+	validateCRDTimeout  time.Duration = 1 * time.Minute
+	actionExpiryTime    time.Duration = 24 * time.Hour
 
 	nameTimeSuffixFormat    string = "2006-01-02-150405"
 	lastmileMigrationPrefix string = "lastmile"
@@ -76,10 +75,10 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
 		return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
 	}
 
-	if action.Status.Stage == storkv1.ActionStageFinal {
-		// Will not requeue once action is in final stage
-		return reconcile.Result{}, nil
-	}
+	// if action.Status.Stage == storkv1.ActionStageFinal {
+	// 	// Will not requeue once action is in final stage
+	// 	return reconcile.Result{}, nil
+	// }
 
 	if !controllers.ContainsFinalizer(action, controllers.FinalizerCleanup) {
 		controllers.SetFinalizer(action, controllers.FinalizerCleanup)
@@ -110,6 +109,9 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) 
 
 	switch action.Spec.ActionType {
 	case storkv1.ActionTypeNearSyncFailover:
+		if action.Status.Stage == storkv1.ActionStageFinal {
+			return nil
+		}
 		ac.updateStatus(action, storkv1.ActionStatusInProgress)
 		if action.Status.Status == storkv1.ActionStatusScheduled {
 			log.ActionLog(action).Info("started failover")
@@ -120,6 +122,22 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) 
 			}
 			resourceutils.ScaleReplicas(action.Namespace, true, ac.printFunc(action, "ScaleReplicas"), ac.config)
 			ac.updateStatus(action, storkv1.ActionStatusSuccessful)
+		}
+	case storkv1.ActionTypeFailover:
+		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
+		switch action.Status.Stage {
+		case storkv1.ActionStageInitial:
+			ac.validationsBeforeFailover(action)
+		case storkv1.ActionStageScaleDownSource:
+			ac.deactivateClusterStage(action)
+		case storkv1.ActionStageLastMileMigration:
+			ac.performLastMileMigration(action)
+		case storkv1.ActionStageScaleUpDestination:
+			ac.activateDuringFailover(action, false)
+		case storkv1.ActionStageScaleUpSource:
+			ac.activateDuringFailover(action, true)
+		case storkv1.ActionStageFinal:
+			return nil
 		}
 	case storkv1.ActionTypeFailback:
 		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
@@ -203,7 +221,7 @@ func (ac *ActionController) createCRD() error {
 	return k8sutils.CreateCRD(resource, validateCRDInterval, validateCRDTimeout)
 }
 
-// getLatestMigrationStatus returns the a migrationschedule's latest migration's policy type and status
+// getLatestMigrationStatus returns the migrationschedule's latest migration's policy type and status
 func getLatestMigrationPolicyAndStatus(migrationSchedule storkv1.MigrationSchedule) (storkv1.SchedulePolicyType, *storkv1.ScheduledMigrationStatus) {
 	var lastMigrationStatus *storkv1.ScheduledMigrationStatus
 	var schedulePolicyType storkv1.SchedulePolicyType
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -75,11 +75,6 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
 		return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
 	}
 
-	// if action.Status.Stage == storkv1.ActionStageFinal {
-	// 	// Will not requeue once action is in final stage
-	// 	return reconcile.Result{}, nil
-	// }
-
 	if !controllers.ContainsFinalizer(action, controllers.FinalizerCleanup) {
 		controllers.SetFinalizer(action, controllers.FinalizerCleanup)
 		return reconcile.Result{Requeue: true}, ac.client.Update(context.TODO(), action)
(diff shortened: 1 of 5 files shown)


pkg/action/actioncontroller.go
```diff
@@ -76,10 +75,10 @@ func (ac *ActionController) Reconcile(ctx context.Context, request reconcile.Req
 		return reconcile.Result{RequeueAfter: controllers.DefaultRequeueError}, err
 	}
 
-	if action.Status.Stage == storkv1.ActionStageFinal {
```
olavangad-px, line 79: the presence of this code block was resulting in action resources in Final stage not entering the handle block and hence we were not able to remove the finalizer to be able to delete the action resource. So have commented this block for now. Side effect is that even action resources in Final state will get requeued.

diptiranjanpx: Let's remove the commented lines, this is the behaviour in most of our other controllers where they enter into handle() and nothing happens because it's in Final stage.

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+		msg := fmt.Sprintf("error getting migrationschedule %s/%s for failover", action.Namespace, action.Spec.ActionParameter.FailoverParameter.MigrationScheduleReference)
+		logEvents := ac.printFunc(action, string(storkv1.ActionStatusFailed))
+		logEvents(msg, "err")
+		action.Status.Status = storkv1.ActionStatusFailed
+		action.Status.Reason = msg
+		ac.updateAction(action)
+		return
+	}
+	// get sourceConfig from clusterPair in the destination cluster
+	sourceConfig, err := getClusterPairSchedulerConfig(migrationSchedule.Spec.Template.Spec.ClusterPair, migrationSchedule.Namespace)
```
diptiranjanpx, line 54: Don't we have to consider "skip-source-actions" flag?

pkg/action/actioncontroller.go
olavangad-px: Makes sense. Will remove it during the revision.

pkg/action/failover.go
olavangad-px: I have not yet updated cli and controller according to the new discussion around skip flag. 
I will take that up in a separate PR.
For now, I have enforced clusterpair for everyone. As part of a susbsequent PR I can modify it such that we skip it if skip-source-operations flag is true

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+			// a migration is still in progress. failover needs to wait
+			msg := fmt.Sprintf("Waiting for the completion of migration %s/%s", srcMigrSched.Namespace, latestMigration.Name)
+			log.ActionLog(action).Infof(msg)
+			ac.recorder.Event(action,
+				v1.EventTypeWarning,
+				string(storkv1.ActionStatusScheduled),
+				msg)
+			ac.updateAction(action)
+			return
+		} else if latestMigration.Status == storkv1.MigrationStatusFailed {
```
pp511, line 128: Is it possible that only one migration happend and it failed. And we will still trigger a failover instead of showing an error?

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+		if slices.Contains(migrationNamespaces, ns) {
+			err := resourceutils.ScaleReplicasReturningError(ns, true, false, ac.printFunc(action, "ScaleReplicas"), config)
+			if err != nil {
+				scaleupStatus = false
+				msg := fmt.Sprintf("scaling up apps in namespace %s failed: %v", ns, err)
+				log.ActionLog(action).Errorf(msg)
+				failoverSummary = &storkv1.FailoverSummary{Namespace: ns, Status: storkv1.ActionStatusFailed, Reason: msg}
+			}
+			msg := fmt.Sprintf("scaling up apps in namespace %s successful", ns)
+			failoverSummary = &storkv1.FailoverSummary{Namespace: ns, Status: storkv1.ActionStatusSuccessful, Reason: msg}
```
pp511, line 327: In case of error we are first updating failoverSummary with "failed" msg and later to successful msg.

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+		} else {
+			msg := fmt.Sprintf("skipping scaling up apps in namespace %s since it is not one of the namespaces being migrated by the MigrationSchedule %s/%s", ns, migrationSchedule.Namespace, migrationSchedule.Name)
+			failoverSummary = &storkv1.FailoverSummary{Namespace: ns, Status: storkv1.ActionStatusSuccessful, Reason: msg}
+		}
+		failoverSummaryList = append(failoverSummaryList, failoverSummary)
+	}
+
+	if reactivate {
+		// marking the stage's status as successful even if it hit error while activating apps up
```
pp511, line 336: explain the rationale here.

pp511 (commented):
Mostly looks ok. Some nits. One question and one possible logging issue.

pkg/action/failover.go
olavangad-px: Just to clarify what all is left even after this PR.
1. Cluster domain deactivate/activate and operations specific to sync-dr usecase.
2. Fixing the updateReplicas method used to scale up apps so that both reactivate and activate work as expected.
3. Design change discussed earlier this week around --skip-source-operations flag (this will require changes in cli as well)

pkg/action/failover.go
olavangad-px: Thanks for pointing this out. I missed an else there. Will fix it as part of the revision
It should be 
```
if err != nil {
	scaleupStatus = false
	msg := fmt.Sprintf("scaling up apps in namespace %s failed: %v", ns, err)
	log.ActionLog(action).Errorf(msg)
	failoverSummary = &storkv1.FailoverSummary{Namespace: ns, Status: storkv1.ActionStatusFailed, Reason: msg}
} else {
        msg := fmt.Sprintf("scaling up apps in namespace %s successful", ns)
	failoverSummary = &storkv1.FailoverSummary{Namespace: ns, Status: storkv1.ActionStatusSuccessful, Reason: msg}
}
```

pkg/action/actioncontroller.go
```diff
@@ -121,6 +123,22 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action)
 			resourceutils.ScaleReplicas(action.Namespace, true, ac.printFunc(action, "ScaleReplicas"), ac.config)
 			ac.updateStatus(action, storkv1.ActionStatusSuccessful)
 		}
+	case storkv1.ActionTypeFailover:
+		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
+		switch action.Status.Stage {
+		case storkv1.ActionStageInitial:
+			ac.validationsBeforeFailover(action)
```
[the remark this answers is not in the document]
olavangad-px: okay will take care in the revision

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+	"github.com/libopenstorage/stork/pkg/resourceutils"
+	"github.com/libopenstorage/stork/pkg/utils"
+	coreops "github.com/portworx/sched-ops/k8s/core"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+// validationsBeforeFailover This method is called as part of the Initial stage of Failover
```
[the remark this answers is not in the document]
olavangad-px: will address in the revision

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+func (ac *ActionController) validationsBeforeFailover(action *storkv1.Action) {
+	// TODO: source cluster-domain deactivation in sync-dr case
+	if action.Status.Status == storkv1.ActionStatusSuccessful {
+		// Move to ScaleUpDestination stage directly in case skipDeactivateSource is true
+		if *action.Spec.ActionParameter.FailoverParameter.SkipDeactivateSource {
```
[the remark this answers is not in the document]
olavangad-px: Already addressed as part of another comment. Will address skipSourceOperations related changes along with sync-dr specific operations in a separate PR

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+		logEvents := ac.printFunc(action, string(storkv1.ActionStatusFailed))
+		logEvents(msg, "err")
+		action.Status.Status = storkv1.ActionStatusFailed
+		action.Status.Reason = msg
+		ac.updateAction(action)
+		return
+	}
+	srcMigrSched, err := remoteOps.GetMigrationSchedule(action.Spec.ActionParameter.FailoverParameter.MigrationScheduleReference, action.Namespace)
+	if err != nil {
+		msg := fmt.Sprintf("Unable to find MigrationSchedule %s/%s for failover. Skipping MigrationSchedule suspension : %v", migrationSchedule.Namespace, migrationSchedule.Name, err)
```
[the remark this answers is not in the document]
olavangad-px: Will address. 
Forgot to recheck and format all the messages/logs/errors as I was in a hurry to raise the PR.

pkg/action/failover.go
```diff
@@ -0,0 +1,384 @@
...
+		logEvents(msg, "err")
+		action.Status.Status = storkv1.ActionStatusFailed
+		action.Status.Reason = msg
+		ac.updateAction(action)
+		return
+	}
+	// get sourceConfig from clusterPair in the destination cluster
+	sourceConfig, err := getClusterPairSchedulerConfig(migrationSchedule.Spec.Template.Spec.ClusterPair, migrationSchedule.Namespace)
+	if err != nil {
+		msg := fmt.Sprintf("Error while getting the clusterpair %s/%s for failover", migrationSchedule.Namespace, migrationSchedule.Spec.Template.Spec.ClusterPair)
```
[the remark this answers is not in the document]
olavangad-px: Will address

pkg/action/failover.go
olavangad-px: Yes it is indeed a possibility as per current code. 
Should we then block the failover instead in last migration failed scenarios ?? 
I guess this will be a problem only in controlled failover cases where the customer can still try and debug the migration failure because in disaster situation if we assume that source is inaccessible anyways these checks are skipped and scale up is done on destination.

pkg/action/failover.go
olavangad-px: I had kind o followed what was written in the failback side. But on a second thought even I am wondering that it would make sense to atleast let the user know if reactivation failed, which namespaces were not reactivated. So that they can go ahead and try to activate those namespaces manually.
@diptiranjanpx what do you think about this ?

diptiranjanpx: The reactivate (essentially rollback) will only happen if something from previous stage has failed , so irrespective of the rollback stage's status, we have to mark the status as failed.

We can mark the reactivate stage's status `Failed` if something fails there. Anyway we will go to next stage irrespective of this stage succeeds or fails. 

We can raise the event saying which all reactivations failed but let's not put that in the reason field as it will have the info of original error.

diptiranjanpx: Also you can fill up the each namespace's status and reason in case of reactivation. That way at least we can have both the original failed reason and activation failure in the CR too.

pkg/action/failover.go
olavangad-px: Have the changes for point 2 ready and will raise that PR once this PR is merged.
Will work on 1 and 3 early next week and raise the PR asap

pkg/action/failover.go
```diff
@@ -0,0 +1,364 @@
...
+	} else {
+		msg := fmt.Sprintf("Scaling up of applications in cluster : %s failed.", config.Host)
+		logEvents := ac.printFunc(action, string(storkv1.ActionStatusFailed))
+		logEvents(msg, "out")
+		action.Status.Status = storkv1.ActionStatusFailed
+	}
+	ac.updateAction(action)
+}
+
+func (ac *ActionController) isClusterAccessible(action *storkv1.Action, config *rest.Config) bool {
```
diptiranjanpx, line 336: We can move this to utils (not necessarily in this PR).

diptiranjanpx (approved):

PR #1667, opened 2024-03-13
[PWX-32203] Adding perform failback command and unit tests

olavangad-px:
**What type of PR is this?**
>feature
>unit-test

Yes, adds a new storkctl command `storkctl perform failback`

```
[root@ip-10-13-194-37 ~]# storkctl perform failback -h
Initiate failback of the given migration schedule

Usage:
  storkctl perform failback [flags]

Flags:
      --exclude-namespaces strings   Specify the comma-separated list of subset namespaces to be skipped during the failback. By default, all namespaces part of the MigrationSchedule are failed back
  -h, --help                         help for failback
      --include-namespaces strings   Specify the comma-separated list of subset namespaces to be failed back. By default, all namespaces part of the MigrationSchedule are failed back
  -m, --migration-reference string   Specify the MigrationSchedule to failback. Also specify the namespace of this MigrationSchedule using the -n flag

```

No

24.2.0

**Testing Details**:
Have added comprehensive unit tests and tested all combinations manually as well

### DIFF
--- a/pkg/storkctl/action.go
+++ b/pkg/storkctl/action.go
@@ -94,7 +94,7 @@ func newFailbackCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
 	var excludeNamespaceList []string
 	performFailbackCommand := &cobra.Command{
 		Use:   failbackCommand,
-		Short: "Initiate failover of the given migration schedule",
+		Short: "Initiate failback of the given migration schedule",
 		Run: func(c *cobra.Command, args []string) {
 			// namespace of the MigrationSchedule is provided by the user using the -n / --namespace global flag
 			migrationScheduleNs := cmdFactory.GetNamespace()
--- a/pkg/storkctl/action.go
+++ b/pkg/storkctl/action.go
@@ -17,6 +17,7 @@ import (
 
 const (
 	failoverCommand                    = "failover"
+	failbackCommand                    = "failback"
 	nameTimeSuffixFormat string        = "2006-01-02-150405"
 	actionWaitTimeout    time.Duration = 10 * time.Minute
 	actionWaitInterval   time.Duration = 10 * time.Second
@@ -30,6 +31,7 @@ func newPerformCommand(cmdFactory Factory, ioStreams genericclioptions.IOStreams
 
 	performCommands.AddCommand(
 		newFailoverCommand(cmdFactory, ioStreams),
+		newFailbackCommand(cmdFactory, ioStreams),
 	)
 	return performCommands
 }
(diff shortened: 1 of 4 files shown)


pkg/storkctl/action.go
```diff
@@ -54,79 +56,29 @@ func newFailoverCommand(cmdFactory Factory, ioStreams genericclioptions.IOStream
...
+			// namespace of the MigrationSchedule is provided by the user using the -n / --namespace global flag
 			migrationScheduleNs := cmdFactory.GetNamespace()
-			migrSchedObj, err := storkops.Instance().GetMigrationSchedule(referenceMigrationSchedule, migrationScheduleNs)
+			namespaceList, err := validationsForPerformDRCommands(storkv1.ActionTypeFailover, migrationScheduleNs, referenceMigrationSchedule, includeNamespaceList, excludeNamespaceList)
 			if err != nil {
-				util.CheckErr(fmt.Errorf("unable to find the reference MigrationSchedule %v in the %v namespace", referenceMigrationSchedule, migrationScheduleNs))
+				util.CheckErr(err)
 				return
 			}
-			if !skipDeactivateSource {
```
olavangad-px, line 73: One change I have made in the failover command is that clusterPair is required irrespective of skipDeactivateSource value. This is because even in skipDeactivateSource we would want to suspend the migrationSchedule in case source cluster is accessible

diptiranjanpx: But for unidirectional clusterpair and source not accessible, we still have to do the failover.

olavangad-px: But I cannot determine if source is accessible or not if clusterpair is not there ..
Maybe then the call to take is that if clusterpair not found in failover case we will assume source is inaccessible and will skip migrationSchedule suspension and source deactivation.
Or like we were discussing we can document that if unidirectional they should create reverse clusterpair with same name. And make having the cluster pair a pre-requisite

olavangad-px: As discussed offline we will be relaxing this check for failover.
If clusterpair is not provided we will just log a warning event and directly go ahead with scaleUpDestination. So in failover CLI we won't enforce anything wrt clusterpair

olavangad-px: We will need to discuss this in a meeting with rest of the team and finalise by end of this week. For now let's enforce it. I will make the change in another PR if required.

diptiranjanpx (approved):

pp511 (approved):
Let's discuss more offline regarding if we should enforce cp on the destination. These changes look ok for now.

PR #1669, opened 2024-03-13
PWX-36282: Improve the GetPodVolumes() implementation for the Portworx Driver.

adityadani:
**What type of PR is this?**

>improvement

- Instead of sending individual Inspect() requests for each volume in a pod send a single Enumerate() request with all volumes in it. This is also less intensive on the Portworx side.
- Modify the Transform function in the SharedInformerCache to only store pod volumes which are relevant for stork.
- Added a framework to unit test the Portworx driver. Added UTs for GetPodVolumes() implementation.
- Modify the health monitor integration test to use mysql spec that has two volumes.

No

```release-note
Improvement: The Portworx driver has optimized the API calls it makes to reduce the time taken for scheduling pods and monitoring pods if they need to be rescheduled if Portworx is down on nodes. 
```

Yes - 24.2.0

### DIFF
--- a/drivers/volume/portworx/portworx.go
+++ b/drivers/volume/portworx/portworx.go
@@ -9,7 +9,6 @@ import (
 	"math"
 	"os"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
@@ -252,15 +251,16 @@ var restoreStateCallBackoff = wait.Backoff{
 }
 
 type portworx struct {
-	store           cache.Store
-	stopChannel     chan struct{}
-	sdkConn         *portworxGrpcConnection
-	id              string
-	endpoint        string
-	tlsConfig       *tls.Config
-	jwtSharedSecret string
-	jwtIssuer       string
-	initDone        bool
+	store             cache.Store
+	stopChannel       chan struct{}
+	sdkConn           *portworxGrpcConnection
+	id                string
+	endpoint          string
+	tlsConfig         *tls.Config
+	jwtSharedSecret   string
+	jwtIssuer         string
+	initDone          bool
+	getAdminVolDriver func() (volume.VolumeDriver, error)
 }
 
 type portworxGrpcConnection struct {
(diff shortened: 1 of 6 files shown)


drivers/volume/portworx/portworx.go
```diff
@@ -506,35 +506,76 @@ func (p *portworx) inspectVolume(volDriver volume.VolumeDriver, volumeID string)
...
+	vols, err := volDriver.Enumerate(
+		&api.VolumeLocator{
+			VolumeIds: volumeNameList,
+		},
+		nil,
+	)
+	if err != nil {
+		return nil, err
+	}
+	for _, volName := range volumeNameList {
```
diptiranjanpx, line 565: Would it have been better to initialize the volMap with only the storkVol in a loop over volumeNameList and then filled the map with more info looping over vols. This way we can avoid the nested loop.

drivers/volume/portworx/portworx.go
```diff
@@ -835,6 +878,18 @@ func (p *portworx) GetPodVolumes(podSpec *v1.PodSpec, namespace string, includeP
 			if pvc.Status.Phase == v1.ClaimPending {
 				// Only include pending volume if requested and storage class has WFFC
 				if includePendingWFFC && isWaitingForFirstConsumer(pvc) {
+					// For pending volumes we won't query Portworx.
+					// Populate at lease something about these volumes
```
pp511, line 882: nit.nit.. least

drivers/volume/portworx/portworx.go
```diff
@@ -506,35 +506,76 @@ func (p *portworx) inspectVolume(volDriver volume.VolumeDriver, volumeID string)
...
+		storkVol := &storkvolume.Info{
+			VolumeID: volName,
+		}
+		for _, vol := range vols {
+			if vol.Locator.Name == volName || vol.Id == volName {
+				storkVol = p.constructStorkVolume(vol)
+				break
+			}
+		}
+		volMap[volName] = storkVol
```
pp511, line 575: Do we want to return volumes which are not in the volDriver.Enumerate() list? I think it makes sense to move this call to just before break.

adityadani: I want to keep the behavior same as before, where we were always returning a storkvolume.Info object regardless of whether we got a successful API response for that volume from Portworx.

drivers/volume/portworx/portworx.go
adityadani: Fixed in follow up commit.

diptiranjanpx (approved):

pp511 (approved):
looks good

PR #1671, opened 2024-03-15
PB-6115 Handling label validation for Data Export CR

vikasit12:
- if the StorageClass name has more than 63 characters we fail to deploy the corresponding pod
- use annotation for storing SC name in DataExport CR

**What type of PR is this?**
>bug

This PR handles SC names with more than 64 char and succeeds the backups.

no

no

release-24.1.0

**Test Screenshots**:
### Before Fix 
image
image
### After Fix
image
image


vsundarraj-px (approved):
Lgtm

vsundarraj-px:
Oops looks like there is issue with this commit. The vendored code is also udpated without changes to go.mod

PR #1676, opened 2024-03-20
PWX-35982: Storkctl create clusterpair will support migration and onetime-migration as mode.

diptiranjanpx:
Also DisasterRecovery clusterpair mode gets added for async DR case in the clusterpair object.

**What type of PR is this?**
bug

storkctl create clusterpair now will support multiple modes like `async-dr, sync-dr, migration, onetime-migration`

**Unit test**:
```
➜  stork git:(PWX-35982) ✗ go test -v -tags unittest -coverprofile=profile.out -covermode=atomic  github.com/libopenstorage/stork/pkg/storkctl -run TestCreateUniDirectionalClusterPairMissingParameters
=== RUN   TestCreateUniDirectionalClusterPairMissingParameters
--- PASS: TestCreateUniDirectionalClusterPairMissingParameters (0.01s)
PASS
coverage: 14.5% of statements
ok      github.com/libopenstorage/stork/pkg/storkctl    3.558s  coverage: 14.5% of statements

➜  stork git:(PWX-35982) ✗ go test -v -tags unittest -coverprofile=profile.out -covermode=atomic  github.com/libopenstorage/stork/pkg/storkctl -run TestGenerateClusterPairForDifferentModes            
=== RUN   TestGenerateClusterPairForDifferentModes
--- PASS: TestGenerateClusterPairForDifferentModes (0.01s)
PASS
coverage: 1.1% of statements
ok      github.com/libopenstorage/stork/pkg/storkctl    3.233s  coverage: 1.1% of statements
```

**Function tests**

All scenarios with tests have been updated in PWX-35982

### DIFF
--- a/pkg/storkctl/clusterpair.go
+++ b/pkg/storkctl/clusterpair.go
@@ -39,6 +39,14 @@ const (
 	pxAdminTokenSecret     = "px-admin-token"
 	secretNamespace        = "openstorage.io/auth-secret-namespace"
 	secretName             = "openstorage.io/auth-secret-name"
+
+	userInputCPModeAsync            = "async-dr"
+	userInputCPModeSync             = "sync-dr"
+	userInputCPModeMigration        = "migration"
+	userInputCPModeOnetimeMigration = "onetime-migration"
+
+	clusterPairDRModeDisasterRecovery = "DisasterRecovery"
+	clusterPairDRModeOnetimeMigration = "OneTimeMigration"
 )
 
 var (
@@ -283,7 +291,12 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 				return
 			}
 
-			if mode == "sync-dr" {
+			if mode != userInputCPModeAsync && mode != userInputCPModeSync && mode != userInputCPModeMigration && mode != userInputCPModeOnetimeMigration {
+				util.CheckErr(fmt.Errorf("invalid mode %s, mode value should either be %s, %s, %s or %s", mode, userInputCPModeAsync, userInputCPModeSync, userInputCPModeMigration, userInputCPModeOnetimeMigration))
+				return
+			}
+
+			if mode == userInputCPModeSync {
 				syncDR = true
 			}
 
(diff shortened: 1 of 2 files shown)


olavangad-px (approved):
LGTM.

pkg/storkctl/clusterpair.go
```diff
@@ -283,7 +290,12 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
 				return
 			}
 
-			if mode == "sync-dr" {
+			if mode != userInputCPModeAsync && mode != userInputCPModeSync && mode != userInputCPModeMigration {
+				util.CheckErr(fmt.Errorf("invalid mode %s, mode value should either be %s, %s or %s", mode, userInputCPModeAsync, userInputCPModeSync, userInputCPModeMigration))
```
pp511, line 294: We will need to update the docs saying that now we need to pass on the mode explicitly? Can you file a PD for this?

pkg/storkctl/clusterpair.go
```diff
@@ -699,14 +711,30 @@ func newCreateClusterPairCommand(cmdFactory Factory, ioStreams genericclioptions
...
+	mode string,
+	ignoreStorageOptions bool) (*storkv1.ClusterPair, error) {
 	opts := make(map[string]string)
 	if !ignoreStorageOptions {
 		opts["ip"] = ip
 		opts["port"] = port
 		// extract token from px-endpoint command
 		opts["token"] = token
 	}
+	if mode == userInputCPModeAsync {
```
pp511, line 733: ignoreStorageOptions is true in case of sync. But shouldn't we also show mode in the CP in that case just for it to be obvious?

pp511 (approved):
Looks good. Just answer my 2 qs.

pkg/storkctl/clusterpair.go
diptiranjanpx: We are relying on that information to decide whether the clusterpair is for sync and async cluster. I think it's better to stick with that. 

I agree it's good to have  that value but because `options` key is assumed to be empty for sync cluster, we will break that behavior.

pkg/storkctl/clusterpair.go
diptiranjanpx: Till now it is not explicitly required, the default value for user input mode is `async-dr`.

PR #1680, opened 2024-03-22
Vendoring 1.2.11 kdmp to stork 24.1.0 .

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
Vendoring

Vendored kdmp 1.2.11 .

no

no

no

### DIFF
--- a/vendor/github.com/portworx/kdmp/pkg/controllers/dataexport/reconcile.go
+++ b/vendor/github.com/portworx/kdmp/pkg/controllers/dataexport/reconcile.go
@@ -267,7 +267,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		var compressionType string
 		var podDataPath string
 		var excludeFileList string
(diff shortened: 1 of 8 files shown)


lalat-das (approved):
lgtm

vendor/github.com/portworx/kdmp/pkg/controllers/dataexport/reconcile.go
```diff
@@ -267,7 +267,7 @@ func (c *Controller) sync(ctx context.Context, in *kdmpapi.DataExport) (bool, er
 		var compressionType string
 		var podDataPath string
 		var excludeFileList string
-		pvcStorageClass := dataExport.Annotations[kdmpStorageClassKey]
+		pvcStorageClass := dataExport.Labels[kdmpStorageClassKey]
```
lalat-das, line 270: This seem to be an unrelated change came because @vikasit12 has merged an chnage directly to the vendor file of stork. It should come via kdmp vendoring to stork. 
It seems @pallav-px  is already raising a PR to correct this up ,so that vendoring will change this file.

PR #1695, opened 2024-03-29
PWX-36485 : action controller dr changes wrt metro-dr

olavangad-px:
**What type of PR is this?**
>feature

Make changes wrt metro-dr in failover and failback.

No

No

Yes, 24.2.0

### DIFF
--- a/pkg/action/drutils.go
+++ b/pkg/action/drutils.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"fmt"
+	
 	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/log"
 	"github.com/libopenstorage/stork/pkg/utils"
--- a/pkg/action/drutils.go
+++ b/pkg/action/drutils.go
@@ -143,18 +143,18 @@ func (ac *ActionController) remoteClusterDomainUpdate(activate bool, action *sto
 		if activate {
 			err := ac.volDriver.ActivateClusterDomain(clusterDomainUpdate)
 			if err != nil {
-				return fmt.Errorf("activation of remote cluster domain: %v failed : %v", remoteDomainName, err)
+				return fmt.Errorf("activation of cluster domain: %v failed : %v", remoteDomainName, err)
 			} else {
-				msg := fmt.Sprintf("Activation of remote cluster domain: %v successful", remoteDomainName)
+				msg := fmt.Sprintf("Activation of cluster domain: %v successful", remoteDomainName)
 				logEvents := ac.printFunc(action, "ActivateClusterDomain")
 				logEvents(msg, "out")
 			}
 		} else {
 			err := ac.volDriver.DeactivateClusterDomain(clusterDomainUpdate)
 			if err != nil {
-				return fmt.Errorf("deactivation of remote cluster domain: %v failed : %v", remoteDomainName, err)
+				return fmt.Errorf("deactivation of cluster domain: %v failed : %v", remoteDomainName, err)
 			} else {
-				msg := fmt.Sprintf("Deactivation of remote cluster domain: %v successful", remoteDomainName)
+				msg := fmt.Sprintf("Deactivation of cluster domain: %v successful", remoteDomainName)
 				logEvents := ac.printFunc(action, "DeactivateClusterDomain")
 				logEvents(msg, "out")
 			}
--- a/pkg/action/drutils.go
+++ b/pkg/action/drutils.go
@@ -2,7 +2,7 @@ package action
 
 import (
 	"fmt"
-	
+
 	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/action/failback.go
```diff
@@ -274,90 +287,9 @@ func (ac *ActionController) performLastMileMigration(action *storkv1.Action) {
...
-		ac.updateAction(action)
-		return
-	}
-	log.ActionLog(action).Infof("Last mile migration %s is still running, currently in stage: %s", migrationObject.Name, migrationObject.Status.Stage)
+	// In failback last-mile-migration is created in destination->source direction
+	config := ac.config
+	ac.createLastMileMigration(action, config, migrationSchedule)
 }
 
 // getLastMileMigrationSpec will get a migration spec from the given migrationschedule spec
```
olavangad-px, line 295: I will move the other methods common to both failover/failback such as activateClusterDuringDR, deactivateClusterDuringDR, getLastMileMigratioonSpec etc. to the drutils file in another PR.
Didn't do that now to make review easier

pkg/action/drutils.go
```diff
@@ -0,0 +1,163 @@
...
+	"fmt"
+	storkv1 "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
+	"github.com/libopenstorage/stork/pkg/log"
+	"github.com/libopenstorage/stork/pkg/utils"
+	"github.com/pborman/uuid"
+	storkops "github.com/portworx/sched-ops/k8s/stork"
+	v1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
```
pp511, line 13: nit..Let's separate the imports in two parts.

pkg/action/failover.go
```diff
@@ -66,6 +65,17 @@ func (ac *ActionController) validateBeforeFailover(action *storkv1.Action) {
...
+			// if clusterPair is absent in destination cluster and skipSourceOperations is provided always move directly to ScaleUpDestination
+			msg := "Skipping source cluster operations and moving to the Failover ScaleUpDestinationStage since SkipSourceOperations is set to true"
+			logEvents := ac.printFunc(action, string(storkv1.ActionStatusSuccessful))
+			logEvents(msg, "out")
+			action.Status.Stage = storkv1.ActionStageScaleUpDestination
+			action.Status.Status = storkv1.ActionStatusInitial
+			action.Status.FinishTimestamp = metav1.Now()
+			ac.updateAction(action)
+			return
+		}
```
pp511, line 78: If SkipSourceOperations & err == nil, do we want to go ahead and execute the rest of this function even if it's async-dr? Or should we make this check before SkipSourceOperations and return from there?

pkg/action/drutils.go
olavangad-px: will address

pkg/action/failover.go
olavangad-px: Addressed. if async && skipSourceOperations we do not need to execute rest of the function

pp511 (approved):
lgtm.

pkg/action/failover.go
```diff
@@ -400,3 +410,89 @@ func (ac *ActionController) createSummary(action *storkv1.Action, ns string, sta
...
+		action.Status.Stage = storkv1.ActionStageScaleUpSource
+		action.Status.Status = storkv1.ActionStatusInitial
+		action.Status.FinishTimestamp = metav1.Now()
+		ac.updateAction(action)
+		return
+	}
+
+	migrationScheduleReference := action.Spec.ActionParameter.FailoverParameter.MigrationScheduleReference
+
+	migrationSchedule, err := storkops.Instance().GetMigrationSchedule(migrationScheduleReference, action.Namespace)
```
diptiranjanpx, line 432: Are we not doing the static copy check or because it has crossed the validation check we are ignoring it?

pkg/action/drutils.go
```diff
@@ -0,0 +1,164 @@
...
+			},
+			Spec: storkv1.ClusterDomainUpdateSpec{
+				ClusterDomain: remoteDomainName,
+				Active:        activate,
+			},
+		}
+		if activate {
+			err := ac.volDriver.ActivateClusterDomain(clusterDomainUpdate)
+			if err != nil {
+				return fmt.Errorf("activation of remote cluster domain: %v failed : %v", remoteDomainName, err)
```
diptiranjanpx, line 146: Let's remove "remote" from here since we are always activating/deactivating the source.

pkg/action/failover.go
olavangad-px: Yes, that has been already covered as part of the validations in InitialStage. So, I didn't feel the need to repeat that check

pkg/action/drutils.go
olavangad-px: okay, will address

diptiranjanpx (approved):

PR #1697, opened 2024-03-31
pb-6459: Excluding the pods that are in completed state, even if the pod selecor label matches, to run the rules.

siva-portworx:
**What type of PR is this?**
>bug

```
    pb-6459: Excluding the pods that are in completed state, even if the pod
    selector label matches, to run the rules.
```

no

```release-note
Issue: In some use-cases, completed pods are getting selected to run and execution of rule fails.
User Impact: If there are two matching pods, one in completed and another in running state, makes the rule execution to fail.
Resolution: Avoiding the pods selection to run the rules, if they are in completed state.

```

release-24.1.0 branch.

**Testing:**
Tested the fix on the QA setup as mentioned in the ticket RCA.

1. When we migrate a VM (say from one node1 to node2) to another, the virt-launcher pod goes to Completed state in the current node (node1)
2. New virt-laucher pod is spawned in the new node (node2) which is up and running. Now this controls the VM ops
3. What we are doing in our backup flow is that, we are executing the default freeze rules on the pods matching the pod selector label kubevirt.io/domain=vm-name
4. But now there are 2 pods in the namespace with the same label - one is in Running state and other other is in Completed state
5. With the fix, it excluded the pod in completed state and backup completed.

### DIFF
--- a/pkg/rule/rule.go
+++ b/pkg/rule/rule.go
@@ -245,15 +245,20 @@ func ExecuteRule(
 		err = fmt.Errorf("failed to generate uuid for rule tasks due to: %v", err)
 		return nil, err
 	}
-
 	pods := make([]v1.Pod, 0)
 	for _, item := range rule.Rules {
 		p, err := core.Instance().GetPods(podNamespace, item.PodSelector)
 		if err != nil {
 			return nil, err
 		}
-
-		pods = append(pods, p.Items...)
+		for _, pod := range p.Items {
+			if core.Instance().IsPodCompleted(pod) {
+				logrus.Infof("Ignoring completed pod for running rule %v", pod.Name)
+				continue
+			}
+			logrus.Debugf("Including pod to run rule %v", pod.Name)
+			pods = append(pods, pod)
+		}
 	}
 
 	if len(pods) > 0 {
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20210112165513-ebc401615f47
 	github.com/pborman/uuid v1.2.1
 	github.com/portworx/px-object-controller v0.0.0-20220804234424-40d3b8a84987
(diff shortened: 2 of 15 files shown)


prashanthpx (approved):

pkg/rule/rule.go
```diff
@@ -245,15 +245,20 @@ func ExecuteRule(
...
 		if err != nil {
 			return nil, err
 		}
-
-		pods = append(pods, p.Items...)
+		for _, pod := range p.Items {
+			if !core.Instance().IsPodCompleted(pod) {
+				logrus.Infof("Including pod to run rule %v", pod.Name)
+				pods = append(pods, pod)
+			} else {
```
vsundarraj-px, line 258: nit: We could avoid else with 
if !core.Instance().IsPodCompleted(pod) { 
   logrus.Infof("Ignoring completed pod for running rule %v/%v", pod.Name, pod.Namespace)
   continue
}
logrus.Debugf("Incuding pod to run rule %v/%v", pod.Name, pod.Namespace)
pods = append(pods, pod)

pkg/rule/rule.go
```diff
@@ -245,15 +245,20 @@ func ExecuteRule(
...
 	for _, item := range rule.Rules {
 		p, err := core.Instance().GetPods(podNamespace, item.PodSelector)
 		if err != nil {
 			return nil, err
 		}
-
-		pods = append(pods, p.Items...)
+		for _, pod := range p.Items {
+			if !core.Instance().IsPodCompleted(pod) {
+				logrus.Infof("Including pod to run rule %v", pod.Name)
```
vsundarraj-px, line 256: Do we need this to be infof? In my opinion its best to keep debugf for include list.

vsundarraj-px (approved):
nit

pkg/rule/rule.go
siva-portworx: Done.

pkg/rule/rule.go
siva-portworx: Done

PR #1708, opened 2024-04-10
increase timeout for VM images download

Rohit-PX:
**What type of PR is this?**
integration-test

* The kubevirt functional test jobs are failing with timeout error when downloading images for VM templates
* Eg. https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%20Cloud%20Jobs/job/stork-ocp-kubevirt-4-13/96/consoleFull
* Increase the timeout when downloading images for VM templates
* Increase the retry interval for the same to reduce redundant logs

no

no

no

### DIFF
--- a/test/integration_test/kubevirt_test.go
+++ b/test/integration_test/kubevirt_test.go
@@ -27,8 +27,8 @@ var templatePVCSpecs = map[string]string{
 const (
 	importerPodPrefix            = "importer"
 	importerPodStartTimeout      = 2 * time.Minute
-	importerPodCompletionTimeout = 20 * time.Minute
-	importerPodRetryInterval     = 10 * time.Second
+	importerPodCompletionTimeout = 60 * time.Minute
+	importerPodRetryInterval     = 30 * time.Second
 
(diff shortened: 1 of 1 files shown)


pp511 (approved):
lgtm

test/integration_test/kubevirt_test.go
```diff
@@ -27,8 +27,8 @@ var templatePVCSpecs = map[string]string{
 const (
 	importerPodPrefix            = "importer"
 	importerPodStartTimeout      = 2 * time.Minute
-	importerPodCompletionTimeout = 20 * time.Minute
-	importerPodRetryInterval     = 10 * time.Second
+	importerPodCompletionTimeout = 30 * time.Minute
```
pureneelesh, line 30: we can increase it to even higher value of 60 min. In general, the timeout value should be such that there is absolutely no point in waiting beyond that time.

pureneelesh (approved):

test/integration_test/kubevirt_test.go
Rohit-PX: Fixed.

PR #1717, opened 2024-04-15
add new VM spec for datavolume based root and data-disk, without WFFC

Rohit-PX:
**What type of PR is this?**
integration-test

* Adding a new VM spec that uses only datavolume based root and datadisks, that don't use WFFC
* This way co-location will always be achieved

no

no

no

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -1729,7 +1729,7 @@ func addTestModeEnvironmentVar() error {
 
 func IsCloud() bool {
 	stc, err := operator.Instance().ListStorageClusters(defaultAdminNamespace)
-	if err == nil {
+	if len(stc.Items) > 0 && err == nil {
 		log.InfoD("Storage cluster name: %s", stc.Items[0].Name)
 		if len(stc.Items) > 0 {
 			if oputils.IsEKS(&stc.Items[0]) {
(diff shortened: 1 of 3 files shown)


test/integration_test/common_test.go
```diff
@@ -1729,7 +1729,7 @@ func addTestModeEnvironmentVar() error {
 
 func IsCloud() bool {
 	stc, err := operator.Instance().ListStorageClusters(defaultAdminNamespace)
-	if err == nil {
+	if len(stc.Items) > 0 && err == nil {
```
pureneelesh, line 1732: remove the same check on line 1734?

pureneelesh (approved):

test/integration_test/specs/kubevirt-fedora-multiple-disks-datavol-only/fedora-vm-multidisk-datavol-only.yaml
```diff
@@ -0,0 +1,95 @@
...
+            chpasswd: { expire: False }
+        name: cloudinitdisk
+  dataVolumeTemplates:
+  - metadata:
+      name: fedora-root-disk-datavol-only
+    spec:
+      source:
+        pvc:
+          name: fedora-template-pvc
+          namespace: openshift-virtualization-os-images
```
pureneelesh, line 72: Can we add another datavolume with "blank" source?
```
blank: {}
```

Rohit-PX: Fixed.

Rohit-PX:
```
2024/04/16 02:37:28 WARN: Skipping testrail update for this case, testID: 93196, testrun: 0
--- PASS: TestKubevirt (2142.81s)
    --- PASS: TestKubevirt/kubeVirtHypercOneLiveMigration (2139.31s)
PASS
```

PR #1718, opened 2024-04-16
[PWX-36891] Updating ApplicationActivated to true/false in the migrationschedule 

olavangad-px:
**What type of PR is this?**
>improvement

Update the ApplicationActivated status in all the relevant migrationschedules when scaling up or scaling down namespaces during failover / failback.
Here relevant migrationschedules are those which migrate resources from atleast one of the namespaces being failed over or failed back.

No

No

Yes, 24.2.0

**Testing Details**:
Have manually verified that only the relevant migrationSchedules are picked up and updated.

### DIFF
--- a/pkg/action/drutils.go
+++ b/pkg/action/drutils.go
@@ -171,8 +171,7 @@ func (ac *ActionController) updateApplicationActivatedInRelevantMigrationSchedul
 	// relevant migrationSchedules are ones which are involved in migration of resources from at least one of the namespaces being activated.
 	// such migrationSchedules can reside in any of the activationNamespaces or in the adminNamespace
 	adminNs := utils.GetAdminNamespace()
-	migrationScheduleNamespaces := make([]string, 0)
-	migrationScheduleNamespaces = activationNamespaces
+	migrationScheduleNamespaces := activationNamespaces
 	if !slices.Contains(activationNamespaces, adminNs) {
 		migrationScheduleNamespaces = append(migrationScheduleNamespaces, adminNs)
 	}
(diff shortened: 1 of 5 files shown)


pkg/action/drutils.go
```diff
@@ -162,3 +162,62 @@ func (ac *ActionController) remoteClusterDomainUpdate(activate bool, action *sto
...
+			msg := fmt.Sprintf("Failed to fetch the list of migrationSchedules in the namespace %v", ns)
+			logEvents := ac.printFunc(action, "ApplicationActivatedStatus")
+			logEvents(msg, "err")
+			ac.updateAction(action)
+		}
+		migrationSchedules.Items = append(migrationSchedules.Items, migrSchedules.Items...)
+	}
+
+	for _, migrSched := range migrationSchedules.Items {
+		isMigrSchedRelevant, err := utils.DoesMigrationScheduleMigrateNamespaces(migrSched, activationNamespaces)
```
diptiranjanpx, line 202: Let's check if the migrationschedule has the `static copy`annotation.

pkg/utils/utils.go
```diff
@@ -424,3 +424,24 @@ func GetAdminNamespace() string {
 	}
 	return adminNs
 }
+
+func DoesMigrationScheduleMigrateNamespaces(migrationSchedule stork_api.MigrationSchedule, activationNs []string) (bool, error) {
```
diptiranjanpx, line 428: nit: Variable name activatedNSList

pkg/action/drutils.go
olavangad-px: Makes sense since updating ApplicationActivated only makes sense in static copy migration schedule.
Will address

pkg/utils/utils.go
olavangad-px: Will address

diptiranjanpx (approved):

PR #1727, opened 2024-04-19
Pb-6675-1: adds support to use AZURE environment variable in the azure objectstore apis

shkumari-px:
**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

It adds support to use AZURE environment variable in the azure
objectstore apis

- Depending upon the azure environment type (AZURE_CHINA/AZURE_PUBLIC), the corresponding storage domain should be framed so that azure objectstore apis can use them to upload objects to the bucket

```release-note
Issue:
User Impact:
Resolution

```

**Unit Testing**

- vendored these stork changes to px-backup
- Passed the azure env type to backup location cr from px-backup and took a backup on that bl
- the backup tries to upload the object on the cloudChina url, this gives us with the storage domain 

ss
Screenshot 2024-04-19 at 5 02 57 PM

Screenshot 2024-04-19 at 5 31 38 PM

azure_public
Screenshot 2024-04-19 at 5 37 01 PM

Screenshot 2024-04-19 at 5 36 50 PM

successful run of the test case with azure_public account
Screenshot 2024-04-19 at 5 39 00 PM

**After Addressing review comments**
Screenshot 2024-04-23 at 7 12 15 PM

Screenshot 2024-04-23 at 9 03 59 PM

### DIFF
--- a/pkg/objectstore/azure/azure.go
+++ b/pkg/objectstore/azure/azure.go
@@ -4,14 +4,20 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	"github.com/Azure/azure-storage-blob-go/azblob"
+	az_autorest "github.com/Azure/go-autorest/autorest/azure"
 	stork_api "github.com/libopenstorage/stork/pkg/apis/stork/v1alpha1"
 	"github.com/libopenstorage/stork/pkg/objectstore/common"
 	"github.com/sirupsen/logrus"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/azureblob"
+)
+
+const (
+	azureEnvKey = "AZURE_ENVIRONMENT"
 )
 
 func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, error) {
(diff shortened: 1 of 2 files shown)


pkg/objectstore/azure/azure.go
```diff
@@ -24,14 +30,46 @@ func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, e
...
+func getAzureURLSuffix(backupLocation *stork_api.BackupLocation) (string, error) {
+	// Give first preference to environment variable setting.
+	azureEnv := os.Getenv(azureEnvKey)
+	if len(azureEnv) == 0 {
+		// If env variable is not set, check the azure environment vaue from backuplocation CR.
+		if len(backupLocation.Location.AzureConfig.Environment) != 0 {
+			azureEnv = backupLocation.Location.AzureConfig.Environment
+		}
+	}
+	// If both env and CR value are empty, set it default as public type.
```
vsundarraj-px, line 42: move this retirm at line 40 with else {
     return
}

pkg/objectstore/azure/azure.go
```diff
@@ -24,14 +30,46 @@ func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, e
...
+	if len(azureEnv) == 0 {
+		return az_autorest.PublicCloud.StorageEndpointSuffix, nil
+	}
+	azureClientEnv, err := az_autorest.EnvironmentFromName(backupLocation.Location.AzureConfig.Environment)
```
vsundarraj-px, line 46: should you use azuerEnv instead of backupLocation.Location.AzureConfig.Environment?

pkg/objectstore/azure/azure.go
```diff
@@ -24,14 +30,46 @@ func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, e
...
+	if err != nil {
```
vsundarraj-px, line 47: you can avoid this err check as you are already checking this at the line 62 below. You can move the logrus.Errorf() there itself. too many err checks.

pkg/objectstore/azure/azure.go
prashanthpx: +1

pkg/objectstore/azure/azure.go
```diff
@@ -24,14 +30,46 @@ func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, e
...
 }
 
+func getAzureURLSuffix(backupLocation *stork_api.BackupLocation) (string, error) {
+	// Give first preference to environment variable setting.
+	azureEnv := os.Getenv(azureEnvKey)
+	if len(azureEnv) == 0 {
+		// If env variable is not set, check the azure environment vaue from backuplocation CR.
+		if len(backupLocation.Location.AzureConfig.Environment) != 0 {
+			azureEnv = backupLocation.Location.AzureConfig.Environment
+		}
```
prashanthpx, line 40: Maybe we should have a log messgae which says where the Endpoint is picked up from. Either Env or BL CR, useful for debugging

pkg/objectstore/azure/azure.go
prashanthpx: if you avoid the error check, our `go errcheck` would fail during compilation. At every place where error is generated, we need to have error check done.

prashanthpx (approved):

pkg/objectstore/azure/azure.go
shkumari-px: Added the log message

pkg/objectstore/azure/azure.go
```diff
@@ -24,14 +30,49 @@ func getPipeline(backupLocation *stork_api.BackupLocation) (pipeline.Pipeline, e
...
+	} else {
+		logrus.Infof("Received azure environment type as environment variable which has higher priority than environment type passed down from backup location cr %s", azureEnv)
+	}
+
+	azureClientEnv, err := az_autorest.EnvironmentFromName(azureEnv)
+	if err != nil {
+		logrus.Errorf("Failed to get azure client env for:%v, err:%v", azureEnv, err)
+		return "", err
+	}
+	// Endpoint suffux based on current cloud type
```
vsundarraj-px, line 54: nit: s/suffux/suffix

vsundarraj-px (approved):

PR #1730, opened 2024-04-22
fix: PB-4394 KDMP restore of a PVC which is completely full fails

vsundarraj-px:
**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

For Direct KDMP if the snapshot taken is larger than the PVC size, then restore will fail with error no space left to write on device. This was handled for CSI based backup/restore by updating the PVC size to match the size of the snapshot.
This PR addresses the same by implementing it in KDMP for direct KDMP case. 

no

```release-note
Issue:  Restore of PVC from backup for which the snapshot size was greater the PVC fails. 
User Impact: Restore failure
Resolution: Fix 

```

release-24.1.0

### DIFF
--- a/drivers/volume/kdmp/kdmp.go
+++ b/drivers/volume/kdmp/kdmp.go
@@ -32,6 +32,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


drivers/volume/kdmp/kdmp.go
```diff
@@ -721,6 +722,14 @@ func (k *kdmp) StartRestore(
 			pvc.Name = bkpvInfo.PersistentVolumeClaim
 			pvc.Namespace = restoreNamespace
 		}
+		// Check if the snapshot size is greated than the pvc size.
```
prashanthpx, line 725: typo: `greated`

prashanthpx (approved):

PR #1732, opened 2024-04-23
[PWX-36989] LastMileMigration should migrate only subset of namespaces

olavangad-px:
**What type of PR is this?**
>bug

Last Mile migration should only migrate the subset of namespaces being failed over / failed back, instead of all the namespaces part of the reference migration schedule.

No

No

Yes, 24.2.0

**Testing Details**:
Have manually tested partial failover and failbacks in sequence. LastMileMigration is picking up only the subset of namespaces being failed over / failed back.
Modified the existing subset of namespaces integration test as well and updated the PR 
https://jenkins.pwx.dev.purestorage.com/job/Users/job/dipti/job/dipti-new-dr-workflow/12/

### DIFF
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -121,6 +121,9 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) 
 			ac.updateStatus(action, storkv1.ActionStatusSuccessful)
 		}
 	case storkv1.ActionTypeFailover:
+		if action.Status.Stage == storkv1.ActionStageFinal {
+			return nil
+		}
 		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
 		switch action.Status.Stage {
 		case storkv1.ActionStageInitial:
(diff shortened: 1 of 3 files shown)


pkg/action/actioncontroller.go
```diff
@@ -121,6 +121,9 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action)
 			ac.updateStatus(action, storkv1.ActionStatusSuccessful)
 		}
 	case storkv1.ActionTypeFailover:
+		if action.Status.Stage == storkv1.ActionStageFinal {
+			return nil
+		}
 		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
```
diptiranjanpx, line 127: Let's just put the action name too in this log.

pkg/action/actioncontroller.go
```diff
@@ -135,10 +138,11 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action)
...
 		case storkv1.ActionStageScaleUpSource:
 			ac.activateClusterDuringFailover(action, true)
-		case storkv1.ActionStageFinal:
-			return nil
 		}
 	case storkv1.ActionTypeFailback:
+		if action.Status.Stage == storkv1.ActionStageFinal {
+			return nil
+		}
 		log.ActionLog(action).Infof("in stage %s", action.Status.Stage)
```
diptiranjanpx, line 146: same as above

diptiranjanpx (approved):

pkg/action/actioncontroller.go
olavangad-px: Everywhere we have used log.ActionLog(action).Infof right.
The actionlog(action) will take care of prefixing every log with action name and namespace
Example, `time="2024-04-22T03:10:10Z" level=info msg="in stage Final" ActionName=failback-reverse-migration-schedule-running-migration-2024-04-21-204419 Namespace=mysql-1-pvc-failback-interval`

pkg/action/actioncontroller.go
olavangad-px: Already taken care. Explained in the below comment

PR #1733, opened 2024-04-29
PB-6687 :: Handle restoring secured px volume

vikasit12:
**What type of PR is this?**
>bug

This PR forces pxd driver to restore using driver.StartRestore irrespective of backuplocation type as we need to support secured portworx volume restore and the access token is known in stork. 

No

No

release-24.1.0

**Tests Screenshots**:
**Before Fix:**
image

**After Fix:**
image
image
image
image

> The extension of this fix and testing details of scenarios are covered in https://github.com/portworx/kdmp/pull/360


prashanthpx (approved):
LGTM

PR #1735, opened 2024-05-02
fix: PB-6598 Scheduled VM backup intermittently takes zero resource backup

vsundarraj-px:
**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

This PR fixes PB-6598. For vm backup we make inmemory objectMap from includeResource. When stork restarts this memory is lost. Hence, we need to re populate this. 

no

```release-note
Issue:
User Impact:
Resolution

```

Unit tested this by restarting stork at various stages of applicationBackup.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -439,6 +439,32 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		return nil
 	}
 
+	// if stork got restarted we would have got IncludeResource Memory map cleaned up.
+	// Hence re-create memory map of it.
+	if IsBackupObjectTypeVirtualMachine(backup) &&
+		backup.Status.Stage != stork_api.ApplicationBackupStageInitial &&
+		backup.Status.Stage != stork_api.ApplicationBackupStageImportResource &&
+		(len(a.vmIncludeResource[string(backup.UID)]) == 0 || len(a.vmIncludeResourceMap[string(backup.UID)]) == 0) {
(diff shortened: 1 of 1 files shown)


siva-portworx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -439,6 +439,32 @@ func (a *ApplicationBackupController) handle(ctx context.Context, backup *stork_
 		return nil
 	}
 
+	// if stork got restarted we would have got IncludeResource Memory map cleandup.
+	// hence re-create memory map of it.
+	if IsBackupObjectTypeVirtualMachine(backup) &&
+		backup.Status.Stage != stork_api.ApplicationBackupStageInitial &&
+		backup.Status.Stage != stork_api.ApplicationBackupStageImportResource &&
```
prashanthpx, line 446: Good to add ApplicationBackupStageFinal stage also as it stork gets restarted here we would end up populating in-memory content which is not needed.

prashanthpx: Do we need them in pre and post exec rule?

prashanthpx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
vsundarraj-px: For final stage its already taken care here 
https://github.com/libopenstorage/stork/blob/1559c07cc48ecb4a56c77d81c3418a67c30ed6b7/pkg/applicationmanager/controllers/applicationbackup.go#L362-L365

We want to ensure that we will do this for all stage except intial and for ImportResourceStage. So this works for preExec stage onwards.

PR #1738, opened 2024-05-05
PB-6687 :: Skipping pvc as per RetainPolicy in case of nfs and pxd

vikasit12:
**What type of PR is this?**
>failing-test

This handles pvc creation as per reatin policy in case of nfs as BL and pxd as driver combination

No

No

release-24.1.0

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -665,6 +665,7 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
 		for driverName, vInfos := range backupVolumeInfoMappings {
 			// Portworx driver restore is not supported via job as it can be a secured px volume
 			// to access the token, we need to run the restore in the same pod as the stork controller
+			// this check is equivalent to (if !nfs || (nfs && driverName == volume.PortworxDriverName))
 			if !nfs || driverName == volume.PortworxDriverName {
 				backupVolInfos := vInfos
 				driver, err := volume.Get(driverName)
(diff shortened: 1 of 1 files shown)


pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -677,7 +677,16 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
 					return err
 				}
 				var preRestoreObjects []runtime.Unstructured
-				if !nfs {
+				if nfs {
+					if restore.Spec.ReplacePolicy == storkapi.ApplicationRestoreReplacePolicyRetain {
```
siva-portworx, line 681: Let us please add comment explaining, why we are doing this. 
Also I prefer to have the check as "if nfs && driverName == volume.PortworxDriverName" to be clear, Even though current check will work.

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -677,7 +677,16 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
...
+						// Skip pv/pvc if replacepolicy is set to retain to avoid creating with empty object
+						backupVolInfos, existingRestoreVolInfos, err = a.skipVolumesFromRestoreList(restore, preRestoreObjects, driver, vInfos)
```
siva-portworx, line 683: let us send the object variable instead of preRestoreObjects even both are nil, to avoid confusion.

prashanthpx (approved):
LGTM

pkg/applicationmanager/controllers/applicationrestore.go
vikasit12: > Let us please add comment explaining, why we are doing this. Also I prefer to have the check as "if nfs && driverName == volume.PortworxDriverName" to be clear, Even though current check will work.

Added this thing as comment

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -676,8 +677,18 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
...
 					return err
 				}
-				var preRestoreObjects []runtime.Unstructured
-				if !nfs {
+				var preRestoreObjects, objects []runtime.Unstructured
+				// this check is equivalent to (if nfs && driverName == volume.PortworxDriverName)
+				if nfs {
+					if restore.Spec.ReplacePolicy == storkapi.ApplicationRestoreReplacePolicyRetain {
+						// Skip pv/pvc if replacepolicy is set to retain to avoid creating with empty object
+						backupVolInfos, existingRestoreVolInfos, err = a.skipVolumesFromRestoreList(restore, objects, driver, vInfos)
```
vsundarraj-px, line 685: pass nil in the place of objects. and there is no need of declaration line at 680 when both preRestoreObjects and objects are only used in else part of the code. They dont need to be declared again and again in the loop.

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -1411,21 +1422,26 @@ func (a *ApplicationRestoreController) skipVolumesFromRestoreList(
...
+			if err != nil {
+				return newVolInfos, existingInfos, err
+			}
+			pvcName = pvcObject.Name
 
-		// get corresponding pvc object from objects list
-		pvcObject, err := volume.GetPVCFromObjects(objects, bkupVolInfo)
-		if err != nil {
-			return newVolInfos, existingInfos, err
+		} else {
```
vsundarraj-px, line 1435: Can you please comment what is selected in case of object is nil? 
How is vInfo.PVC different from volume.GetPVCFromObjects(objects, bkupVolinfo).

vsundarraj-px (commented):
nit: comments. I need to understand bit more for logic. Another pair of eyes will help for logic check.

PR #1741, opened 2024-05-06
[PWX-37142] Removing error log in case update action fails

olavangad-px:
**What type of PR is this?**
>bug
>improvement

In case update Action fails for any reason an error message is logged and we eventually retry. As seen in [migration volumes code path](https://purestorage.atlassian.net/browse/PWX-27015), this kind of logs may cause unnecessary confusion. So going ahead and removing the log.

No

No

Yes, 24.2.0

**Notes for reviewer**:
Please ignore the first commit during review. It's a simple moving of methods to utils file.

### DIFF
--- a/pkg/action/actioncontroller.go
+++ b/pkg/action/actioncontroller.go
@@ -159,7 +159,8 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action) 
 			ac.activateClusterDuringFailback(action, true)
 		}
 	default:
-		ac.updateStatus(action, storkv1.ActionStatusFailed)
+		action.Status.Status = storkv1.ActionStatusFailed
+		ac.updateAction(action)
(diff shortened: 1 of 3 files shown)


pkg/action/actioncontroller.go
```diff
@@ -159,7 +159,8 @@ func (ac *ActionController) handle(ctx context.Context, action *storkv1.Action)
 			ac.activateClusterDuringFailback(action, true)
 		}
 	default:
-		ac.updateStatus(action, storkv1.ActionStatusFailed)
+		action.Status.Status = storkv1.ActionStatusFailed
+		ac.updateAction(action)
```
strivedi-px, line 163: Couldn't make sense of this change. The `updateStatus` is also updating the status of the `action` object right? Or am I missing something @diptiranjanpx?

diptiranjanpx (approved):

PR #1742, opened 2024-05-06
PB-6903: Fix Cloud Native Restore Path

pallav-px:
**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: In case of Old Cloud Native Backups when we restore in 24.1.0 we were facing issue with cloud native backups as they were going as native csi and in resource phase we skip csi based pv and pvc because we apply them later on 

**Does this PR change a user-facing CRD or CLI?**: no 

```release-note
Issue:
User Impact:
Resolution

```

Screenshot 2024-05-06 at 5 17 45 PM

Screenshot 2024-05-06 at 5 17 55 PM

Screenshot 2024-05-06 at 5 18 15 PM

Screenshot 2024-05-06 at 5 18 24 PM

### DIFF
--- a/drivers/volume/volume.go
+++ b/drivers/volume/volume.go
@@ -67,6 +67,16 @@ var (
 		CSIDriverName,
 		KDMPDriverName,
 	}
+
+	orderedListOfDriversForRestore = []string{
+		PortworxDriverName,
+		AWSDriverName,
+		AzureDriverName,
+		GCEDriverName,
+		LinstorDriverName,
+		CSIDriverName,
+		KDMPDriverName,
+	}
 )
 
 // Driver defines an external volume driver interface.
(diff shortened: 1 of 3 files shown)


vikasit12 (approved):

drivers/volume/volume.go
```diff
@@ -399,6 +409,24 @@ func GetPVDriver(pv *v1.PersistentVolume) (string, error) {
 	}
 }
 
+// GetPVDriverForRestore gets the driver associated with a PV. Returns ErrNotFound if the PV is
+// not owned by any available driver
+func GetPVDriverForRestore(pv *v1.PersistentVolume) (string, error) {
+	for _, driverName := range orderedListOfDriversForRestore {
```
lalat-das, line 415: Why similar change is not there in  kdmp/pkg/executor/nfs/nfsrestoreresources.go path ? is there any specific reason. there is a similar definition exist ion NFS path also.

pallav-px: will raise a another PR in KDMP for nfs path

lalat-das (approved):
Add the changes to KDMP repo for NFS backup location scenario else it will fail there. Rest LGTM

drivers/volume/volume.go
```diff
@@ -67,6 +67,16 @@ var (
 		CSIDriverName,
 		KDMPDriverName,
 	}
+
+	orderedListOfDriversForRestore = []string{
+		AWSDriverName,
+		AzureDriverName,
+		GCEDriverName,
+		LinstorDriverName,
+		PortworxDriverName,
```
siva-portworx, line 76: Portworx should on the top.

prashanthpx:
@pallav-px can u cross check once the git commit user u have used? It is showing as "root"

prashanthpx (approved):

vsundarraj-px (approved):

PR #1746, opened 2024-05-08
pb-6910: add uid & gid in each volume for psa support

lalat-das:
**What type of PR is this?** feature
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: When PDSA is enabled and a uid or gid is used for backup using a job pod then each volume need to be storing that so that it can be used later during restore process in destination cluster.

**Does this PR change a user-facing CRD or CLI?**: Yes, but it is in the status field of the CR hence user intervention is not desired here from the perspective of  CR creation or updating ops. We are adding to ApplicationBackup CR's status.

**Is a release note needed?**: Nope

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: Nope

### DIFF
--- a/pkg/apis/stork/v1alpha1/applicationbackup.go
+++ b/pkg/apis/stork/v1alpha1/applicationbackup.go
@@ -93,6 +93,11 @@ type ApplicationBackupResourceInfo struct {
 	ObjectInfo `json:",inline"`
 }
 
+type VolumeJobSecurityContext struct {
+	RunAsUser  int64 `json:"runAsUser"`
+	RunAsGroup int64 `json:"runAsGroup"`
+}
+
 // ApplicationBackupVolumeInfo is the info for the backup of a volume
 type ApplicationBackupVolumeInfo struct {
 	PersistentVolumeClaim    string                      `json:"persistentVolumeClaim"`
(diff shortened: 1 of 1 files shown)


pkg/apis/stork/v1alpha1/applicationbackup.go
```diff
@@ -110,6 +110,8 @@ type ApplicationBackupVolumeInfo struct {
 	StorageClass             string                      `json:"storageClass"`
 	Provisioner              string                      `json:"provisioner"`
 	VolumeSnapshot           string                      `json:"volumeSnapshot"`
+	Uid                      int64                       `json:"uid"`
```
vsundarraj-px, line 113: Could you use a new Structure object to store security Context details. Today its just UID and GID but tomorrow we may end up storing more information like labels.
And since these fields are used only for PSA enabled clusters, for non PSA these are unused, it makes more sense to capsulate the fields into relevant structure.

lalat-das: done

PR #1750, opened 2024-05-10
PWX-36965: functional tests for VM live-migrations during PX update

pureneelesh:
**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
integration-test

Added 2 tests:

- perform PX update with multiple VMs and verify that
  - VMs are evicted from the node before PX pod is deleted
  - VMs stay up during the update

- perform PX update with a VM that can be scheduled only on one of the nodes and verify that
  - the update is resumed only after the VM is stopped
  - FailedToEvictVM is generated

Also, vendor in sched-ops from master.

no

no

no


test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -1074,3 +1242,372 @@ func verifyInitialHyperconvergence(t *testing.T, ctxs []*scheduler.Context, allN
...
+			found = true
+			break
+		}
+	}
+	eventWatcherData.Unlock()
+	Dash.VerifyFatal(t, found, true, fmt.Sprintf("Did not find %s event after %v", reason, afterTime))
+}
+
+func addRemoveTestLabelOnNode(t *testing.T, nodeName string, add bool) {
+	for i := 0; i < 100; i++ {
```
Rohit-PX, line 1554: Any reason for this `for` loop to run 100 times?

test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -357,6 +362,158 @@ func kubeVirtSimulateOCPUpgrade(t *testing.T) {
...
+	checkPXUpdateValidationFailures(t, startTime)
+
+	log.InfoD("Destroying apps")
+	destroyAndWait(t, ctxs)
+	// If we are here then the test has passed
+	testResult = testResultPass
+	log.InfoD("Test status at end of %s test: %s", t.Name(), testResult)
+}
+
+func kubeVirtUpdatePXBlocked(t *testing.T) {
```
Rohit-PX, line 426: Can you add a summary of what this test is trying to do?

Rohit-PX (approved):

PR #1754, opened 2024-05-13
pb-6910: add uid & gid in each volume for psa support

lalat-das:
- If PSA is enabled and a volume is backed-up with a specific Uid/Gid then that value need to be preserved for getting used during restore process.
- This is needed for all the job PODS to inherit above valuec so that no permission denied issues appear while restoring from a file backup

**What type of PR is this?** feature
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: This is needed to support PSA for backup and restore.

**Does this PR change a user-facing CRD or CLI?**: No, it updates the status of Application Backup CR

**Is a release note needed?**: Nope

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: 24.2.2

### DIFF
--- a/pkg/apis/stork/v1alpha1/applicationbackup.go
+++ b/pkg/apis/stork/v1alpha1/applicationbackup.go
@@ -93,6 +93,13 @@ type ApplicationBackupResourceInfo struct {
 	ObjectInfo `json:",inline"`
 }
 
+// This object is used in VolumeInfo for PSA enabled cluster to retain the runAsUser ID and runAsGroup ID used by
+// Job pod(KDMP/NFS)during backup. We will use the same IDs to spin up Job Pods during restore.
+type VolumeJobSecurityContext struct {
+	RunAsUser  int64 `json:"runAsUser"`
+	RunAsGroup int64 `json:"runAsGroup"`
+}
+
 // ApplicationBackupVolumeInfo is the info for the backup of a volume
 type ApplicationBackupVolumeInfo struct {
 	PersistentVolumeClaim    string                      `json:"persistentVolumeClaim"`
@@ -110,6 +117,9 @@ type ApplicationBackupVolumeInfo struct {
 	StorageClass             string                      `json:"storageClass"`
 	Provisioner              string                      `json:"provisioner"`
 	VolumeSnapshot           string                      `json:"volumeSnapshot"`
+	// It preserves the uid and gid of the pod that is run by the backup job
+	// that helps in restore operation. this is required only when PSA is enforced.
+	VolumeJobSecurityContext VolumeJobSecurityContext `json:",inline"`
 }
 
 // ApplicationBackupStatusType is the status of the application backup


lalat-das:
@vsundarraj-px this is the same PR that you have reviewed. Just that it is created against master now so that we can cherry-pick to 24.2.2 whenever we have that ready. Hence will you please review and approve the same.

pkg/apis/stork/v1alpha1/applicationbackup.go
```diff
@@ -110,6 +115,7 @@ type ApplicationBackupVolumeInfo struct {
 	StorageClass             string                      `json:"storageClass"`
 	Provisioner              string                      `json:"provisioner"`
 	VolumeSnapshot           string                      `json:"volumeSnapshot"`
+	VolumeJobSecurityContext VolumeJobSecurityContext    `json:",inline"`
```
vsundarraj-px, line 118: Please add comment here to describe what is this for. These are not very obvious elements that can self explain.

siva-portworx (approved):

dbinnal-px (approved):

pkg/apis/stork/v1alpha1/applicationbackup.go
```diff
@@ -110,6 +116,9 @@ type ApplicationBackupVolumeInfo struct {
 	StorageClass             string                      `json:"storageClass"`
 	Provisioner              string                      `json:"provisioner"`
 	VolumeSnapshot           string                      `json:"volumeSnapshot"`
+	// it preserves the uid and gid of the pod that is run by the backup job
```
vsundarraj-px, line 119: nit: Capital I it

pkg/apis/stork/v1alpha1/applicationbackup.go
```diff
@@ -93,6 +93,12 @@ type ApplicationBackupResourceInfo struct {
 	ObjectInfo `json:",inline"`
 }
 
+// This structure stores metadata related to PSA feature that is required for volume backup/restore
```
vsundarraj-px, line 96: I would have added more specifically who is the consumer 
// This object is used in VolumeInfo for PSA enabled cluster to retain the runAsUser ID and runAsGroup ID used by
//  Job pod(KDMP/NFS)during backup. We will use the same IDs to spin up Job Pods during restore.

vsundarraj-px (approved):
nit: comments provided. But still approved.

pkg/apis/stork/v1alpha1/applicationbackup.go
lalat-das: Done

pkg/apis/stork/v1alpha1/applicationbackup.go
lalat-das: Done as suggested

PR #1760, opened 2024-05-16
pwx-37280 | test: use spec.storageClassName instead of annotation for storageclass

strivedi-px:
**What type of PR is this?**
>failing-test
>integration-test

Addresses [PWX-37280](https://purestorage.atlassian.net/browse/PWX-37280). The PX volume provisioning has been failing in the Bidirectional Migration test job latelya and it was suggested to replace the annotation `volume.beta.kubernetes.io/storage-class: "mysql-sc"` from the PVC and instead use the field `spec.storageClassName : "mysql-sc"`. Post this, the volume provisioning for the pvc is successful. It is better for us to remove the annotation since it is deprecated according to kubernetes [docs](https://kubernetes.io/docs/reference/labels-annotations-taints/#volume-beta-kubernetes-io-storage-class-deprecated).

No

No

`release-24.2.0` branch.

**Successful job run**
Link to the successful job with the changes on the fork: https://jenkins.pwx.dev.purestorage.com/job/Users/job/strivedi/job/strivedi-bidirectional-migration/1/

### DIFF
--- a/test/integration_test/specs/cassandra/cassandra.yaml
+++ b/test/integration_test/specs/cassandra/cassandra.yaml
@@ -91,9 +91,8 @@ spec:
   volumeClaimTemplates:
   - metadata:
       name: cassandra-data
-      annotations:
-        volume.beta.kubernetes.io/storage-class: mysql-sc
     spec:
+      storageClassName: "mysql-sc"
       accessModes: [ "ReadWriteOnce" ]
       resources:
         requests:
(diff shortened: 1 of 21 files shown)


diptiranjanpx:
Let's modify all the specs using storage class with annotation in test/integration_test/specs

diptiranjanpx (approved):

PR #1764, opened 2024-05-16
PB-6976: code to return error if azure env in storage cluster is not mentioned in case of azure china

shkumari-px:
**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: code to return error if azure env in storage cluster is not mentioned in case of azure china

**Does this PR change a user-facing CRD or CLI?**:no

**Is a release note needed?**: no

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: no

Screenshot 2024-05-17 at 12 40 19 AM

Screenshot 2024-05-17 at 12 44 44 AM
Screenshot 2024-05-17 at 1 10 22 AM
Screenshot 2024-05-17 at 2 02 27 AM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	err_pkg "errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -33,6 +34,7 @@ import (
 	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/portworx/sched-ops/k8s/externalsnapshotter"
 	kdmpShedOps "github.com/portworx/sched-ops/k8s/kdmp"
+	operatorOps "github.com/portworx/sched-ops/k8s/operator"
 	"github.com/portworx/sched-ops/k8s/storage"
 	storkops "github.com/portworx/sched-ops/k8s/stork"
 	"github.com/sirupsen/logrus"
@@ -106,6 +108,7 @@ const (
 	// optCSISnapshotClassName is an option for providing a snapshot class name
 	optCSISnapshotClassName              = "stork.libopenstorage.org/csi-snapshot-class-name"
 	defaultVolumeSnapshotClassAnnotation = "snapshot.storage.kubernetes.io/is-default-class"
+	azureEnvironmentType                 = "AZURE_ENVIRONMENT"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -865,6 +868,61 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+						var azureEnv string
+						storageCluster, err := operatorOps.Instance().GetStorageCluster("local-px-int", "kube-system")
+						if err != nil {
+							return err
+						}
+						for _, val := range storageCluster.Spec.Env {
+							logrus.Debugf("Checking azure env type in storage cluster...")
+							if val.Name == azureEnvironmentType {
+								azureEnv = val.Value
+								logrus.Infof("Azure env type from storage Cluster %s", azureEnv)
```
vsundarraj-px, line 886: Any use of printing this as Info?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -865,6 +868,61 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 				if err != nil {
 					return err
 				}
+				if driverName == "pxd" {
```
vsundarraj-px, line 871: Do you need to check this everytime we comeback ? Shouldnt this be checked during the validation step before we start any of the stages?

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -865,6 +868,61 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+					backupLocation, err := storkops.Instance().GetBackupLocation(backup.Spec.BackupLocation, backup.Namespace)
```
vsundarraj-px, line 872: Can you make this logic more of a utils call than embedding into applicationbackup and applicationrestore. The check is same is for both only the error reporting is different.
Please make ita utils.API call to avoid duplicate code.

vsundarraj-px (changes requested):

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -689,6 +692,61 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
...
 				//	BL NFS + EBS/GKE/Azure = legacy code path
 				//	s3 + EBS/GKE/Azure = legacy code path
+				if driverName == "pxd" {
+					backupLocation, err := storkops.Instance().GetBackupLocation(backup.Spec.BackupLocation, backup.Namespace)
+					if err != nil {
+						return err
+					}
+					if backupLocation.Location.Type == stork_api.BackupLocationAzure {
+						var azureEnv string
+						storageCluster, err := operatorOps.Instance().GetStorageCluster("local-px-int", "kube-system")
```
siva-portworx, line 702: Name of the stc can not be harcoded. Also the namespace.

vsundarraj-px:
I feel this check should be documented rather than handled. We are doing too many checks just to error out without any possibility of fixing the problem. And this is not per instance of backup but one time activity for azure bl and PXD cluster. For one time setting for life time of pxb should we check every 15 mins for 15mins schedule backups?

shkumari-px:
closing this pr
The error is being returned from the portworx, need to capture it
will raise a separate pr

PR #1766, opened 2024-05-20
PWX-37271: Reconciling the remote migration schedule with source side changes.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
bug

Remote Migrationschedule static copy will get reconciled with the changes in the source side so that the newly added namespaces can get failed over as part of perform failover.
Kept the reconciling just before the triggering of migration as this is not very frequent operation and better to limit the remote calls to do the comparision.

no

no

yes, 24.2.0

Test:
1. Tested new migrationschedule
2. Old migration schedule getting updated
3. changes in migrationschedule is getting updated.
4. Perform failover for later added namespaces.
Results have been updated in PWX-37271.

### DIFF
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -280,7 +280,7 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 					return fmt.Errorf(msg)
 				}
 				if updatedHash != 0 {
-					// update the hash value in the source migration schedule
+					// update the hash value in the remote migration schedule along with the source migration schedule changes
 					if remoteMigrSched.Annotations == nil {
 						remoteMigrSched.Annotations = make(map[string]string)
 					}
@@ -600,6 +600,8 @@ func (m *MigrationScheduleController) reconcileRemoteMigrationSchedule(sourceMig
 		// update directly with source hash
 		updatedHash = sourceMigrationScheduleHash
 	}
+
+	// sending the default value if there is both source and destination hash value is same
 	return updatedHash, nil
 }
 
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -597,11 +597,8 @@ func (m *MigrationScheduleController) reconcileRemoteMigrationSchedule(sourceMig
 			updatedHash = sourceMigrationScheduleHash
 		}
 	} else {
-		// get the hash of the dest migration schedule
-		updatedHash, err = getMigrationScheduleHash(destMigrationSchedule)
-		if err != nil {
-			return updatedHash, err
-		}
+		// update directly with source hash
+		updatedHash = sourceMigrationScheduleHash
 	}
 	return updatedHash, nil
 }
--- a/pkg/migration/controllers/migrationschedule.go
+++ b/pkg/migration/controllers/migrationschedule.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/migration/controllers/migrationschedule.go
```diff
@@ -174,6 +179,11 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
 			suspend := true
 			remoteMigrSchedObj.Spec.Suspend = &suspend
 			remoteMigrSchedObj.Status = stork_api.MigrationScheduleStatus{}
+			objectHash, err := getMigrationScheduleHash(migrationSchedule)
```
adityadani, line 182: Do we need to add this objectHash even when there exists a MigrationSchedule on the remote side (Think of an upgrade path)?

pkg/migration/controllers/migrationschedule.go
```diff
@@ -257,6 +270,31 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
...
+			// Reconciling the migrationschedule in remote cluster if there is a change
+			// Doing it here to limit the reconcilation check frequency
+			if remoteMigrSched != nil {
+				updatedHash, err := m.reconcileRemoteMigrationSchedule(migrationSchedule, remoteMigrSched)
+				if err != nil {
+					msg := fmt.Sprintf("Error reconciling remote migration schedule: %v", err)
+					log.MigrationScheduleLog(migrationSchedule).Error(msg)
+					return fmt.Errorf(msg)
+				}
+				if updatedHash != 0 {
```
adityadani, line 282: updateHash is always not equal to 0, since the function returns either the sourceHash or destHash. So we will keep updating it on the remote side.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -534,3 +572,47 @@ func (m *MigrationScheduleController) createCRD() error {
...
+		if err != nil {
+			return updatedHash, fmt.Errorf("unable to parse hash value %v of destination migration schedule, err: %v", val, err)
+		}
+		if sourceMigrationScheduleHash != destMigrationScheduleHash {
+			log.MigrationScheduleLog(sourceMigrationSchedule).Infof("Source migration schedule hash %v, destination migration schedule hash %v", sourceMigrationScheduleHash, destMigrationScheduleHash)
+			updatedHash = sourceMigrationScheduleHash
+		}
+	} else {
+		// get the hash of the dest migration schedule
+		updatedHash, err = getMigrationScheduleHash(destMigrationSchedule)
```
adityadani, line 601: Or you could set the sourceHash here when you know that the annotation for hash on the destination side is empty.

pkg/migration/controllers/migrationschedule.go
diptiranjanpx: This will be taken care of by the reconcileRemoteMigrationSchedule. As the annotation will not be present, there will be one update happening on existing remote static copies.

pkg/migration/controllers/migrationschedule.go
diptiranjanpx: updatedHash is an uint64. The variable inside the reconcileRemoteMigrationSchedule function is only getting updated if
1. Remote migration schedule has the hash value in annotation and sourceMigrationScheduleHash != destMigrationScheduleHash
2. If no annotation is present.

Once the hash value gets updated in remote copy, the updatedHash return from that function will be the default value (0).

pkg/migration/controllers/migrationschedule.go
diptiranjanpx: Right, this can be directly be the sourceHash. Will modify this.

pkg/migration/controllers/migrationschedule.go
```diff
@@ -257,6 +270,31 @@ func (m *MigrationScheduleController) handle(ctx context.Context, migrationSched
...
+			// Doing it here to limit the reconcilation check frequency
+			if remoteMigrSched != nil {
+				updatedHash, err := m.reconcileRemoteMigrationSchedule(migrationSchedule, remoteMigrSched)
+				if err != nil {
+					msg := fmt.Sprintf("Error reconciling remote migration schedule: %v", err)
+					log.MigrationScheduleLog(migrationSchedule).Error(msg)
+					return fmt.Errorf(msg)
+				}
+				if updatedHash != 0 {
+					// update the hash value in the source migration schedule
```
pp511, line 283: nit...I think you mean remote migration schedule in the comments here?

pkg/migration/controllers/migrationschedule.go
pp511: If annotation is present in remote and hash has not updated, we return default value of uninitialized var  updatedHash which is 0. Let's add a comment in reconcileRemoteMigrationSchedule() so it's explicit and doesn't cause confusion.

pp511 (approved):

pkg/migration/controllers/migrationschedule.go
diptiranjanpx: Done

adityadani (approved):

PR #1775, opened 2024-05-21
 PB-6976: Update backup cr status to failed status if backup failed to get started by portworx driver

shkumari-px:
Update backup cr status to failed status if backup failed to get started by portworx driver

**What type of PR is this?**
bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**RCA**
**Situation** - Before running the post exec rule, we check if the snapshot id is populated for every volume
**Bug** - If the snapshot id is not populated, we return nil at this point and we keep looping in the `volume in Progress` until the snapshot id is populated.
In case where the backup fails in the inital stages only, before the backup starts (because of wrong credential or may be any other reason), we will keep looping in the in progress state only. Since we don't mark the backup cr status to failed because of the failed snapshot status

**Updated screenshot**
Backup with pxs+non pxd vol azurechina bl, empty env type in stc
Screenshot 2024-05-24 at 5 25 49 PM
Screenshot 2024-05-24 at 5 35 00 PM

Backup with pxd+non pxd vol azureglobal bl, azurechina in cluster stc
In case some non-pxd vol fails before the pxd vol, the error msg will be that of the last failure
Screenshot 2024-05-24 at 6 03 59 PM

Check if post exec rule runs for cancelled backup
locazcstcemptycancelbkp
AzureChina backup location
Screenshot 2024-05-24 at 3 29 25 PM
Screenshot 2024-05-24 at 3 29 34 PM
Screenshot 2024-05-24 at 3 30 31 PM

Azure china bl, cluster stc empty
Screenshot 2024-05-24 at 2 46 06 PM
successful post exec
azurechinablstcempty

Azure public bl but AzureChina Cloud in stc
Screenshot 2024-05-24 at 3 18 59 PM
blazpublicstcazchina

**Previous unit testing results**
**Use Case** 
`AzureChinaCloud` env type set in cluster stc but we are taking backup on azure global backup location
Screenshot 2024-05-21 at 9 49 15 PM

'AzurePublicCloud' env type set in cluster stc but we are taking backup on azure china backup location
Screenshot 2024-05-21 at 10 26 34 PM

No environment type set in cluster stc but we are taking backup on azure china backup location
Screenshot 2024-05-21 at 10 41 21 PM

**Does this PR change a user-facing CRD or CLI?**:no

**Is a release note needed?**:no

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**:no

ss of successful post exec rule execution
Screenshot 2024-05-22 at 10 02 57 AM
Screenshot 2024-05-22 at 10 04 04 AM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -942,7 +942,19 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 					return fmt.Errorf("error getting backup status: %v", err)
 				}
 				for _, volInfo := range volumeInfos {
-					if volInfo.BackupID == "" {
+					if volInfo.Status == stork_api.ApplicationBackupStatusFailed {
+						errorMsg := fmt.Sprintf("Error backing up volume %v from namespace: %v : %v", volInfo.Volume, volInfo.Namespace, volInfo.Reason)
+						a.recorder.Event(backup,
+							v1.EventTypeWarning,
+							string(volInfo.Status),
+							errorMsg)
+
+						backup.Status.Stage = stork_api.ApplicationBackupStageFinal
+						backup.Status.FinishTimestamp = metav1.Now()
+						backup.Status.Status = stork_api.ApplicationBackupStatusFailed
+						backup.Status.Reason = errorMsg
+						break
+					} else if volInfo.BackupID == "" {
 						log.ApplicationBackupLog(backup).Infof("Snapshot of volume [%v] from namespace [%v] hasn't completed yet, retry checking status", volInfo.PersistentVolumeClaim, volInfo.Namespace)
 						// Some portworx volume snapshot is not completed yet
 						// hence we will retry checking the status in the next reconciler iteration


pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -942,7 +942,19 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
...
+						a.recorder.Event(backup,
+							v1.EventTypeWarning,
+							string(volInfo.Status),
+							errorMsg)
+
+						backup.Status.Stage = stork_api.ApplicationBackupStageFinal
+						backup.Status.FinishTimestamp = metav1.Now()
+						backup.Status.Status = stork_api.ApplicationBackupStatusFailed
+						backup.Status.Reason = errorMsg
+						break
```
vsundarraj-px, line 956: why do we need to break here and continue out side of for loop. You are setting StageFinal and status Failed. We should be return nil for reconciler to take care of failure case. 
And I think you need to incorporate partial success here.. This is a partial success for us. 
I feel instead of handling this through this bug, it is better partial success implementation implicitly takes care of it.

shkumari-px: @vsundarraj-px 
- the break here is to ensure we don't check any other volume status once one has failed
- the outer for loop is to check only portworx driver, also we continue outside the for loop so that post exec rules run properly
- This has been handled in partial success but since partial success will be implemented in 2.7.2, we have made this fix for 2.7.1
- if we return nil from here, the reconciler will keep looping in "Volume in progress" state

siva-portworx: @shkumari-px It is better to return instead of break. The execution of posterec rule in the case of failure happens in the DeleteTimestamp handling 
https://github.com/libopenstorage/stork/blob/f97b096d1235815924fb1dc23eed6e5700b22f2c/pkg/applicationmanager/controllers/applicationbackup.go#L285

shkumari-px: @vsundarraj-px https://github.com/libopenstorage/stork/blob/f97b096d1235815924fb1dc23eed6e5700b22f2c/pkg/applicationmanager/controllers/applicationbackup.go#L285 won't take care of running post exec rule since, post exec rule is being run here only when backup is in volume stage

PR #1782, opened 2024-05-23
pb-7035: fixed issue in the check to route the PXD/NFS to not nfs logic for restore.

siva-portworx:
**What type of PR is this?**
>bug
```
pb-7035: fixed issue in the check to route the PXD/NFS to not nfs logic for restore.
```

No

NA. 

24.2.1 branch
**Testing:
Validate restore of following cases:**
1) restore from backup with CSI/PXD from NFS BL.
2) restore from backup with KDMP/PXD from NFS BL.
3) restoring from backup CSI/PXD from NFS BL with restoring only PXD and only CSI.

### DIFF
--- a/pkg/applicationmanager/controllers/applicationrestore.go
+++ b/pkg/applicationmanager/controllers/applicationrestore.go
@@ -611,7 +611,6 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
 	pvcCount := 0
 	restoreDone := 0
 	backupVolumeInfoMappings := make(map[string][]*storkapi.ApplicationBackupVolumeInfo)
-	hasPXDdriver := false
 	for _, namespace := range backup.Spec.Namespaces {
 		if _, ok := restore.Spec.NamespaceMapping[namespace]; !ok {
 			continue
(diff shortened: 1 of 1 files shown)


lalat-das (approved):
lgtm

pkg/applicationmanager/controllers/applicationrestore.go
```diff
@@ -679,10 +675,10 @@ func (a *ApplicationRestoreController) restoreVolumes(restore *storkapi.Applicat
 		// Portworx driver restore is not supported via job as it can be a secured px volume
 		// to access the token, we need to run the driver.startRestore in the same context as the stork controller
 		// this check is equivalent to (if !nfs || (nfs && driverName == volume.PortworxDriverName))
-		if !nfs || hasPXDdriver {
+		for driverName, vInfos := range backupVolumeInfoMappings {
 			// Here backupVolumeInfoMappings is framed based on driver name mapping, hence startRestore()
 			// gets called once per driver
-			for driverName, vInfos := range backupVolumeInfoMappings {
+			if !nfs || (nfs && driverName == volume.PortworxDriverName) {
```
vikasit12, line 681: nit :: We can have a simplified check like
if !nfs || driverName == volume.PortworxDriverName {}

vsundarraj-px: Oops didnt realise the flag would regress like this.

siva-portworx: Yes, checking of driverName alone will working but I prefer check is gives clarity in understanding the logic :)

PR #1786, opened 2024-05-28
Fixing duplicate logging

aghodke1312:
**What type of PR is this?**
>bug
>feature
improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

Reduced the duplicate logging. there were 12 log lines repeated. Reduced it to 6. 
https://jenkins.pwx.dev.purestorage.com/job/Stork/job/stork-ocp-kubevirt-4-13/171/consoleFull
Now thees 6 are because `kubectl logs stork-test  -n kube-system -f
for i in $(seq 1 100) ; do
    test_status=$(kubectl get pod stork-test -n kube-system -o json | jq ".status.phase" -r)
    if [ "$test_status" = "Running" ]; then
        echo "Test is still running, status: $test_status"
        kubectl logs stork-test  -n kube-system -f
    else
        sleep 5
        break
    fi
done` 

no

no

no

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -13,7 +13,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 	"testing"
 	"time"
 
@@ -1936,11 +1935,7 @@ func TestMain(m *testing.M) {
 		1,
 		"Number of different kubevirt VMs deployed and validated in kubevirt tests")
 	flag.Parse()
-	var once sync.Once
-	once.Do(func() {
-		doDashboardSetup()
-	})
-
+	doDashboardSetup()
 	if err := setup(); err != nil {
 		log.Error("Setup failed with error: %v", err)
 		os.Exit(1)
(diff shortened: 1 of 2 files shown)


test/integration_test/common_test.go
```diff
@@ -1939,15 +1939,14 @@ func TestMain(m *testing.M) {
 	var once sync.Once
 	once.Do(func() {
 		doDashboardSetup()
+		if err := setup(); err != nil {
+			log.Error("Setup failed with error: %v", err)
+			os.Exit(1)
+		}
+		exitCode := m.Run()
```
pureneelesh, line 1946: So, is TestMain getting invoked multiple times in the same job? If yes, that seems to be a problem since that would mean the same set of tests are running multiple times?

pureneelesh (approved):

pureneelesh:
@aghodke1312 : changes look good. Can you please fix the commit/PR title and description since this is now a different change from the original one?

aghodke1312:
Running one more set of test with line causing duplicate logs removed.

Rohit-PX (approved):
LGTM.

PR #1790, opened 2024-05-30
PB-7173: Bug fix - Backup to not fail when there are non-PVC resources to be backed up

kgarg-px:
**What type of PR is this?**
>bug

Earlier when there were only non-PVC resources were present for the backup we were entering the loop `backup.Status.Volumes) == backup.Status.FailedVolCount` and failing the backup. Now the entry into this loop is changed and backup does not fail if only non-PVC resources are backed up

```release-note
Issue:
User Impact:
Resolution

```

After the fix: As it can be seen in the screenshot, we have taken backup of only non-PVC resources and backup is successful.
Screenshot 2024-05-30 at 5 01 34 PM
Screenshot 2024-05-30 at 5 02 07 PM

Before this PR:
Screenshot 2024-05-30 at 5 11 01 PM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -1174,7 +1174,7 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 			backup.Status.FailedVolCount++
 		}
 	}
-	if len(backup.Status.Volumes) == backup.Status.FailedVolCount {
+	if (len(backup.Status.Volumes) != 0) && (len(backup.Status.Volumes) == backup.Status.FailedVolCount) {
 		// This case signifies that none of the volumes are successfully backed up
 		// hence marking it as failed
 		backup.Status.Stage = stork_api.ApplicationBackupStageFinal


shkumari-px (approved):
LGTM

prashanthpx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -1174,7 +1174,7 @@ func (a *ApplicationBackupController) backupVolumes(backup *stork_api.Applicatio
 			backup.Status.FailedVolCount++
 		}
 	}
-	if len(backup.Status.Volumes) == backup.Status.FailedVolCount {
+	if (len(backup.Status.Volumes) != 0) && (len(backup.Status.Volumes) == backup.Status.FailedVolCount) {
```
siva-portworx, line 1177: Should we also return from this block, after updating the CR as Failed and final?

siva-portworx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
kgarg-px: yes, if failing we can return directly from here. Will raise a separate PR for that as well as the finishTimeStamp one

PR #1800, opened 2024-06-25
PWX-37845: Fixing the inmemory volumesnapshotschedule copy with the latest CR copy.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
bug

As part of createvolumesnapshot the inmemory copy becomes stale as we a getandupdate and later prune function picks the stale copy. The fix is to avoid that.

no

```release-note
Issue:
User Impact:
Resolution

```

yes, 24.2.3

### DIFF
--- a/pkg/snapshot/controllers/snapshotschedule.go
+++ b/pkg/snapshot/controllers/snapshotschedule.go
@@ -134,7 +134,7 @@ func (s *SnapshotScheduleController) handle(ctx context.Context, snapshotSchedul
 
 		// Start a snapshot for a policy if required
 		if start {
-			err := s.startVolumeSnapshot(snapshotSchedule, policyType)
+			snapshotSchedule, err = s.startVolumeSnapshot(snapshotSchedule, policyType)
 			if err != nil {
 				msg := fmt.Sprintf("Error triggering snapshot for schedule(%v): %v", policyType, err)
 				s.recorder.Event(snapshotSchedule,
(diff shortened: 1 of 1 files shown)


pkg/snapshot/controllers/snapshotschedule.go
```diff
@@ -301,12 +301,16 @@ func (s *SnapshotScheduleController) formatVolumeSnapshotName(snapshotSchedule *
...
 }
 
-func (s *SnapshotScheduleController) startVolumeSnapshot(inputSnapshotSchedule *stork_api.VolumeSnapshotSchedule, policyType stork_api.SchedulePolicyType) error {
+func (s *SnapshotScheduleController) startVolumeSnapshot(snapshotSchedule *stork_api.VolumeSnapshotSchedule, policyType stork_api.SchedulePolicyType) (*stork_api.VolumeSnapshotSchedule, error) {
+	vssKind := snapshotSchedule.GetObjectKind().GroupVersionKind().Kind
+	vssAPIVersion := snapshotSchedule.GetObjectKind().GroupVersionKind().GroupVersion().String()
+
 	// Get the latest copy of snapshotschedule for updating
-	snapshotSchedule, err := storkops.Instance().GetSnapshotSchedule(inputSnapshotSchedule.Name, inputSnapshotSchedule.Namespace)
+	snapshotSchedule, err := storkops.Instance().GetSnapshotSchedule(snapshotSchedule.Name, snapshotSchedule.Namespace)
```
pp511: nil pointer deref. needs to be handled in this function as well as the caller .

pp511 (commented):
LGTM.
Needs to be tested before merging.
We should add some UTs to handle nil pointer checks atleast.

pkg/snapshot/controllers/snapshotschedule.go
pp511: This is handled in current PR.

diptiranjanpx:
> LGTM. Needs to be tested before merging. We should add some UTs to handle nil pointer checks atleast.

UTs for the controllers will be added later in another PR.

pp511 (approved):
Let's make sure CBTs pass. 
Let's make sure we have a ticket to track adding UTs in the controller codes in Stork.
LGTM.

PR #1802, opened 2024-06-25
pb-7392: handled upgrade issues with psa

lalat-das:
- If backup is taken without psa enabled and restored in a psa environment then application backup CR won't have jobsecurityContext in its volumeinfo
- This caused the function to return user as root instead of undefinedId this caused the securityContext to get added and job pod failed
- handled the root user defination exclusively when deciding whether to add security context or not

**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test
>integration-test

**What this PR does / why we need it**: if backup taken in a release where in no spa feature present then we need to handle the restore path how to deal with the securityContext

```release-note
Issue:
User Impact:
Resolution

```

**Does this change need to be cherry-picked to a release branch?**: release-24.3.0

t
Unite test:
Updated in pb ticket : https://purestorage.atlassian.net/browse/PB-7392

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -488,6 +488,8 @@ func GetPodFromPVC(pvcName, namespace string) (*v1.Pod, error) {
 	return podList[0], nil
 }
 
+// This function will return the UID and GID from the security context of the pod or container
+// If the security context is not defined in the pod or container, it will return UndefinedId
 func getIdFromSecurityContext(securityContext interface{}) (int64, int64) {
 	uid := UndefinedId
 	gid := UndefinedId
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -613,6 +614,15 @@ func GetAppUidGid(pvcName string, namespace string, backup *stork_api.Applicatio
 			logrus.Errorf("%s: %v", fn, err)
 			return uid, gid, err
 		}
+
+	}
+	if uid == 0 {
```
vsundarraj-px, line 619: runAsUser: 0 is a valid value intended for pod to run as root. Not sure just checking this will ensure that we are denying for legitimate uid 0. Is there no way to figure out if it was legacy backup?

lalat-das: privilege and exclusive set as root are both expects same behavior when we run our job pod in a privileged PSA setting . Hence this is been taken care.

pkg/utils/utils.go
```diff
@@ -538,28 +547,20 @@ func GetPodUserId(pod *v1.Pod) (int64, int64) {
 			// No need to handle the case for the initContainers here
 			// since it is a different list in pod spec i.e. pod.Spec.InitContainers
 			securityContext := container.SecurityContext
-			uid, gid = getIdFromSecurityContext(securityContext)
-			if uid != UndefinedId || gid != UndefinedId {
-				return uid, gid
-			}
+			containerUid, containerGid = getIdFromSecurityContext(securityContext)
```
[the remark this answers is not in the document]
lalat-das: Yes we should have broken from the loop. 
But as per our discussion let me add the logic of getting the UID and GID in whichever container it is set. Not just the first one. It may happen that in an ore. than one container scenario 3rd or 4th container has set the security context then we will loose it completely , hence we can implement that logic here.
But if each container has a different securityContext defined in its own setup then it is not handle today. Infact if you think all the DBs none of them have such a security context set.
So lets agree top go with this approach now.

vsundarraj-px (approved):
LGTM

PR #1809, opened 2024-07-10
PWX-37154: stabilize KubeVirt Hyperconvergence tests

pureneelesh:
**What type of PR is this?**
>bug
>feature
>improvement
>cleanup
>api-change
>design
>documentation
>failing-test
>unit-test

integration-test

Tests had become stale and had started failing.

- remove the test specs that don't conform to our recommended settings e.g. no WFFC
- separate out specs that use populators and the ones that don't
- special handling for the specs that use the populators to get around known issues
	- use "copy" instead of "clone" in the specs since clones bypass the PreCreate() call where we do all the KubeVirt-related stuff
	- for a single-disk VM, perform one initial live-migration to get VM to a bind-mounted state
	- for a multi-disk VM, restart the VM to get to a bind-mounted state initially

- disable vps fix job until the initial hyperconvergence is verified to avoid interference

- skip inactive virt-launcher pods by checking the VM container status in validation of UpdatePX to avoid false alarms

- don't rely on the PX volume labels to get the PVC because that is unreliable

no

no

no

### DIFF
--- a/test/integration_test/kubevirt_hyperconv_test.go
+++ b/test/integration_test/kubevirt_hyperconv_test.go
@@ -607,20 +607,20 @@ func gatherInitialVMIInfo(t *testing.T, testState *kubevirtTestState) {
 
 		pvc, err := core.Instance().GetPersistentVolumeClaim(pv.Spec.ClaimRef.Name, pv.Spec.ClaimRef.Namespace)
 		log.FailOnError(t, err, "Failed to get PVC %s/%s for volume %s of context %s",
-			pv.Spec.ClaimRef.Name, pv.Spec.ClaimRef.Namespace, vol.ID, appCtx.App.Key)
+			pv.Spec.ClaimRef.Namespace, pv.Spec.ClaimRef.Name, vol.ID, appCtx.App.Key)
 
 		vmDisk.pvcName = pvc.Name
 
 		Dash.VerifyFatal(t, pvc.Spec.StorageClassName != nil, true, fmt.Sprintf(
-			"PVC %s/%s has no storageClassName", appCtx.App.NameSpace, vmDisk.pvcName))
+			"PVC %s/%s has no storageClassName", pvc.Namespace, pvc.Name))
 
 		Dash.VerifyFatal(t, pvc.Spec.VolumeName != "", true, fmt.Sprintf(
-			"PVC %s/%s has no volumeName", appCtx.App.NameSpace, vmDisk.pvcName))
+			"PVC %s/%s has no volumeName", pvc.Namespace, pvc.Name))
 		vmDisk.storageClassName = *pvc.Spec.StorageClassName
 
 		sc, err := core.Instance().GetStorageClassForPVC(pvc)
 		log.FailOnError(t, err, "Failed to get storageClass for PVC %s/%s for volume %s of context %s",
-			appCtx.App.NameSpace, vmDisk.pvcName, vol.ID, appCtx.App.Key)
+			pvc.Namespace, pvc.Name, vol.ID, appCtx.App.Key)
 
 		if sc.VolumeBindingMode != nil && *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
 			vmDisk.waitForFirstConsumer = true
@@ -642,7 +642,8 @@ func gatherInitialVMIInfo(t *testing.T, testState *kubevirtTestState) {
 				break
 			}
 		}
-		Dash.VerifyFatal(t, vmDisk.diskName != "", true, fmt.Sprintf("Failed to find disk name for PVC %s", vmDisk.pvcName))
+		Dash.VerifyFatal(t, vmDisk.diskName != "", true, fmt.Sprintf(
+			"Failed to find disk name for PVC %s/%s", pvc.Namespace, pvc.Name))
 		log.InfoD("%s attached to node %s", vmDisk, vmDisk.attachedNode.Name)
 	}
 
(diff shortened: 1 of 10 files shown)


pureneelesh:
Local run on my cluster:
```
--- PASS: TestKubevirt (8987.60s)
    --- PASS: TestKubevirt/kubevirtDeployFedoraVMWithClonePVC (10.63s)
    --- PASS: TestKubevirt/kubevirtDeployWindowsServerWithClonePVC (261.46s)
    --- PASS: TestKubevirt/kubevirtDeployFedoraVMWithClonePVCWaitFirstConsumer (191.13s)
    --- PASS: TestKubevirt/kubevirtDeployWindowsServerWithClonePVCWaitFirstConsumer (281.44s)
    --- PASS: TestKubevirt/kubevirtDeployFedoraVMMultiVolume (30.66s)
    --- PASS: TestKubevirt/kubeVirtSimulateOCPUpgrade (1892.25s)
    --- PASS: TestKubevirt/kubeVirtHypercTwoLiveMigrations (1431.55s)
    --- PASS: TestKubevirt/kubeVirtHypercHotPlugDiskCollocation (1237.78s)
    --- PASS: TestKubevirt/kubeVirtHypercVPSFixJob (457.29s)
    --- PASS: TestKubevirt/kubeVirtUpdatePX (1580.86s)
    --- PASS: TestKubevirt/kubeVirtUpdatePXBlocked (1609.01s)
PASS
time="2024-07-10T01:55:09Z" level=info msg="TestSetId 693047 update successfully"
time="2024-07-10T01:55:09Z" level=info msg="Dashboard URL : https://aetos.pwx.purestorage.com/resultSet/testSetID/693047"

DONE 12 tests in 9036.127s
```

Rohit-PX (approved):
LGTM.

Rohit-PX:
@pureneelesh can we run a round of the kubevirt-integration with a private `stork_test` container?

pureneelesh:
> @pureneelesh can we run a round of the kubevirt stork-integration test with a private `stork_test` container?

These changes are already running in jenkins. Jobs are passingfso far.

test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -560,21 +596,31 @@ func gatherInitialVMIInfo(t *testing.T, testState *kubevirtTestState) {
...
-		Dash.VerifyFatal(t, vmDisk.pvcName != "", true, "PVC name found in volume labels")
+		// Don't rely on the PX volume labels to get the PVC because it could point to a template PVC in a
+		// different namespace in some abnormal cases e.g. when "clone" instead of "copy" is used to populate
+		// the volume. We avoid clones in our test specs but let's not create an unnecessary dependency.
+		pv, err := core.Instance().GetPersistentVolume(vol.ID)
+		log.FailOnError(t, err, "Failed to get PV %s of context %s", vol.ID, appCtx.App.Key)
 
-		pvc, err := core.Instance().GetPersistentVolumeClaim(vmDisk.pvcName, appCtx.App.NameSpace)
+		Dash.VerifyFatal(t, pv.Spec.ClaimRef != nil, true, fmt.Sprintf(
+			"PV %s of app %s has no claimRef", pv.Name, appCtx.App.Key))
```
Rohit-PX, line 606:
```suggestion
			"PV %s for PVC %s used by app %s has no claimRef", pv.Name, vol.ID, appCtx.App.Key))
```

test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -797,14 +839,12 @@ func verifyHotPlugDisk(t *testing.T, testState *kubevirtTestState, hpDisk *hotPl
...
 	}
-	// set the cluster option to reduce the wait time
-	setFixVPSJobFrequency(t, testState.allNodes)
 	require.Eventuallyf(t, func() bool {
 		return isHotplugDiskCollocated(testState, hpDisk)
 	}, time.Hour, 5*time.Second, "%s was not collocated", hpDisk)
 }
 
-func setFixVPSJobFrequency(t *testing.T, allNodes map[string]node.Node) {
+func setFixVPSJobFrequency(t *testing.T, allNodes map[string]node.Node, frequencyInMinutes int) {
```
Rohit-PX, line 847: Do we need this function since all it's doing is calling `updateFixVPSJobFrequency` ?

pureneelesh: > Do we need this function since all it's doing is calling `updateFixVPSJobFrequency` ?

It is just to preserve the commented out code which should really work. At some point, we need to figure out why `volumeDriver.SetClusterOpts` is broken, fix it and use the commented out code.

test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -916,41 +951,96 @@ func verifyDisksAttachedOnSameNode(t *testing.T, testState *kubevirtTestState) {
...
+			if vmDisk.apiVol.Locator.VolumeLabels[usePopulatorAnnotation] == "true" {
+				log.Info("%s is set on volume %s of app %s. "+
+					"Need to live-migrate the VM once before verifying hyperconvergence.",
+					usePopulatorAnnotation, vmDisk.apiVol.Id, testState.appCtx.App.Key)
+				usePopulator = true
+				break
 			}
-			return true
 		}
+		Dash.VerifyFatal(t, usePopulator, true, fmt.Sprintf(
```
Rohit-PX, line 972: To clarify, we are always expecting `usePopulator` to be true here?

pureneelesh: > To clarify, we are always expecting `usePopulator` to be true here?

Yes, see the error message. If usePopulator was false, we expect the check on line 956 to be a success.

test/integration_test/kubevirt_hyperconv_test.go
```diff
@@ -916,41 +951,96 @@ func verifyDisksAttachedOnSameNode(t *testing.T, testState *kubevirtTestState) {
...
-		require.Eventuallyf(t, isBindMounted, 10*time.Minute, 30*time.Second, "%s did not switch to a bind-mount", vmDisk)
+	for _, vmDisk := range testState.vmDisks {
+		log.InfoD("Checking bind mount for %s", vmDisk)
+		mountType, err := getVMDiskMountType(testState.vmPod, vmDisk)
+		if err != nil {
+			log.Warn("Failed to get mount type of %s: %v", vmDisk, err)
+			return false, err
+		}
+		if mountType != mountTypeBind {
+			log.Warn("%s is not bind-mounted yet (mountType=%q)", vmDisk, mountType)
```
Rohit-PX, line 1039: Printing the vmName or vmID here would make debugging easier.

test/integration_test/kubevirt_hyperconv_test.go
Rohit-PX: What I meant to say was, if some other spec was used to run this test, will the logic still hold true?

pureneelesh: We are not assuming any particular spec here. We are only checking if the usePopulator annotation is present. So, it should be ok, I think. I don't understand what you mean by "some other spec".

test/integration_test/kubevirt_hyperconv_test.go
pureneelesh: > Additionally printing the vmName or vmID here would make debugging easier.

added the namespace (app context) to the error message

test/integration_test/kubevirt_hyperconv_test.go
pureneelesh: vol.ID and pv.Name are both the same. In fact, we used vol.ID to retrieve the PV a few lines above. However, your point is valid. I added apiVol.Id to the message.

test/integration_test/kubevirt_hyperconv_test.go
Rohit-PX: I mean that in both the cases where: 
`cdi.kubevirt.io/storage.usePopulator: "false"`
or
`cdi.kubevirt.io/storage.usePopulator: "true"`

In both cases the test will pass if everything else goes well

pureneelesh: Yes, we now have specs to cover both cases. Previous specs had `usePopulator=false` annotation. I have renamed them with `no-ppltr` suffix. Then added new specs without `usePopulator=false` to test them with the workaround of live-migrating or restarting the VM. These new specs have to use `copy` as a cloneStrategy. So, created a new storageClass with `copy` option.

diptiranjanpx (approved):

PR #1812, opened 2024-07-12
pb-7563: fixed the error handling in getting snapshot timeout value during kdmp csi restore.

siva-portworx:
**What type of PR is this?**
>bug

```
pb-7563: fixed the error handling in getting snapshot timeout value during kdmp csi restore.
```

no

```release-note
Issue: configmap way of setting the CSI snapshot timeout was broken.
User Impact: user were not able to change the default 30m, if the snapshot was taking more time.
Resolution: Fixed to take the value of snapshot timeout from the kdmp-config configmap.

```

23.3.0 branch.


aks-px (approved):
LGTM

PR #1829, opened 2024-08-12
PWX-37811: Ignoring witness node during clusterdomain activate/deactivate with failover/failback.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
bug

Sometimes the witness clusterdomain can get deactivated instead of source clusterdomain as part of the failover workflow.

no

TBD
```release-note
Issue:
User Impact:
Resolution

```

yes, 24.3.0

Test:
```
Manual verification and with the integration tests as mentioned in https://github.com/libopenstorage/stork/pull/1828 
```


drivers/volume/volume.go
```diff
@@ -512,6 +514,11 @@ func (c *ClusterDomainsNotSupported) DeactivateClusterDomain(*storkapi.ClusterDo
 	return &errors.ErrNotSupported{}
 }
 
+// DeactivateClusterDomain deactivates a cluster domain
```
strivedi-px, line 517:
```suggestion
// GetClusterDomainNodes returns all the cluster domains with their nodes.
```

drivers/volume/portworx/portworx.go
```diff
@@ -3168,6 +3168,52 @@ func (p *portworx) DeactivateClusterDomain(cdu *storkapi.ClusterDomainUpdate) er
...
+		if err != nil {
+			return nil, err
+		}
+		if nodeConfig.ClusterDomain == "" {
+			continue
+		}
+		if _, ok := domainToNodeMap[nodeConfig.ClusterDomain]; !ok {
+			domainToNodeMap[nodeConfig.ClusterDomain] = []*api.Node{}
+		}
+		domainToNodeMap[nodeConfig.ClusterDomain] = append(domainToNodeMap[nodeConfig.ClusterDomain], node)
```
strivedi-px, line 3212: Can't we append the node to the map inside the `if` condition without creating an empty slice?

strivedi-px (approved):
Added couple of minor nits. Rest, looks good.

drivers/volume/portworx/portworx.go
diptiranjanpx: I was trying to avoid the differntiation of slice initalization for a particular domain and appending inside the if condition.

strivedi-px (approved):

drivers/volume/volume.go
diptiranjanpx: done

drivers/volume/portworx/portworx.go
```diff
@@ -3168,6 +3168,49 @@ func (p *portworx) DeactivateClusterDomain(cdu *storkapi.ClusterDomainUpdate) er
 	return err
 }
 
+func (p *portworx) GetClusterDomainNodes() (map[string][]*api.Node, error) {
```
pp511, line 3171: This function needs UT. We can mock the clusterManager interface.

diptiranjanpx: Though clusterManager interface is mockable but to get it there should be a mockable `getClusterManagerClient` . Currently I went with the approach of adding UTs with the `getRemoteNodeName` in the latest commit. Please check.

pp511 (approved):

PR #1831, opened 2024-08-13
PWX-38560: add wait for fetching destination volume id in dataExportTest CBT

strivedi-px:
**What type of PR is this?**
>integration-test

Attempts to fix the flakiness in `dataExportTest` CBT by waiting for one minute to get the volume id from the PVC. It was failing since in some of the test runs blank value of volume id was getting returned.


diptiranjanpx:
Why not directly merge this to master?

diptiranjanpx (approved):

PR #1836, opened 2024-08-26
pb-7944: In the backupResources(), not returning  the reconciler cycle, after checking and updating the largeResource flag in the backup CR.

siva-portworx:
**What type of PR is this?**
>bug

```
1) We we exit the reconciler cycle after determining the largeResource status and updating the largeResource flag in the applicationBackup CR.
2) When the control re-enter again, we again fetch all the resources with GetResources and proceed with the upload resource step.
3) In the case of large resource, it doubles up the backup completion time.
```

no.

```release-note
Issue: In the case of large resource, the resources fetched twice, which increase the backup completion time.
User Impact: Large resource Backup takes longer time.
Resolution: Avoid the double fetch and uploading resources in the same reconciler context.

```

24.4.1

**Testing:**
1) Validated the large resource with and without fix for 750 Namespace and 91629 CM resources
Without fix: 71 mint
With fix: 42 mint.
2) Also validated that update command return updated  CR content:
```
time="2024-08-26T07:37:52Z" level=info msg="sivakumar --- before CR content &{TypeMeta:{Kind:ApplicationBackup APIVersion:stork.libopenstorage.org/v1alpha1} ObjectMeta:{Name:xxxx-1-67d0014 GenerateName: Namespace:siva-1 SelfLink: UID:e803bbd6-3ceb-4797-81c3-9e8d34f8a363 ResourceVersion:1697227 Generation:6 


time="2024-08-26T07:37:52Z" level=info msg="sivakumar --- after CR content &{TypeMeta:{Kind:ApplicationBackup APIVersion:stork.libopenstorage.org/v1alpha1} ObjectMeta:{Name:xxxx-1-67d0014 GenerateName: Namespace:siva-1 SelfLink: UID:e803bbd6-3ceb-4797-81c3-9e8d34f8a363 ResourceVersion:1697228 Generation:7 
```
3) Also validated the resource.json content.
4) Validated non-large resource case as well.
Screenshot 2024-08-26 at 5 05 57 PM

### DIFF
--- a/pkg/applicationmanager/controllers/applicationbackup.go
+++ b/pkg/applicationmanager/controllers/applicationbackup.go
@@ -2061,7 +2061,6 @@ func (a *ApplicationBackupController) backupResources(
 		if err != nil {
 			return err
 		}
-		return nil
 	}
 	// Do any additional preparation for the resources if required
 	if err = a.prepareResources(backup, allObjects); err != nil {


prashanthpx (approved):

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -2061,7 +2061,6 @@ func (a *ApplicationBackupController) backupResources(
 		if err != nil {
 			return err
 		}
-		return nil
```
lalat-das, line 2064: Couple of questions:
1. we are doing the same in case of NFS scenario too  Do we need to  take care  of it there too ? 
2. Now in case of large resource we will go ahead with rancher changes will it have any impact in rancher cluster with large resource combination. since we removed the return statement.

lalat-das: As per our discussion, for NFS we will have two times getResource() invocation.

one during large resource determination time and other is  actual resource upload time via resource backup(& RE) CR. this can be irradiated only by propagating back the large resource flag from RB --> RE --> AB and not executing this logic for NFS case in AB CR handler function.

this we can do as part of later releases.

2. regrading rancher please ignore it seems we will call it nevertheless in second time reconciler enters.

pkg/applicationmanager/controllers/applicationbackup.go
```diff
@@ -2061,7 +2061,6 @@ func (a *ApplicationBackupController) backupResources(
...
 	}
```
lalat-das, line 2064: One more thing we should take care. I remember seeing this version mismatch error while testing large resource, hence added that return since we will work on the latest AB CR version. Now if we don't return from update then the subsequent prepareResource() will work on stale copy right.. I understand today that call is empty but it may not work when it is populated for some reason in future

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: Yes, for NFS, We need to modify the RE/RB CR to get the largeResources details from executor and update the ApplicationBackup CR. Will pick it up in the next release.

pkg/applicationmanager/controllers/applicationbackup.go
siva-portworx: As posted in the description, update returns the updated copy of the CR, So we will be working on the latest copy of the CR.
```
time="2024-08-26T07:37:52Z" level=info msg="sivakumar --- before CR content &{TypeMeta:{Kind:ApplicationBackup APIVersion:stork.libopenstorage.org/v1alpha1} ObjectMeta:{Name:xxxx-1-67d0014 GenerateName: Namespace:siva-1 SelfLink: UID:e803bbd6-3ceb-4797-81c3-9e8d34f8a363 ResourceVersion:1697227 Generation:6 


time="2024-08-26T07:37:52Z" level=info msg="sivakumar --- after CR content &{TypeMeta:{Kind:ApplicationBackup APIVersion:stork.libopenstorage.org/v1alpha1} ObjectMeta:{Name:xxxx-1-67d0014 GenerateName: Namespace:siva-1 SelfLink: UID:e803bbd6-3ceb-4797-81c3-9e8d34f8a363 ResourceVersion:1697228 Generation:7 

```

lalat-das (approved):

PR #1838, opened 2024-08-28
PWX-37036: GetPersistentVolumeClaim needs to use cache in extender, webhook and monitor calls.

diptiranjanpx:
**What type of PR is this?**
improvement

Now informercache will keep the trimmed PVC information.
This will be useful to enhance performance in webhook, extender and monitor.

no

```release-note
Issue:
User Impact:
Resolution

```

**Test**
1. Tested with 100 pod deployments with webhook enabled and those using one RWO px volume.
2. Took down px in 3 nodes in a 9 node cluster to check if drivermonitor is working fine.

Integration test - https://jenkins.pwx.dev.purestorage.com/job/Users/job/strivedi/job/strivedi-extender/2/

### DIFF
--- a/pkg/extender/extender.go
+++ b/pkg/extender/extender.go
@@ -202,13 +202,15 @@ func (e *Extender) processFilterRequest(w http.ResponseWriter, req *http.Request
 		}
 		var pvc *v1.PersistentVolumeClaim
 		var err error
+		var msg string
 		if storkcache.Instance() != nil {
 			pvc, err = storkcache.Instance().GetPersistentVolumeClaim(vol.PersistentVolumeClaim.ClaimName, pod.Namespace)
+			msg = fmt.Sprintf("Unable to find PVC %s in informer cache, err: %v", vol.Name, err)
 		} else {
 			pvc, err = core.Instance().GetPersistentVolumeClaim(vol.PersistentVolumeClaim.ClaimName, pod.Namespace)
+			msg = fmt.Sprintf("Unable to find PVC %s, err: %v", vol.Name, err)
 		}
 		if err != nil {
-			msg := fmt.Sprintf("Unable to find PVC %s, err: %v", vol.Name, err)
 			storklog.PodLog(pod).Warnf(msg)
 			e.Recorder.Event(pod, v1.EventTypeWarning, schedulingFailureEventReason, msg)
 			http.Error(w, msg, http.StatusBadRequest)
@@ -505,7 +507,6 @@ func (e *Extender) collectExtenderMetrics() error {
 		if err != nil {
 			log.Errorf("failed to watch pods with informer cache for health monitoring, err: %v", err)
 		}
-		log.Errorf("failed to watch pods with informer cache for metrics, err: %v", err)
 	} else {
 		log.Warnf("Shared informer cache has not been initialized, using watch for extender metrics.")
 		if err := core.Instance().WatchPods("", fn, metav1.ListOptions{}); err != nil {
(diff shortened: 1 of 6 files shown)


strivedi-px (approved):

strivedi-px:
Lets fix the UT `TestExtenderWithCache/setup`. Failing with the following error:
```
panic: http: multiple registrations for /
```

diptiranjanpx:
> Lets fix the UT `TestExtenderWithCache/setup`. Failing with the following error:
> 
> ```
> panic: http: multiple registrations for /
> ```

Have fixed this by creating an instance of http server multiplexer and associating that to the http server of externder.

The reason the UTs were failing is because for mocking cache another suite of tests were running whose setup was failing with above error. Also the same test suite cant be used as if we set a cache instance then for all tests we have to mock the cache.

pkg/extender/extender.go
```diff
@@ -107,8 +107,9 @@ func (e *Extender) Start() error {
 		return fmt.Errorf("Extender has already been started")
 	}
 	// TODO: Make the listen port configurable
-	e.server = &http.Server{Addr: ":8099"}
-	http.HandleFunc("/", e.serveHTTP)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", e.serveHTTP)
+	e.server = &http.Server{Addr: ":8099", Handler: mux}
```
pp511, line 112: curious about the intent of this change and what does the new change do differently?

pkg/extender/extender_test.go
```diff
@@ -372,7 +395,19 @@ func TestExtender(t *testing.T) {
...
+	t.Run("teardown", func(t *testing.T) {
+		teardown(t, false)
+	})
+}
+
+func TestExtenderWithCache(t *testing.T) {
+	t.Run("setup", func(t *testing.T) {
+		setup(t, true)
+	})
+	t.Run("multipleVolumeTestWithCache", multipleVolumeTestWithCache)
```
pp511, line 407: As far as UTs are concerned, we can run all above UTs with cache enabled too right? Let's do it then.

pkg/webhookadmission/webhook.go
```diff
@@ -182,9 +183,19 @@ func (c *Controller) checkVolumeOwner(volumes []v1.Volume, namespace string) (bo
...
-		pvc, err := core.Instance().GetPersistentVolumeClaim(v.PersistentVolumeClaim.ClaimName, namespace)
-		if err != nil {
-			return false, err
+
+		var pvc *v1.PersistentVolumeClaim
+		var err error
+		if storkcache.Instance() != nil {
+			pvc, err = storkcache.Instance().GetPersistentVolumeClaim(v.PersistentVolumeClaim.ClaimName, namespace)
+			if err != nil {
+				log.Errorf("failed to watch pods with informer cache for health monitoring, err: %v", err)
```
pp511, line 192: 1. We need to return error from here.
2. Error message doesn't seem right.

pkg/extender/extender.go
diptiranjanpx: The explanation was here https://github.com/libopenstorage/stork/pull/1838#issuecomment-2322069879 . However , dropped the idea of creating separate test suite with cache enabled and added the test in the same test suite which means no need to run the extender server again, thus avoiding the error 
```
panic: http: multiple registrations for /
```

pkg/webhookadmission/webhook.go
diptiranjanpx: done

pkg/extender/extender_test.go
diptiranjanpx: Since an UT with mocked informer cache for GetPVC is enough, I feel no need to mock it in all existing tests.

pp511 (approved):
lgtm

pkg/extender/extender_test.go
strivedi-px: Is the `multipleVolumeTestWithCache` dependent on `setCacheInstance` test to configure the cache test instance here? If that's the case, running it alone would fail, right?

strivedi-px (approved):
Looks good. Had a doubt so added a comment. Was just thinking if we should be adding dependency between tests if there's any.

Update: Let's ignore it for now. I see its already being followed in the unit tests example with the `setup()` function etc.

pkg/extender/extender_test.go
diptiranjanpx: yes, running alone that will fail.

PR #1842, opened 2024-09-04
PWX-38511: Ratelimiting the number of offline nodes from which pods need to be evicted.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

Instead of evicting pods from all nodes where storage is offline, rate limiting the number of nodes from which pods need to evicted in parallel.

no

no

yes, 24.3.1

**Test**:
Tested manually in a 9 node cluster and with nodeBatchSizeForPodDeletion 2 and made the storage offline in 3 nodes.
 
Will add the UT.

### DIFF
--- a/pkg/monitor/monitor_test.go
+++ b/pkg/monitor/monitor_test.go
@@ -536,7 +536,6 @@ func testOfflineStorageNodesBatchTest(t *testing.T) {
 		podList := v1.PodList{}
 		podList.Items = append(podList.Items, *pods[i])
 		mockCacheInstance.EXPECT().ListTransformedPods(fmt.Sprintf("node%d.domain", i)).Return(&podList, nil).AnyTimes()
-		//mockCacheInstance.EXPECT().GetPersistentVolumeClaim("cvolume1", defaultNamespace).Return(pvc1, nil).Times(1)
 	}
 
 	driverNodes := getDriverNodes(len(nodes.Items))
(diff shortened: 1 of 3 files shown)


pp511 (commented):
Changes look good.
Can you attempt UTs to cover the functionality of batchChannel
or file a PWX to handle it if you want to unblock QA

strivedi-px (approved):

pkg/monitor/monitor_test.go
```diff
@@ -433,8 +516,75 @@ func testOfflineStorageNode(t *testing.T) {
...
+				VolumeName: fmt.Sprintf("volume%d", i),
+				DataNodes:  []string{driverNodes[0].StorageID},
+			},
+		}
+		wffcVolumes := make([]*volume.Info, 0)
+		volumeDriver.EXPECT().InspectNode(fmt.Sprintf("node%d", i)).Return(driverNodes[i], nil).AnyTimes()
+		volumeDriver.EXPECT().GetPodVolumes(&pods[i].Spec, pods[i].Namespace, false).Return(volumes, wffcVolumes, nil).AnyTimes()
+	}
+
+	// Since the monitor runs every 30 seconds and to ensure with certainilt one batch of pod deletes, waiting for a mx of 95 seconds
```
pp511, line 562: nit.."certainty"  
"max" 
remove "Since"

pkg/monitor/monitor_test.go
```diff
@@ -433,8 +516,75 @@ func testOfflineStorageNode(t *testing.T) {
...
+	require.NoError(t, err, "failed to get pods")
+	require.Equal(t, numNodes-nodeBatchSizeForPodDeletion, len(podList.Items), "only batch number of pods should have been deleted")
+
+	time.Sleep(65 * time.Second)
+	podList, err = core.Instance().ListPods(nil)
+	require.NoError(t, err, "failed to get pods")
+	require.Equal(t, numNodes-(2*nodeBatchSizeForPodDeletion), len(podList.Items), " Another batch number of pods should have been deleted")
+
+	// Last iterations
+	time.Sleep(65 * time.Second)
```
pp511, line 580: Help me understand how did you come up with 95, 65 and 65 numbers here?

pp511 (commented):
LGTM. Just one minor nit and one question.

pkg/monitor/monitor_test.go
pp511: 60 minutes is the exp Backoff. So first set of Pods will be deleted by 60 + 5 buffer followed by a wait of 30 mins. Then you are checking if the Pods are deleted. Understood.

pp511 (approved):
lgtm

pkg/monitor/monitor_test.go
diptiranjanpx: Removed the comment itself and put the break up of wait times

PR #1845, opened 2024-09-05
Memory consumption improvements in health monitor.

diptiranjanpx:
**What type of PR is this?**
improvement

This PR has following fixes
1. Reducing memory usage in stork health monitor
     a. Filtering the pods for each offline node
     b. Limiting number of goroutines wrt offline nodes need to be processed.
     c. Doing a fresh inspect of node before proceeding with pod deletes.
     d. Pod deletions to happen in batch.
2. Informer cache to hold PVC data which need to be used in extender, webhook and monitor.
3. Stopping retrying of volumesnapshots after it hit error.

no

no

yes, 24.3.1


strivedi-px (approved):

pkg/snapshot/controllers/snapshotschedule.go
```diff
@@ -246,6 +263,54 @@ func (s *SnapshotScheduleController) updateVolumeSnapshotStatus(snapshotSchedule
 	return nil
 }
 
+// cleanupErroredSnapshots cleans any volumesnapshot CRs that are in Error state. It doesn't remove them
+// from the volumesnapshotschedule items list though.
+func (s *SnapshotScheduleController) cleanupErroredSnapshots(snapshotSchedule *stork_api.VolumeSnapshotSchedule) (err error) {
```
pp511, line 268: I didn't review the original CR so trying to understand this better about the intent here.
We delete the snapshots but not remove from the schedule list because we rely on the prune to clean it up?
If yes then prune can delete the error snapshots too right?

pkg/snapshot/controllers/snapshotschedule.go
```diff
@@ -246,6 +263,54 @@ func (s *SnapshotScheduleController) updateVolumeSnapshotStatus(snapshotSchedule
...
+				continue
+			}
+
+			if snapshotStatus == snapv1.VolumeSnapshotConditionError {
+				log.VolumeSnapshotScheduleLog(snapshotSchedule).Infof("Going to delete the errored out snapshot: %v,createdAt: %v,error: %v", snapshot.Name, snapshot.CreationTimestamp.Time, snapshotError)
+				err := k8sextops.Instance().DeleteSnapshot(snapshot.Name, snapshotSchedule.Namespace)
+				if err != nil && !errors.IsNotFound(err) {
+					log.VolumeSnapshotScheduleLog(snapshotSchedule).Warnf("Error deleting errored out snapshot %v: %v", snapshot.Name, err)
+					continue
+				}
```
pp511, line 292: nit.. We should try to aim for complete coverage unless something is actually blocking us. If "Added lines #L290 - L291 were not covered by tests" messages come up then there has to be a reason for ignoring these.

pkg/snapshot/controllers/snapshotschedule.go
diptiranjanpx: We don't want to wait till prune takes care of deleting it as prune will work based on `retain` value and till that time the retries on the errored snapshots will keep happening with old cloudsnap id. So this fix will make sure the volumesnapshot gets deleted but its info remains in schedule till pruning removes it as per its own logic.

pkg/snapshot/controllers/snapshotschedule.go
strivedi-px: @pp511 Got it. I'll add tests for these scenarios as well.

strivedi-px: But ideally, in these cases codecov should fail as well, right? Most of the times, we don't check the `Files changed` tab of a PR where these messages come up.

PR #1851, opened 2024-09-16
PWX-36167: Duplicate offline nodes should be excluded for pod eviction if another online node exists with same scheduler ID.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
bug
unit-test

Currently stork has the logic of removing the duplicate offline nodes if there exists an online node with same IP.
Extending this logic to check wrt to schedulerID and remove any duplicate offline nodes.

no

yes
```release-note
Issue: Stork health monitor was considering stale node entries which have storage offline for pod eviction.
User Impact:  Pods were getting evicted from an online node if this node gets repaired and come with a different IP address.
Resolution: Storage offline nodes will not get considered for pod evictions if there exists another node with storage status as online but with same scheduler ID as that of the offline node.
```

yes, 24.3.1, 24.3.2 and 24.3.3

### DIFF
--- a/drivers/volume/volume.go
+++ b/drivers/volume/volume.go
@@ -650,23 +650,26 @@ func IsNodeMatch(k8sNode *v1.Node, driverNode *NodeInfo) bool {
 }
 
 // RemoveDuplicateOfflineNodes Removes duplicate offline nodes from the list which have
-// the same IP as an online node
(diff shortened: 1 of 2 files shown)


pp511 (approved):
lgtm

PR #1852, opened 2024-09-19
PWX-39128: Support env variable SKIP_AWS_DRIVER_INIT to skip the aws driver init in stork.

diptiranjanpx:
Signed-Off-By: Diptiranjan

**What type of PR is this?**
improvement

In some of the clusters we are seeing stork binary takes time to load the import modules and do their inits (like aws sdk go's Session creation). Till this gets root caused, adding an environment variable "SKIP_AWS_DRIVER_INIT" if set will skip the aws driver init.

Note: This should not be used with px-backup.

no

yes, 
```release-note
We can use SKIP_AWS_DRIVER_INIT="true" as environment variable to stork pod if stork init gets hung for a long time.
```

yes, 24.3.1, 24.3.2 and 24.3.3

**Test**:

```
➜  stork git:(PWX-39128) ✗ ks exec -it stork-557475cf87-kg7fc -- env | grep AWS
SKIP_AWS_DRIVER_INIT=true
```

stork logs:
```
➜  stork git:(PWX-39128) ✗ ks logs -f stork-557475cf87-kg7fc
time="2024-09-19T10:47:44Z" level=info msg="Skipping aws driver init"
time="2024-09-19T10:47:47Z" level=warning msg="Export USE_GKE_GCLOUD_AUTH_PLUGIN=True"
time="2024-09-19T10:47:47Z" level=info msg="Starting stork version 99.9.9-7333f41"
time="2024-09-19T10:47:47Z" level=info msg="Setting the controller cache sync-timeout as 2 mins."
time="2024-09-19T10:47:47Z" level=info msg="shared informer cache has been intialized"
time="2024-09-19T10:47:47Z" level=info msg="Using driver pxd"
```

Test - https://jenkins.pwx.dev.purestorage.com/job/Stork/view/Stork%2024.3.1-dev/job/new-dr-workflow/227/ 

Ran the above test with the env variable set.

### DIFF
--- a/drivers/volume/aws/aws.go
+++ b/drivers/volume/aws/aws.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	aws_sdk "github.com/aws/aws-sdk-go/aws"
@@ -672,13 +673,17 @@ func (a *aws) IsVirtualMachineSupported() bool {
 }
 
 func init() {
-	a := &aws{}
-	err := a.Init(nil)
-	if err != nil {
-		logrus.Debugf("Error init'ing aws driver: %v", err)
-	}
-	if err := storkvolume.Register(storkvolume.AWSDriverName, a); err != nil {
-		logrus.Panicf("Error registering aws volume driver: %v", err)
+	if os.Getenv("SKIP_AWS_DRIVER_INIT") == "true" {
+		logrus.Infof("Skipping aws driver init")
+	} else {
+		a := &aws{}
+		err := a.Init(nil)
+		if err != nil {
+			logrus.Debugf("Error init'ing aws driver: %v", err)
+		}
+		if err := storkvolume.Register(storkvolume.AWSDriverName, a); err != nil {
+			logrus.Panicf("Error registering aws volume driver: %v", err)
+		}
 	}
 }
 


drivers/volume/aws/aws.go
```diff
@@ -672,13 +673,17 @@ func (a *aws) IsVirtualMachineSupported() bool {
...
 
 func init() {
-	a := &aws{}
-	err := a.Init(nil)
-	if err != nil {
-		logrus.Debugf("Error init'ing aws driver: %v", err)
-	}
-	if err := storkvolume.Register(storkvolume.AWSDriverName, a); err != nil {
-		logrus.Panicf("Error registering aws volume driver: %v", err)
+	if os.Getenv("SKIP_AWS_DRIVER_INIT") == "true" {
```
strivedi-px, line 676: The changes look good. Had a query which out of context of this PR but is there any specific reason to use `init` function here? Pointing out since in Go its not considered a good practice nowadays. Should we remove it from here or will it be an unnecessary change?

drivers/volume/aws/aws.go
```diff
@@ -672,13 +673,17 @@ func (a *aws) IsVirtualMachineSupported() bool {
...
+		logrus.Infof("Skipping aws driver init")
+	} else {
+		a := &aws{}
+		err := a.Init(nil)
+		if err != nil {
+			logrus.Debugf("Error init'ing aws driver: %v", err)
```
strivedi-px, line 682: Is there any reason to move to registering the driver if we have already failed here? Should we be panicking here if the initialisation fails?

drivers/volume/aws/aws.go
strivedi-px: Just checked the code and found that it is a practice which we are following for all the drivers to register them before stork starts (they are all imported in `cmd/stork/stork.go`). Looks like then its a valid case for using `init` function. Can ignore this review comment.

drivers/volume/aws/aws.go
diptiranjanpx: I agree it can be an incomplete init and later accessing the driver may fail elsewhere. But for now I don't want to change anything in the existing code flow. We can discuss on this later with the backup team.

pp511 (approved):

pp511:
Need to checkand fix  CBT failures.

strivedi-px (approved):

PR #1854, opened 2024-09-26
pwx-22735: add auth annotations in volumesnapshot

strivedi-px:
**What type of PR is this?**
bug

Adds the authentication related annotations in stork volumesnapshots in order to fix volume snapshotting in auth clusters.

No

No

Yes, `24.4.0` branch.

**Test run**
https://jenkins.pwx.dev.purestorage.com/job/Users/job/strivedi/job/snapshot-auth-cluster/22/
https://jenkins.pwx.dev.purestorage.com/job/Users/job/strivedi/job/snapshot-auth-cluster/17/

### DIFF
--- a/test/integration_test/snapshot_test.go
+++ b/test/integration_test/snapshot_test.go
@@ -24,7 +24,6 @@ import (
 	storkops "github.com/portworx/sched-ops/k8s/stork"
 	"github.com/portworx/sched-ops/task"
 	"github.com/portworx/torpedo/drivers/scheduler"
-	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


test/integration_test/snapshot_test.go
```diff
@@ -1571,30 +1582,138 @@ func validateSnapshotCleanup(scheduleName, namespace string) (err error) {
...
+	defer func() {
+		err = storkops.Instance().DeleteSchedulePolicy(schedPolicy.Name)
+		log.FailOnError(t, err, fmt.Sprintf("Schedule policy %v not deleted", schedPolicy.Name))
+	}()
+	log.InfoD("Created schedulepolicy %v with %v minute interval", policyName, policyInterval)
+
+	/////////////////////////////////////
+	// Create and deploy storage class //
+	/////////////////////////////////////
+	storageClassName := "autosnap-sc"
```
diptiranjanpx, line 1639: Why are we creating this dynamically instead of applying from a spec from integration test?

strivedi-px: This is added to configure the storageclass and PVC to use the `px-user-token` instead of the admin token which gets automatically configured if we deploy applications using spec in integration tests.

diptiranjanpx (approved):

PR #1863, opened 2024-10-07
test: modify cbt tests

strivedi-px:
**What type of PR is this?**
Adds couple of required CBT tests and removes a test since it uses the old way of DR. That test is being run in integration tests pipeline periodically.

No

No

No

### DIFF
--- a/test/integration_test/common_test.go
+++ b/test/integration_test/common_test.go
@@ -197,11 +197,12 @@ func TestSnapshot(t *testing.T) {
 func TestStorkCbt(t *testing.T) {
 	t.Run("deploymentTest", deploymentMigrationTest)
 	t.Run("dataExportTest", TestDataExportRsync)
-	t.Run("testMigrationFailoverFailback", testMigrationFailoverFailback)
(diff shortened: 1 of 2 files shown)


test/integration_test/common_test.go
```diff
@@ -195,13 +195,35 @@ func TestSnapshot(t *testing.T) {
...
+	log.InfoD("Backup path being used: %s", backupLocationPath)
+	setDefaultsForBackup(t)
+
+	// get the destination kubeconfig from configmap in source cluster so that it can be passed to storkctl commands
+	// since both the dr cli commands run in destination cluster
+	destinationKubeConfigPath, err = getDestinationKubeConfigFile()
+	log.FailOnError(t, err, "Error getting destination kubeconfig file")
+
+	// Get the source kubeconfig from configmap.
+	srcKubeConfigPath, err = getDestinationKubeConfigFile()
```
diptiranjanpx, line 216: why the srcKubeconfigPath when the function called is getDestinationKubeConfigFile?

strivedi-px: Fixed

diptiranjanpx (approved):

